PR #3, opened 2016-02-12
Apply RBAC in SSUI

AparnaKarve:
The new attribute ```?attributes=authorization``` in the /api entrypoint (https://github.com/ManageIQ/manageiq/pull/6597) retrieves all product features identifiers that the user is entitled to.
SSUI needs to hide/show the UI elements based on what the user is entitled to do.

https://trello.com/c/Dr2IKoAt/150-5-ssui-enforce-rbac-at-login
https://bugzilla.redhat.com/show_bug.cgi?id=1298393

### DIFF
--- a/client/app/services/session.service.js
+++ b/client/app/services/session.service.js
@@ -77,8 +77,8 @@
     }
 
     function setRBAC(productFeatures) {
-      setRBACForActions(productFeatures);
-      setRBACForNavigation(productFeatures);
+      $state.actionFeatures = setRBACForActions(productFeatures);
+      $state.navFeatures = setRBACForNavigation(productFeatures);
     }
 
     function setRBACForNavigation(productFeatures) {
--- a/client/app/services/session.service.js
+++ b/client/app/services/session.service.js
@@ -83,9 +83,9 @@
 
     function setRBACForNavigation(productFeatures) {
       var features = {
-        dashboard: {show: entitledForDashboard(productFeatures)},
-        services: {show: entitledForServices(productFeatures)},
-        requests: {show: entitledForRequests(productFeatures)},
+        dashboard:   {show: entitledForDashboard(productFeatures)},
+        services:    {show: entitledForServices(productFeatures)},
+        requests:    {show: entitledForRequests(productFeatures)},
         marketplace: {show: entitledForServiceCatalogs(productFeatures)}
       };
       model.navFeatures = features;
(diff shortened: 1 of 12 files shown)


client/app/services/session.service.js
```diff
@@ -73,5 +74,69 @@
...
+
+      return angular.isDefined(serviceFeature);
+    }
+
+    function entitledForServiceCatalogs(productFeatures) {
+      if (angular.isDefined(productFeatures.svc_catalog_provision)) {
+        return true;
+      } else {
+        return false;
+      }
```
Fryguy: Do you need the whole if/then construct?  Can't this just be...

```js
function entitledForServiceCatalogs(productFeatures) {
  return angular.isDefined(productFeatures.svc_catalog_provision);
}
```

Same goes for the other similar methods below.

client/app/services/session.service.js
```diff
@@ -73,5 +74,69 @@
...
+        service_retire_now: {show: angular.isDefined(productFeatures.service_retire_now)}
+      };
+      model.actionFeatures = features;
+
+      return model.actionFeatures;
+    }
+
+    function entitledForServices(productFeatures) {
+      var serviceFeature = lodash.find(model.actionFeatures, function(o) {
+        return o.show === true;
```
Fryguy: Does this function need o.show to be a literal true?  Can this just be...

```js
    function entitledForServices(productFeatures) {
      var serviceFeature = lodash.find(model.actionFeatures, function(o) { return o.show; });
      return angular.isDefined(serviceFeature);
    }
```

client/app/services/session.service.js
```diff
@@ -73,5 +74,69 @@
...
+      setRBACForActions(productFeatures);
+      setRBACForNavigation(productFeatures);
+    }
+
+    function setRBACForNavigation(productFeatures) {
+      var features = {
+        dashboard: {show: entitledForDashboard(productFeatures)},
+        services: {show: entitledForServices(productFeatures)},
+        requests: {show: entitledForRequests(productFeatures)},
+        marketplace: {show: entitledForServiceCatalogs(productFeatures)}
```
Fryguy: I feel like alignment would look nicer here and in the next function.

h-kataria:
:+1:  verified in SSUI.

AparnaKarve:
Waiting for UX to provide layout suggestions for a screen where the RBAC settings are applied for a user who is not entitled to view the ```Requests``` section, as shown in the screenshot below -

screen shot 2016-02-22 at 11 33 05 am

AparnaKarve:
@dclarizio I was thinking of getting the layout issue (above) addressed in this PR but on second thoughts, can we get this merged and handle it in a separate PR after we hear back from UX?

I have added RBAC for ```Set Ownership``` and have addressed PR feedback as well.

AparnaKarve:
@miq-bot remove_label wip

dclarizio:
@AparnaKarve please a few tests that validate the allowed feature ids are honored on the JS side.

AparnaKarve:
@dclarizio Added a couple tests to set RBAC rules based on feature identifiers.

client/app/services/session.service.js
AparnaKarve: The above function does need ```o.show``` to be a literal true since that in turn will determine if ```serviceFeature``` should be an ```undefined``` or not.  
```serviceFeature``` should stay ```undefined``` if ```o.show``` is not true.

PR #5, opened 2016-02-12
Angular version upgraded

romanblanco:
I think this one is possible to merge now. There is no need to wait for other projects to adapt to new AngularJS version.

@zeari @AparnaKarve @abonas @himdel


himdel:
@romanblanco can you try adding this change here please? It *might* fix the bower failures, or it might make it worse, I'm not sure yet, but we should test it I guess.. :)

```diff
diff --git a/.travis.yml b/.travis.yml
index e6df7e6..39fcdd4 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ cache:
   - node_modules
 before_install:
 - gem install sass
+- bower install -F
 notifications:
   webhooks:
     urls:
```

AparnaKarve:
@romanblanco I'm on board with this upgrade. :+1:

romanblanco:
@chessbyte I think this one is ready to merge :blush:

himdel:
The `phantomjs-polyfill` lib is needed only for tests (whatever's running under PhantomJS) - that's because PhantomJS does not implement `Function#bind`, which some code depends on. (Bard tries to provide one, but not compatible enough.)

himdel:
The `travis.yml` change is to sync with https://github.com/ManageIQ/manageiq-appliance/pull/59, + we're always cleaning the cache now.

I'd like to do something like `bower install -F || rm -rf bower_components/ && bower install -F` but unfortunately bower does not return a non-ok status if it fails because of a conflict => forcing clean slate until we have a better way.

PR #7, opened 2016-02-16
Group switcher - disable currently chosen group

himdel:
this changes the group switcher to show the currently chosen group disabled (non-clickable)

also the title was changed from "Select to change to another Group" to "Change to this Group" or "Currently Selected Group"

grp_other


himdel:
classic ui PR: https://github.com/ManageIQ/manageiq/pull/6687

PR #8, opened 2016-02-18
[WIP] Rails 5 fix - depend on an express-http-proxy version that can handle Transfer-Encoding: chunked & IPv6

himdel:
This fixes the `Parse error` messages (in node console) and the corresponding `net::ERR_EMPTY_RESPONSE` (in browser).

Also change the proxy from `127.0.0.1:3000` to `http://[::1]:3000`, and update `express-http-proxy` to parse that correctly.

Not sure if we want to merge like this, or wait for https://github.com/villadora/express-http-proxy/pull/47 and https://github.com/villadora/express-http-proxy/pull/48.

Also relevant: https://github.com/nodejs/node/issues/5308 - PR https://github.com/nodejs/node/pull/5314

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     "dateformat": "^1.0.11",
     "debug": "^2.1.3",
     "del": "^2.0.2",
-    "express-http-proxy": "^0.6.0",
     "glob": "^5.0.3",
     "gulp": "^3.8.11",
     "gulp-angular-gettext": "^2.1.0",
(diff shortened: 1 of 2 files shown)


himdel:
@miq-bot add_label wip

himdel:
WIP because the IPv6 change is not quite finished.. `http.request` with `hostname: '::1'` and `port: 3000` will try to access `::1:3000` instead of `[::1]:3000`, which fails.

himdel:
Closing in favour of #11 - `express-http-proxy` development is ..less active these days..

PR #9, opened 2016-02-18
Bundle sass gem

imtayadeway:


himdel:
@imtayadeway just noticed you added `Gemfile.lock` - do we really want it in git or was that an oversight? (given that we don't have it in manageiq..)

imtayadeway:
@himdel that wasn't an oversight - it's more unusual for a project *not* to track the `Gemfile.lock` to specify dependencies. This way you can be certain which version of the gem your developers are using, and upgrading it should only involve an update to the `Gemfile.lock` file. The `Gemfile` should only be updated if there's a problem with a specific version of the gem. There's a good guide on this here: http://www.prioritized.net/blog/embracing-promiscuous-gemfiles

himdel:
OK, thanks :) Mostly just checking, I ran my usual `bundle update && npm install && gulp serve-dev` combo and got surprised that I have changes..

PR #11, opened 2016-02-23
Switch from express-http-proxy to node-http-proxy, proxy to [::1]:3000

himdel:
node-http-proxy development is more active, and it seems it can handle both `Transfer-Encoding: chunked` and proxying to `http://[::1]:3000/api`.

Closes #8

@AparnaKarve can you test it too please?

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,10 @@ The Self Service UI for the [ManageIQ](http://github.com/ManageIQ/manageiq) proj
 - In the `manageiq-ui-self_service` directory, start the development version of
   the self service UI with `gulp serve-dev`, which will start the UI listening
   on http://localhost:3001, and talking to the REST API at
-  http://localhost:3000.  This command will also open a browser page to
-  http://localhost:3001/self_service/login .
+  http://[::1]:3000.  This command will also open a browser page to
+  http://localhost:3001/self_service/login.  
+  The REST API host can be overriden via a PROXY\_HOST environment variable, for
+  example: `PROXY_HOST=127.0.0.1:3000 gulp serve-dev`.
 
 ## Deployment
 
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     "dateformat": "^1.0.11",
     "debug": "^2.1.3",
     "del": "^2.0.2",
-    "express-http-proxy": "^0.6.0",
     "glob": "^5.0.3",
     "gulp": "^3.8.11",
     "gulp-angular-gettext": "^2.1.0",
@@ -92,6 +91,7 @@
   "dependencies": {
     "body-parser": "^1.12.2",
     "express": "^4.12.3",
+    "http-proxy": "^1.13.2",
     "morgan": "^1.5.2",
     "serve-favicon": "^2.2.0"
   },
--- a/server/app.js
+++ b/server/app.js
@@ -2,7 +2,7 @@
 'use strict';
 
 var express = require('express');
-var proxy = require('express-http-proxy');
+var httpProxy = require('http-proxy');
 var app = express();
 var router = express.Router();
 var bodyParser = require('body-parser');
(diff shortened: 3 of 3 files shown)


himdel:
`node-http-proxy` also supports websockets...

AparnaKarve:
@himdel This does seem to work most of the times. However, I occasionally see an error when I click on Service Catalogs - it is simply unable to fetch the list and this is what I see in the log -
```
[----] I, [2016-02-22T19:24:38.548839 #66659:3fcb26abcabc]  INFO -- : Processing by ApiController#show as JSON

[----] I, [2016-02-22T19:24:38.548985 #66659:3fcb26abcabc]  INFO -- :   Parameters: {"expand"=>"resources", "attributes"=>"picture,picture.image_href,service_template_catalog.name", "filter"=>["service_template_catalog_id>0", "display=true"], "collection"=>"service_templates"}

[----] I, [2016-02-22T19:24:38.551148 #66659:3fcb26ab8fc0]  INFO -- : Processing by ApiController#show as JSON

[----] I, [2016-02-22T19:24:38.551245 #66659:3fcb26ab8fc0]  INFO -- :   Parameters: {"expand"=>"resources", "sort_by"=>"name", "sort_options"=>"ignore_case", "collection"=>"service_catalogs"}

[----] I, [2016-02-22T19:24:38.603918 #66659:3fcb26ab8fc0]  INFO -- : Completed 400 Bad Request in 53ms (Views: 0.6ms | ActiveRecord: 0.0ms)

```

One other thing that I noticed was, none of the images are loaded in the list views and detailed views. 
It looks like it is not able to fetch the image hrefs.

himdel:
> One other thing that I noticed was, none of the images are loaded in the list views and detailed views. 
> It looks like it is not able to fetch the image hrefs.

Oh, thanks for noticing, that's caused by the SSUI split .. only affects SSUI in devel mode, otherwise /pictures is mapped fine .. but d6571a4 should fix it by also proxying the `/pictures` folder to the old UI.

himdel:
The Service Catalog thing.. sometimes the API returns `
{"error":{"kind":"bad_request","message":"name is not a valid attribute for ServiceTemplateCatalog","klass":"ApiController::BadRequestError"}}` .. but I'm thinking it may be a problem on the api side (at least, that response is our ruby exception, so it shouldn't be caused by the proxy ..  I hope :)

himdel:
@AparnaKarve I've created https://github.com/ManageIQ/manageiq/issues/6878 for the Service Catalog problem.

himdel:
I'd like to get this merged.. @h-kataria @imtayadeway would you mind giving this a whilr as well?

h-kataria:
ran SSUI with this PR, seems to be working fine. Labeled this PR as bug as this is related to some people seeing issues with Rails 5.

dclarizio:
@matthewd please take a look

matthewd:
Sounds reasonable.

My only concern is that as we're hard-coding the IPv6 IP, this won't work for someone who has IPv6 disabled, say... or has a more unusual value for localhost. Perhaps it'd be worth making it (plus the port, I guess) overridable with an env var?

himdel:
@matthewd That's my concern too, unfortunately it seems impossible to force puma to listen on both `::1` and `127.0.0.1` when both are valid `localhost` resolutions, and node always seems to pick the other choice than ruby :).

So.. overridable with an env var sounds really great.. will do that, thanks! :)

himdel:
Done, we're now reading a `PROXY_HOST` environment variable and building the urls from that, defaulting to `[::1]:3000`. (Aand added an example of just that in the README :). )

imtayadeway:
@AparnaKarve I believe your issue should be addressed in https://github.com/ManageIQ/manageiq/pull/6892

imtayadeway:
@himdel that does the trick for me :thumbsup:

AparnaKarve:
@imtayadeway Thanks! I will try how https://github.com/ManageIQ/manageiq/pull/6892 works.

@himdel @matthewd Thanks for the environment variable trick. :+1:  Backward compatibility with ```ipv4``` was the first thing that I thought about when I saw this PR.

PR #12, opened 2016-02-24
Enable $compile debugging info in DEV mode (+ add a concept of dev-mode)

himdel:
`$compileProvider.debugInfoEnabled(false)` makes the code faster but at
the cost of not being able to inspect scopes and bindings.

Since we don't have any concept of dev/production environments in SSUI
yet (and we can't rely on handling it in the node server since that's not used in
production), adding a simple logic that looks at the current hostname,
checking for `localhost`, `127.0.0.1` and `::1`.

### DIFF
--- a/client/app/core/config.js
+++ b/client/app/core/config.js
@@ -5,14 +5,22 @@
     .config(configure)
     .run(init);
 
+  var DEVEL_DOMAINS = [
+    'localhost',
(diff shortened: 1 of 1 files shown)


himdel:
(For the curious, the main difference is that `angular.element('body').scope()` returns undefined when false.)

Issue #14, opened 2016-02-25
Required fields not honored when ordering a service


eclarizio:
Go to the service catalog, choose a service to order that has a dialog attached to it with required fields. Ensure the required field(s) are not filled out, and observe that the service is still submitted "successfully".

chriskacerguis:
@chalettu - can you take a look at this one?

chriskacerguis:
@AllenBW here you go!

eclarizio:
Tested again for @AllenBW since this is pretty old. Works now, so closing.

PR #15, opened 2016-02-25
Add support for custom buttons with attached dialogs

eclarizio:
There is currently an issue where when a required field is not filled in or given, the API call still returns a 200 success, and so the SSUI returns back to the service with a message saying "Success! X field is required".

While testing this, an issue was also discovered where while ordering a service through a service template custom dialogs do not honor the required fields and allow submissions with "incomplete" forms. See git issue: #14

~~I'm also having a lot of issues with specs, so I figured I'd throw this WIP out there to see if travis works any better.~~

### DIFF
--- a/client/app/components/custom-button/custom-button.directive.js
+++ b/client/app/components/custom-button/custom-button.directive.js
@@ -26,7 +26,7 @@
 
     function link(scope, element, attrs, vm, transclude) {
       vm.activate();
-      
+
       var win = angular.element($window);
       win.bind('resize', function() { 
         scope.$apply();
@@ -86,7 +86,7 @@
           $state.go('services.custom_button_details', {
             button: button,
             buttonId: button.id,
-            dialog: button.resource_action.dialog,
+            dialogId: button.resource_action.dialog_id,
             serviceId: serviceId,
             serviceTemplateCatalogId: serviceTemplateCatalogId
           });
--- a/client/app/components/custom-button/custom-button.directive.js
+++ b/client/app/components/custom-button/custom-button.directive.js
@@ -11,7 +11,9 @@
       replace: true,
       scope: {
         customActions: '=',
-        actions: '=?'
+        actions: '=?',
+        serviceId: '=',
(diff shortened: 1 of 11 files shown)


gmcculloug:
Tested PR with @eclarizio and found a few issues:

1. Button group names are appended with " | ServiceTemplate | <service_template_id>"
2. Button groups are hidden by the UI if the screen width is decreased.  (Exacerbated by item 1)
3. When the service is loaded the Custom Buttons are fully resolved and any dynamic dialog fields are processed at that time.  The dialogs should only be resolved when the custom button is about to be displayed and the dynamic fields should be resolved at that time.

Item 1 example
image

Item 2 example
image

eclarizio:
Depends on: https://github.com/ManageIQ/manageiq/pull/7117

eclarizio:
@gmcculloug I've fixed the issues mentioned above, but on Friday when I rebased I ran into issues actually getting the UI up and running so I couldn't actually test. Some of the specs are also missing tests due to that setback, but I figured I'd throw up what I have if you can test it.

eclarizio:
@gmcculloug Alright, this should be good to go, just updated with the latest. Still trying to get my environment up and working but the code is ready.

gmcculloug:
Further testing found that refreshing a dynamic field for a custom button fails because it currently references the dialog assigned to the "provision" actions for the service's template.  The previous issues appear to be working at this point.

eclarizio:
Currently waiting for https://github.com/ManageIQ/manageiq/pull/7147 to be un-wip'd and merged, since this now uses an API call introduced in that PR that was meant to fix the issue described above.

eclarizio:
Reference https://github.com/ManageIQ/manageiq/pull/7330 above is an additional feature that will include the custom buttons that have been assigned via the "Automate/Customization" interface.

AparnaKarve:
@dclarizio Verified in SSUI. LGTM :+1:

gmcculloug:
@dclarizio Looks good.

PR #22, opened 2016-03-01
Fix broken redirect to /login when user is not logged in

himdel:
reintroduce the login transition removed in e3d67eb21

@AparnaKarve this shouldn't break anything else, right?

Without this, going to /self_service without being logged in (but with a previous session's token saved), you get stuck on the loading screen. (Going directly to /login works either way.)

### DIFF
--- a/client/app/blocks/router/router-helper.provider.js
+++ b/client/app/blocks/router/router-helper.provider.js
@@ -90,7 +90,7 @@
         msg = 'Error routing to ' + destination + '. '
           + (error.data || '') + '. <br/>' + (error.statusText || '')
           + ': ' + (error.status || '');
-        logger.warning(msg, [toState]);
+        logger.warning(msg, [toState, error]);
         $location.path('/');
       }
 
--- a/client/app/config/authorization.config.js
+++ b/client/app/config/authorization.config.js
@@ -56,7 +56,7 @@
   }
 
   /** @ngInject */
-  function init($rootScope, $state, Session, jQuery, $sessionStorage) {
+  function init($rootScope, $state, Session, jQuery, $sessionStorage, logger) {
     $rootScope.$on('$stateChangeStart', changeStart);
     $rootScope.$on('$stateChangeError', changeError);
     $rootScope.$on('$stateChangeSuccess', changeSuccess);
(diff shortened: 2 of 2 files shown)


himdel:
@romanblanco FY__

AparnaKarve:
@himdel This might break when you are switching groups for the user.
Would have to test this a little bit.

himdel:
@AparnaKarve found another issue .. the missing return would cause every reload to go to:

1. send an authentication request
1. go to /login in the mean time
1. when the reponse to the request was recieved, go to /dashboard

And the other thing it going to dashboard before the `/api` features request is responded to would break because the state has no navFeatures yet - so far I've wrapped the check in a helper so that it doesn't error, but I'm looking into forcing it to wait for that request to complete before we even start going to dashboard.

himdel:
@miq-bot add_label wip

AparnaKarve:
@himdel Not sure what commit#4 exactly addresses... and perhaps we don't need it if we do this -

```javascript
Session.loadUser()
          .then(function() {
            if (Session.activeNavigationFeatures()) {
              $state.go('dashboard');
            } else {
              Session.privileges_error = true;
              $state.go('login');
            }
          });
		  event.preventDefault();
		  return;
      }
```

himdel:
It's the `Could not go to Dashboard` error you get when you remove that commit, go to non-dashboard and reload the page :) .. because that `$state.navFeatures.something` condition fails because `navFeatures` is undefined. 

I originally thought it was because the response returns too late but it seems you're actually copying `navFeatures` only into the current state at that point .. which is not always dashboard.

AparnaKarve:
Can you try the above code snippet, it should fix the ```Could not go to Dashboard``` error.
The ```event.preventDefault();``` before the ```return``` will address the error.
I'm pretty positive we do not need the last commit :)

himdel:
I don't know how to apply that, it's either longer that it should (the last 3 lines are outside of that block) or there are some missing parentheses .. what should be the actual change?

AparnaKarve:
In your earlier commit 17ae2dcd130f3263714d8fd82e840c171773d338 (commit#2), just add ```event.preventDefault();``` before the ```return;```

himdel:
But no.. that transition to login is needed there because without it, you won't get redirected to login when you have no saved tokens. And that return is also needed because without it, you will go to login even if you should go somewhere else.

It may have the side effect of forcing `$state` to mean dashboard when you run `setRBAC` but that's not a solution

himdel:
Oh, ok, that makes sense!

himdel:
@AparnaKarve thanks, that fixes it! :)

OK.. so.. un-wip-ifying :)

AparnaKarve:
OK to merge. This should resolve the redirect issue.

PR #27, opened 2016-03-21
SDUI: Initial Blueprints nav. tab

dtaylor113:
Just the Blueprints navigational menu item/tab and minimal supporting 'state' files.

### DIFF
--- a/client/app/config/navigation.config.js
+++ b/client/app/config/navigation.config.js
@@ -32,6 +32,12 @@
             state: 'marketplace',
             icon: 'fa fa-copy',
             tooltip: N_('The total number of available catalog items'),
+          },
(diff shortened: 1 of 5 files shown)


AparnaKarve:
@dtaylor113 To view the Blueprints nav menu for now, just add -
`blueprints:  {show: true}` in `setRBACForNavigation` in `client/app/services/session.service.js`

dclarizio:
Merging this now so @dtaylor113 can move on with further PRs.  The Blueprints tab will not actually show up yet in the SSUI as it is being developed with mock data until the back end is implemented.

@h-kataria just a reminder that we will need to figure add an RBAC feature to enable this along with the temporary config setting.

PR #30, opened 2016-03-31
Fix extraneous auto-refresh API calls

eclarizio:
This will unbind and rebind window listeners so they don't accumulate and cause strange behavior.

This also fixes an issue where if none of the dialogs respond to an auto refresh, the API call was made with no dialogs to refresh, which caused issues. In this instance, the API call doesn't even need to be made so that saves a hit.

I think this may be related to quite a few BZs, but I don't think this fixes the whole problem. There appears to be another issue on the manageiq side involving the API and the data that is being returned.

https://bugzilla.redhat.com/show_bug.cgi?id=1303169

### DIFF
--- a/client/app/services/dialog-field-refresh.service.js
+++ b/client/app/services/dialog-field-refresh.service.js
@@ -1,3 +1,4 @@
+/* jshint -W117 */
 (function() {
   'use strict';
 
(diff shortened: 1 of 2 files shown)


gmcculloug:
:+1:

PR #32, opened 2016-04-04
Install bower, repo name in bower.json to resolve travis failures

AparnaKarve:
Install bower components with -
```
npm install -g gulp bower
```
to resolve travis failures

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ cache:
   - node_modules
 before_install:
 - bundle install
-- bower install -F
+- npm install -g gulp bower
 notifications:
   webhooks:
     urls:
--- a/bower.json
+++ b/bower.json
@@ -1,5 +1,5 @@
 {
-  "name": "ManageIQ Self Service",
+  "name": "manageiq-ui-self_service",
   "version": "0.0.1",
   "description": "ManageIQ Self Service",
   "authors": [


Fryguy:
I have a feeling that bower was in the cache, but not in the Travis image, so somehow the `npm install bower` command was removed under the assumption it "was already there".  Something must have been whacked with the cache, possibly.

Fryguy:
So, @AparnaKarve, might be best to just do `npm install -g bower gulp`, and remove the `which` check.  Also, can you make it the very first thing in the before_install section?

.travis.yml
```diff
@@ -6,7 +6,9 @@ cache:
   directories:
   - node_modules
 before_install:
+- npm install -g bower gulp
 - bundle install
+- npm install
```
Fryguy: This should not be here as the default install does this already

Fryguy:
The callback stack on travis is:

```yaml
before_install: # default: nothing
install:        # default: `npm install`
before_script:  # default: nothing
script:         # default: `npm test`
```

So, I think the best way to do this is:

```yaml
before_install:
- bundle install
- npm install -g gulp bower
before_script:
- bower install -F
```

Fryguy:
Looking a little closer, `npm install` runs `bower install -F` and the gulp stuff automatically, so some of that can go away

I think the minimum you need for this PR is the following, which also coincidentally aligns with what's in the README.md already.

```yaml
before_install:
- bundle install
- npm install -g gulp bower
```

AparnaKarve:
@Fryguy I think, this can now be merged.

himdel:
I think it's ready :)

PR #33, opened 2016-04-04
Shopping cart

himdel:
Shopping cart for service catalog items

Images:

Add to shopping cart under Service Catalogs > item detail
sc-add-button

Notification that item was added (stays on the same screen)
sc-added

Shopping cart modal when empty
sc-empty

Shopping cart modal with some items (updated, twice)
sc_full

Top right bar shopping cart indicator - empty and full
sc_icon_empty
sc_icon_count

still missing: 

   * [x] API changes - https://github.com/ManageIQ/manageiq/pull/7971
   * [x] RBAC - show shoppign cart icon only if `svc_catalog_provision`
   * order history - separate PR


shopping_cart.txt
```diff
@@ -0,0 +1,15 @@
+Myself and Dan had a meeting with Tina and GM yesterday abt shopping cart backend work
```
himdel: TODO remove

himdel:
@epwinchell , @serenamarie125 .. could you please look at the images if it's OK?

epwinchell:
@himdel I tend to use "btn-danger" very sparingly. I'd suggest "btn-default" instead for "reset" and "delete". Also, please use "pficon-delete" instead of "fa-trash-o". I also, think we need some styling between cart items.

himdel:
@epwinchell thanks, fixed both btn-danger and the icon.. Any tips on how to style the rows?  I was hoping to use pf-data-list, but it seems like it can't compare items based on identity, so it looked like a pain to implement deletion (or generate ids), but maybe I can rewrite that once I add persistence and actually have some ids.

epwinchell:
@himdel I'll will look around.

epwinchell:
@himdel How about: https://www.patternfly.org/widgets/#list-group ?

himdel:
> list-group 

Works for me :) I'll add it in the next push..

serenamarie125:
I agree with @epwinchell on using list-group as well as using pficon-delete.  Can you tag me when you have that available?  Would love to see what it looks like and possibly suggest something similar for PatternFly.  Thanks!

himdel:
@serenamarie125 , @epwinchell I updated to use list-group (updated the relevant screenshot), so from that point of view it should be ready..

Still some API changes needed before merging - right now, I can't remove a service request, nor update it to set process=true... (looking into it..)

client/app/states/marketplace/details/details.state.js
```diff
@@ -113,11 +113,14 @@
...
+        name: vm.serviceTemplate.name,
+
         dialogUrl: dialogUrl,
         serviceTemplateId: serviceTemplate.id,
+
         data: {
           action: 'order',
-          resource: dialogFieldData,
+          resource: _.cloneDeep(dialogFieldData),
         },
```
AparnaKarve: @himdel Would it be possible to inject `lodash` and call `cloneDeep` via `lodash`?
I believe that is what is done in other places, so just to keep things uniform...

client/app/services/shopping-cart.service.js
```diff
@@ -31,6 +32,14 @@
       notify();
     }
 
+    function removeItem(item) {
+      state.items = _.filter(state.items, function(i) {
```
AparnaKarve: Also here. Can you call `.filter` via `lodash` ?

client/app/services/shopping-cart.service.js
```diff
@@ -19,6 +19,12 @@
     };
 
     var persistence = {
+      theCard: function() {
+        // api/service_orders?filter[]=state=cart
```
AparnaKarve: probably a typo... `theCard` or `theCart`?

AparnaKarve:
Overall, the UI looks great - but you do need to take RBAC into account.
For e.g. -
 (a) do not display the shopping cart icon 
(b) do not display `Add to shopping cart` button 
if the user is not entitled for that feature

AparnaKarve:
@dclarizio @himdel I will resume testing once the APIs are in.

himdel:
Thanks @AparnaKarve, I'll fix the `lodash` thing, good catch! (Though, I *may* do a `s/lodash/_/` PR later, because that's how we're using it in the main UI.)

As for RBAC.. do we actually have a feature for that? Or should I add one? Or maybe use the same feature that shows/hides the section with the "add to cart" buttons? (That would be `svc_catalog_provision`, I think.)

AparnaKarve:
@himdel But don't you think `lodash.someThing()` reads better than `_.someThing()` ?

Yes, I think we should use the same feature id that lets us order services, which would be `svc_catalog_provision`

client/app/services/shopping-cart.service.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    function reload() {
+      CollectionsApi.query('service_requests', {
+        expand: 'resources',
+        filter: [ 'process=false' ],
+      })
+      .then(function(response) {
+        console.log('reload', response);
+
```
AparnaKarve: Maybe a leftover `console.log` here.

dclarizio:
@himdel doesn't `Add to Cart` replace `Submit Request` as the new way to order a Service?  I'm thinking we don't need the submit button anymore.

himdel:
@dclarizio Could be, I'll remove it :)

himdel:
Should be ready now, removing WIP...

@AparnaKarve could you re-review please? (You will need ManageIQ/manageiq#7971.)

AparnaKarve:
@himdel This works great for the most part. One of the things that I noticed could be API related.
Here's what I found so far -

- Some Service Catalogs could not be ordered and instead gave this error message -
screen shot 2016-05-06 at 4 24 52 pm

If it's a valid error message, then the Service Catalog is not orderable and perhaps should not be even shown in the list. (just thinking out loud here)

- When I order a service catalog, I see a 404 in the console
```
Ordering error
angular.js:11630 GET http://localhost:3001/api/service_orders/cart/service_requests?expand=resources 404 (Not Found)
```

- (Minor) The console.log messages like `Item added to shopping cart` should be removed

himdel:
@aparnakarve what's generating the first error is the automate validation code.. So it should be orderable, just not with these exact values..  But yeah I'd rather we could run some validation during, not after, but.. Api limitation :-)
And the same goes for that 404, when you just ordered, the shipping cart doesn't exist anymore and needs to be created again, hence the 404.

As for the console output, I'll check, but I think Notification is doing that at the same time it displays it to the user...

PR #39, opened 2016-04-19
Console support under My Services

himdel:
This adds support for showing a VNC or SPICE consoles under My Services... (And openstack remote_url consoles.)

Depends on https://github.com/ManageIQ/manageiq/pull/8306 (and https://github.com/ManageIQ/manageiq/pull/8378 for openstack).

Screenshot (one machine is off, one is on but doesn't support consoles, and the other is on and does, so has an extra button):

console1

Spice console: 
console_spice

VNC console:
console_vnc

Openstack console:
console_openstack

TODO

   * ~~devel-mode ws proxy in server/app.js~~ - will do in a separate PR - https://github.com/ManageIQ/manageiq-ui-self_service/pull/44
   * [x] move to service

Note that without a proxy in the node app, these changes can only work on the appliance.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -25,9 +25,9 @@
   "strict"        : false,    // true: Requires all functions run in ES5 Strict Mode
   "maxparams"     : 10,       // {int} Max number of formal params allowed per function
   "maxdepth"      : 4,        // {int} Max depth of nested blocks (within functions)
-  "maxstatements" : 30,       // {int} Max number statements per function
-  "maxcomplexity" : 10,       // {int} Max cyclomatic complexity per function
-  "maxlen"        : 140,      // {int} Max number of characters per line
+  "maxstatements" : 32,       // {int} Max number statements per function
+  "maxcomplexity" : 12,       // {int} Max cyclomatic complexity per function
+  "maxlen"        : 160,      // {int} Max number of characters per line
 
   // Relaxing
   "asi"           : false,     // true: Tolerate Automatic Semicolon Insertion (no semicolons)
--- a/bower.json
+++ b/bower.json
@@ -38,7 +38,8 @@
     "patternfly": "~2.3.0",
     "sprintf": "~1.0.3",
     "toastr": "~2.1.1",
-    "no-vnc": "~0.5.1"
+    "no-vnc": "~0.5.1",
+    "spice-html5-bower": "^0.1.6"
   },
   "overrides": {
     "patternfly": {
@@ -64,6 +65,9 @@
       ]
     },
     "no-vnc": {
+      "main": []
+    },
+    "spice-html5-bower": {
       "main": []
     }
   }
--- a/bower.json
+++ b/bower.json
@@ -37,7 +37,8 @@
     "ngstorage": "~0.3.10",
     "patternfly": "~2.3.0",
     "sprintf": "~1.0.3",
-    "toastr": "~2.1.1"
+    "toastr": "~2.1.1",
+    "no-vnc": "~0.5.1"
   },
   "overrides": {
     "patternfly": {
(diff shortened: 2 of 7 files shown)


himdel:
@abellotti We're doing `GET /api/services/10000000000304?attributes=picture,picture.image_href,evm_owner.name,miq_group.description,vms,aggregate_all_vm_cpus,aggregate_all_vm_memory,aggregate_all_vm_disk_count,aggregate_all_vm_disk_space_allocated,aggregate_all_vm_disk_space_used,aggregate_all_vm_memory_on_disk,actions,custom_actions,provision_dialog,service_template` to get the VMs under that service.

I tried adding a virtual column to VM .. like this..

```diff
diff --git a/app/models/vm_or_template.rb b/app/models/vm_or_template.rb
index 734f4ea..407fe36 100644
--- a/app/models/vm_or_template.rb
+++ b/app/models/vm_or_template.rb
@@ -166,6 +166,7 @@ class VmOrTemplate < ApplicationRecord
   virtual_column :v_annotation,                         :type => :string,     :uses => :hardware
   virtual_column :has_rdm_disk,                         :type => :boolean,    :uses => {:hardware => :disks}
   virtual_column :disks_aligned,                        :type => :string,     :uses => {:hardware => {:hard_disks => :partitions_aligned}}
+  virtual_column :supports_console,                     :type => :boolean
 
   virtual_has_many   :processes,              :class_name => "OsProcess",    :uses => {:operating_system => :processes}
   virtual_has_many   :event_logs,                                            :uses => {:operating_system => :event_logs}
@@ -1870,6 +1871,10 @@ class VmOrTemplate < ApplicationRecord
     false
   end
 
+  def supports_console
+    console_supported?('spice') || console_supported('vnc')
+  end
+
   # Return all archived VMs
   ARCHIVED_CONDITIONS = "vms.ems_id IS NULL AND vms.storage_id IS NULL"
   def self.all_archived
```

And while it works if I ask for the vm specifically, with `?attributes=supports_console`, I can't seem to find a way to get that virtual column in the first request... tried adding `vms.supports_console` to attributes, but it seems to not work with subcollections. Am I missing something?

abellotti:
You are correct, the vms being returned in the above query is done via the "vms" virtual attribute (i.e. the virtual_has_many :vms reflection on the service).  The virtual attribute added to the vms would be returned when asking for it as ?attributes=supports_console only applies to the service being queries. 

possible alternatives:
- list of vms needs to be queried a second time (with querying supports_console virtual attribute, or even exposing the action to the vms for request_console to be dynamic in nature, like we do for the "reconfigure" action on services, we only show that if the service as a Reconfigure resource action)
- expose vms as subcollections of services, so adding to above query --expand=vms, but we can fetch those in the API layer and additionally getting supports_console

possible others too.

himdel:
OK, I'd prefer not to have to add another API call (because we'd need to do it for every service we open in SSUI), so I guess the second possibility would be better (if I understand correctly that exposing it as a subcollection means `expand=vms` + that `attributes=...vms.supports_console` would cause the service get to return what I want)...

Could you take care of it please?

imtayadeway:
> expose vms as subcollections of services, so adding to above query --expand=vms, but we can fetch those in the API layer and additionally getting supports_console

This seems to be the way to go.

@himdel do you need all the services returned, regardless if they support console? If not, you may be able to filter on it here with `vms.supports_console=true`

himdel:
@imtayadeway you mean all the vms? In that case, yes I need all of them :) (it's a screen with the details of a particular service, including all the vms, and a button to open the console for all the vms that support it)

himdel:
@miq-bot add_label darga/yes

himdel:
The only thing missing now is a devel-mode proxy to proxy the websocket requests to the right place, similar to what's being done for /api requests.. Will add those in a separate PR.

A way to test on devel is to open the console and tweak the `?host` and `?port` parameters in the query string to point to the right place.

himdel:
@miq-bot remove_label wip

PR #40, opened 2016-04-22
Reload - go to original state, not dashboard

himdel:
Previously, `rbacDashboardOrLogin` - would redirect to dashboard when everything is fine - that breaks expectations of what reloading means, *and* is terrrible for development. So this PR makes reload work as expected.

@romanblanco you mind find this useful ;)

@AparnaKarve mind reviewing? I *think* it shouldn't really break RBAC that much, but if we really want to check everything, I *can* make RBAC checks happen on any state change (and associate the required features with each state)..

### DIFF
--- a/client/app/config/authorization.config.js
+++ b/client/app/config/authorization.config.js
@@ -88,17 +88,19 @@
       });
 
       Session.loadUser()
-        .then(rbacDashboardOrLogin)
+        .then(rbacReloadOrLogin(toState, toParams))
         .catch(badUser);
     }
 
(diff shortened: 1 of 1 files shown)


AparnaKarve:
@himdel Yeah, we need this...to be able to retain the original state at reload. 👍 
Code looks fine, will try it out in a bit.

AparnaKarve:
Works well!
@dclarizio OK to merge and can be cherry-picked for darga as well.

kbrock:
Thanks Martin. This is nice

PR #45, opened 2016-05-04
Fix language code returned by ManageIQ rest API

mzazrivec:
When setting user language is self-service, we rely on user's language preference as returned
by the rest api. In the case of language variants (such as zh_CN), the rest api returns the language
name with underscores (i.e. as 'zh_CN'), while the catalogs as returned from translators
contain a dash (i.e. 'zh-CN').

This mismatch would cause the self-service application not to be able to find the chinese translation
catalog.

@himdel

### DIFF
--- a/client/app/config/gettext.config.js
+++ b/client/app/config/gettext.config.js
@@ -10,9 +10,9 @@
     gettextCatalog.debug = false;
 
     gettextCatalog.loadAndSet = function(lang) {
-      gettextCatalog.setCurrentLanguage(lang);
-
       if (lang) {
+        lang = lang.replace('_', '-');
+        gettextCatalog.setCurrentLanguage(lang);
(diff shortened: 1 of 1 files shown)


client/app/config/gettext.config.js
```diff
@@ -10,6 +10,7 @@
     gettextCatalog.debug = false;
 
     gettextCatalog.loadAndSet = function(lang) {
+      lang = lang.replace('_','-');
```
himdel: `lang` will be `null` when used with older version of the API .. might either want to remove that `if (lang)` if we don't support it anymore, or do the replace only when not null :).

Also, not that replace replaces just one occurence, not all of them.. (But still no idea what the API would send for Han simplified, so maybe we'll end up needing a lookup table eventually anyway.)

mzazrivec:
@himdel Fixed, thanks.

mzazrivec:
@himdel @dclarizio Review / merge?

PR #46, opened 2016-05-05
Shopping cart - order history

himdel:
This is a follow-up to #33, adding Order history to My requests (under the actual request list right now).

Screenshots:

My Requests with added tabs..
oh-requests

My Orders..
oh-orders

Order detail..
oh-order-detail

(the individual requests have a click-through to a request detail screen)

TODO in a separate `darga/no` PR:
   * rename "My Requests" as "My Orders and Requests"
   * put the "My Orders" tab first
   * remember the last chosen tab (separate url for each?)
   * remove the `updated_at || placed_at` logic

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -25,7 +25,7 @@
   "strict"        : false,    // true: Requires all functions run in ES5 Strict Mode
   "maxparams"     : 10,       // {int} Max number of formal params allowed per function
   "maxdepth"      : 4,        // {int} Max depth of nested blocks (within functions)
-  "maxstatements" : 32,       // {int} Max number statements per function
+  "maxstatements" : 64,       // {int} Max number statements per function
   "maxcomplexity" : 12,       // {int} Max cyclomatic complexity per function
   "maxlen"        : 160,      // {int} Max number of characters per line
 
--- a/client/app/components/request-list/request-list.directive.js
+++ b/client/app/components/request-list/request-list.directive.js
@@ -1,0 +1,28 @@
+(function() {
+  'use strict';
+
+  angular.module('app.components')
+    .directive('requestList', RequestListDirective);
+
+  /** @ngInject */
+  function RequestListDirective() {
+    var directive = {
+      restrict: 'E',
+      scope: {
+        'items': '<',
+        'config': '<?',
+      },
+      templateUrl: 'app/components/request-list/request-list.html',
+      controller: RequestListController,
+      controllerAs: 'vm',
+      bindToController: true
+    };
+
+    return directive;
+
+    /** @ngInject */
+    function RequestListController() {
+      var vm = this;
+    }
+  }
+})();
--- a/client/app/resources/collections-api.factory.js
+++ b/client/app/resources/collections-api.factory.js
@@ -58,6 +58,9 @@
       options = options || {};
 
       if (options.expand) {
+        if (angular.isArray(options.expand)) {
+          options.expand = options.expand.join(',');
+        }
         params.push('expand=' + options.expand);
       }
 
--- a/client/app/states/requests/list/list.state.js
+++ b/client/app/states/requests/list/list.state.js
@@ -51,11 +51,7 @@
     vm.requestsList = angular.copy(vm.requests);
 
     vm.orders = orders.resources;
-    vm.orderListConfig = {
-      showSelectBox: false,
-      selectionMatchProp: 'id',
-      onClick: handleOrderClick,
-    };
+    vm.ordersList = angular.copy(vm.orders);
 
     if (angular.isDefined($rootScope.notifications) && $rootScope.notifications.data.length > 0) {
       $rootScope.notifications.data.splice(0, $rootScope.notifications.data.length);
@@ -68,6 +64,12 @@
       onClick: handleRequestClick,
     };
 
+    vm.orderListConfig = {
+      showSelectBox: false,
+      selectionMatchProp: 'id',
+      onClick: handleOrderClick,
+    };
+
     vm.toolbarConfig = {
       filterConfig: {
         fields: [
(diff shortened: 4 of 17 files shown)


himdel:
@dclarizio it *should* be ready.. not sure how many details we want in the service order detail screen, etc.. but the main functionality is done. And I've put "My Orders" second now, so that we can switch the order only for Euwe..

dclarizio:
@AparnaKarve can you please review?

AparnaKarve:
@himdel Still need to do a detailed review, but overall looks great! :)

Tried running it in the UI, and couple of things that I noticed -

- Not sure if I'm dealing with bad data in my DB here, but I'm seeing orders with 0 items. 
Even if it's a bug, it would be API related.
screen shot 2016-06-28 at 12 16 31 pm

screen shot 2016-06-28 at 12 16 46 pm

- And maybe a few width adjustments needed...
screen shot 2016-06-28 at 12 14 47 pm

client/app/resources/collections-api.factory.js
```diff
@@ -58,6 +58,9 @@
       options = options || {};
 
       if (options.expand) {
+        if (angular.isArray(options.expand)) {
+          options.expand = options.expand.join(',');
+        }
```
AparnaKarve: So you're converting an Array to a String here. Is that necessary for the API to work correctly if we pass multiple `expand` parameters?
`expand: ['resources', 'service_requests']`

client/app/components/request-list/request-list.directive.js
```diff
@@ -0,0 +1,28 @@
...
+  angular.module('app.components')
+    .directive('requestList', RequestListDirective);
+
+  /** @ngInject */
+  function RequestListDirective() {
+    var directive = {
+      restrict: 'E',
+      scope: {
+        'items': '<',
+        'config': '<?',
```
AparnaKarve: What does this exactly do?
```javascript
'items': '<',
'config': '<?',
```

AparnaKarve:
Couple more things that need to be considered in this PR or a follow-up PR -

1. Remember the last active tab
2. Display Order Date in the "My Orders" list
3. Make the "My Orders" page look more like the "My Requests" page for consistency

client/app/resources/collections-api.factory.js
himdel: well, it's not strictly necessary in that we could just pass `expand: 'resources,service_requests'`, but it's to make it more consistent with `attributes` and `decorators`, which already accept arrays..

client/app/components/request-list/request-list.directive.js
himdel: It's the one way binding thing, like `=` and `=?` but doesn't let you change it from the directive.. essentially it takes that `items="vm.something"` and binds that to that directive's controller's `.items`

himdel:
> 0 items

You should probably clear your db, looks like some remnants from experimenting with that shoping cart PR ;).

> Display Order Date in the "My Orders" list

It's being displayed, but you need https://github.com/ManageIQ/manageiq/pull/9478 to actually have it in the db..

> Make the "My Orders" page look more like the "My Requests" page for consistency

Do we actually want filters and sorting for my orders? I'm trying hard to think about what you'd want to filter or search by ... since all the orders are just "Order #123456".. that's why I didn't add the toolbar..

> Remember the last active tab

Yeah, that'd make sense, I'll have to play with the router a bit for that, but yup, will do.. (=> back to [WIP])

himdel:
> > Remember the last active tab
> Yeah, that'd make sense, I'll have to play with the router a bit for that, but yup, will do.. (=> back to [WIP])

@AparnaKarve actually, this will have to be done in a different PR, the currently used version of angular-ui-bootstrap doesn't support setting active tab properly.. so I'll add this bit in the darga/no PR, after we merge the update-all-the-libs PR..

AparnaKarve:
> Do we actually want filters and sorting for my orders? I'm trying hard to think about what you'd want to filter or search by ... since all the orders are just "Order #123456".. that's why I didn't add the toolbar..

hmm, I beg to differ on that 😉 
I think a filter to search by `Order Date` will be very handy once we have the list display the date the order was placed.
And for that matter, even a filter for `Status` of the order will be useful.

himdel:
@AparnaKarve OK, ok, I'll add the toolbar :)

But status filter won't be that useful if the only status is `ordered` ;)

himdel:
The data migration must be darga/no, so it was split of in https://github.com/ManageIQ/manageiq/pull/9539 .. which means past orders won't have a date in darga, until the user updates to the newer release..

himdel:
Added the toolbar (+ filtering & sorting), and also added the `(placed_at || updated_at)` logic everywhere, so that we do use `updated_at` when `placed_at` is nil..

@AparnaKarve can you re-review please? :) (especially since adding order history was `+226, -45` and the toolbar is `+202, -65` on top of that :))

himdel:
Keeping the version before rebasing on new patternfly in [order-list-darga](https://github.com/himdel/manageiq-ui-self_service/tree/order-list-darga) branch..

client/app/states/requests/list/list.state.js
```diff
@@ -43,7 +43,7 @@
   }
 
   /** @ngInject */
-  function StateController($state, requests, orders, RequestsState, $filter, $rootScope, lodash) {
+  function StateController($state, requests, orders, RequestsState, OrdersState, $filter, $rootScope, lodash) {
```
AparnaKarve: This is a generic comment ... not for this PR per se.
What I've noticed is, the inject list keeps on growing...I think the linter will warn after a cap of 10 is reached. 
Do you think at that point we should consolidate the parameters and then pass them as arrays?
For e.g. Consolidate all states in a `States` Array

client/app/states/services/list/list.state.js
```diff
@@ -238,15 +238,15 @@
     }
 
     function matchesFilter(item, filter) {
-      if ('name' === filter.id) {
+      if (filter.id === 'name') {
```
AparnaKarve: Didn't know something called `Yoda conditions` existed
https://en.wikipedia.org/wiki/Yoda_conditions
Good to know :)

But looks like the constant portion should be on the left side?... and you flipped it?
(A yoda condition places the constant portion of the expression on the left side of the conditional statement)

.jshintrc
```diff
@@ -25,7 +25,7 @@
   "strict"        : false,    // true: Requires all functions run in ES5 Strict Mode
   "maxparams"     : 10,       // {int} Max number of formal params allowed per function
   "maxdepth"      : 4,        // {int} Max depth of nested blocks (within functions)
-  "maxstatements" : 32,       // {int} Max number statements per function
+  "maxstatements" : 64,       // {int} Max number statements per function
```
AparnaKarve: So looks like this is the place to bend the rules to make tests pass? ;)

.jshintrc
```diff
@@ -25,7 +25,7 @@
   "strict"        : false,    // true: Requires all functions run in ES5 Strict Mode
   "maxparams"     : 10,       // {int} Max number of formal params allowed per function
```
AparnaKarve: And this is what I was referring to earlier -
`"maxparams"     : 10`
Most of our `ngInjects` are dangerously close to 10

chessbyte:
@himdel Running `git cherry-pick -x -m 1 04c386a` on darga results in a merge conflict.

Please make a PR against Darga branch. /cc @dclarizio

.jshintrc
himdel: Yup O:-) .. but this one should be a warning, not something that makes the tests fail.. as soon as I'm done making miq_bot lint JS too, I'd like to remove `jshint` and `jscs` from the SSUI tests and just have miq_bot run it..

client/app/states/services/list/list.state.js
himdel: Love the name I do, but like them I don't. So flipped they get :)

client/app/states/requests/list/list.state.js
himdel: Well, I'm kinda hoping to "solve" this by not having the test fail when that happens :).. 

As for making the list smaller, well, you *could* write a factory that gets injected a bunch of dependencies and returns them wrapped in an object.. I wonder how useful that'd be, maybe it does make sense for commonly injected combinations..

himdel:
Darga PR created .. #76

chessbyte:
Backported to Darga via #76

PR #47, opened 2016-05-05
Request - fix request_state vs approval_state in detail, use clock not spinner for pending

himdel:
In My Requests, an icon is used for each approval_state (approved, pending_approval, denied) - but a spinner was used for pending_approval, which looks misleading because a spinner is usually used to indicate something is loading and does not quite convey the meaning of "waiting for somebody". - updated to use `fa-clock-o` instead.

Also, in the detail, the same icons are used, but were used for request_state, not approval_state, which has completely different states - fixed so that both use approval_state, and request_state has no icon.

@epwinchell thanks for the clock icon idea :)


himdel:
@miq-bot add_label bug

PR #50, opened 2016-05-10
SDUI: Blueprint Details, Multi-Delete

dtaylor113:
New Blueprint Details modal, used for Create and [edit] Details:

image

New Multi-Delete modal:

image

Plus:
- New Mock Data Factory
- Misc updates to blueprint list mock data (last_modified, # nodes, visibility, catalog name)
- Updates to Designer Toolbox Tabs

### DIFF
--- a/client/app/resources/mock-data.factory.js
+++ b/client/app/resources/mock-data.factory.js
@@ -1,0 +1,432 @@
+(function() {
+  'use strict';
+
+  angular.module('app.resources')
+    .factory('MockData', MockDataFactory);
+
+  /** @ngInject */
+  function MockDataFactory() {
+    var service = {
+      retrieveMockData: retrieveMockData
+    };
+
+    return service;
+
+    function retrieveMockData(collection) {
+
+      if(collection == 'service_catalogs'){
+        return { "resources": [
+          {
+          "href": "http://localhost:8001/api/service_catalogs/1000000000007",
+          "id": 1000000000007,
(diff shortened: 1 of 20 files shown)


client/app/resources/mock-data.factory.js
```diff
@@ -0,0 +1,431 @@
...
+  /** @ngInject */
+  function MockDataFactory() {
+    var service = {
+      retrieveMockData: retrieveMockData
+    };
+
+    return service;
+
+    function retrieveMockData(collection) {
+      if (collection === 'service_catalogs') {
```
h-kataria: @dtaylor113 instead of mocking this here you could have used existing REST calls to fetch service_templates and same for list of service_dialogs.

dtaylor113: Hi, I do in fact have those REST api's in place in the code, but I choose to mock them anyway because I do not have a working database in my dev environment.  To switch to REST api's, just need to remove the 'mock' option in the CollectionsApi call, ex:

```
function resolveServiceCatalogs(CollectionsApi) {
   var options = {
   mock: true,       <---- remove this
   expand: 'resources',
   sort_by: 'name',
   sort_options: 'ignore_case'};

   return CollectionsApi.query('service_catalogs', options);
}
```

dclarizio:
I'm going to merge this with the API calls mocked for now.  @h-kataria can we get Dave a database with some useful records so that he can turn on the API calls in his next PR?

PR #51, opened 2016-05-12
Fix Services and Requests count in dashboard

AparnaKarve:
Requests count would now be derived from 3 Request promise arrays for `Pending`, `Approved` and `Denied`.
Each promise array would comprise of 2 promises for the two types of Request -
`ServiceReconfigureRequest` and `ServiceTemplateProvisionRequest`

This change is mainly done due to the API design limitation on how it handles the logical `OR` operation.

https://bugzilla.redhat.com/show_bug.cgi?id=1333942

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -33,7 +33,7 @@
     "angular-ui-router": "~0.2.15",
     "font-awesome": "~4.4.0",
     "jquery": "~2.1.0",
-    "lodash": "~3.9.0",
+    "lodash": "~4.12.0",
     "moment": "~2.10.0",
     "ngprogress": "~1.1.2",
     "ngstorage": "~0.3.10",
--- a/client/app/states/dashboard/dashboard.state.js
+++ b/client/app/states/dashboard/dashboard.state.js
@@ -34,6 +34,11 @@
   }
 
   /** @ngInject */
+  // TODO API in question: /api/requests
+  // TODO with filterValues = ['type=ServiceReconfigureRequest', 'or type=ServiceTemplateProvisionRequest', 'approval_state=pending_approval'];
+  // TODO API OR-ing bug/"design limitation" has forced an implementation change in how we gather count data for Requests
+  // TODO One API call would now be split into two - one for ServiceTemplateProvisionRequest and other for ServiceReconfigureRequest
+
   function resolvePendingRequests(CollectionsApi, $state) {
     if (!$state.navFeatures.requests.show) {
       return undefined;
--- a/client/app/states/dashboard/dashboard.state.js
+++ b/client/app/states/dashboard/dashboard.state.js
@@ -145,23 +145,17 @@
     return CollectionsApi.query('services', options);
   }
 
-  function chainRequestPromises(promiseArray, vm, type) {
-    var count = 0;
-    if (promiseArray.length > 0) {
-      promiseArray[0].then(function success(data) {
-        count = data.subcount;
-        promiseArray[1].then(function success(data) {
-          count += data.subcount;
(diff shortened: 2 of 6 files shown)


AparnaKarve:
@himdel Please review

client/app/states/dashboard/dashboard.state.js
```diff
@@ -129,24 +173,31 @@
...
+        vm.servicesCount.current = vm.servicesCount.total - vm.servicesCount.retired - vm.servicesCount.soon;
       }
 
       vm.servicesFeature = true;
     }
 
     vm.requestsFeature = false;
-    if (angular.isDefined(pendingRequests)) {
+
+    if (angular.isDefined(pendingRequests) &&
```
himdel: Shouldn't this be `||` instead of `&&`? I assume we still want the counts even if only one of them is defined...

client/app/states/dashboard/dashboard.state.js
```diff
@@ -129,24 +173,31 @@
...
-      vm.requestsCount.total = pendingRequests.subcount + approvedRequests.subcount + deniedRequests.subcount;
-      vm.requestsCount.pending = pendingRequests.subcount;
-      vm.requestsCount.approved = approvedRequests.subcount;
-      vm.requestsCount.denied = deniedRequests.subcount;
+      vm.requestsCount.total = 0;
+
+      var allRequests = [pendingRequests, approvedRequests, deniedRequests];
+      var allRequestTypes = ['pending', 'approved', 'denied'];
+
+      lodash.times(3, function(n) {
```
himdel: It isn't immediately clear the the 3 is the number of things in the array, maybe something like `allRequests.forEach(function(promises, idx) { ...`?

Or you could make it an object, like `{ pending: pendingRequests, ... }` and iterate over `_.pairs(obj)`?

client/app/states/dashboard/dashboard.state.js
```diff
@@ -90,34 +130,38 @@
...
+    var options = {expand: false, filter: ['service_id=nil'] };
 
     return CollectionsApi.query('services', options);
   }
 
-  /** @ngInject */
-  function resolveServicesWithDefinedServiceIds(CollectionsApi, $state) {
-    if (!$state.navFeatures.services.show) {
-      return undefined;
+  function chainRequestPromises(promiseArray, vm, type) {
```
himdel: Might prefer to use `$q.all` here, so that we don't hardcode the numer of promises in the array...

Something along the lines of

```js
$q.all(promiseArray).then(function(values) {
  var count = _.sumBy(values, 'subcount');
  vm.requestsCount[type] = count;
 vm.requestsCount.total += count;
});
```

AparnaKarve: Thanks @himdel -- `$q` was an elegant approach alright!
I also liked the idea of using `sumBy`...however, had to bump up the version of `lodash` to 4.12.0 to be able to use that.

bower.json
```diff
@@ -33,7 +33,7 @@
     "angular-ui-router": "~0.2.15",
     "font-awesome": "~4.4.0",
     "jquery": "~2.1.0",
-    "lodash": "~3.9.0",
+    "lodash": "~4.12.0",
```
himdel: We'll need to check all the lodash calls, there may be some incompatibilities between the major versions..

himdel: ..went through the changelog, only `assign`, `merge` and `pick` *may* be affected...

himdel: .. aaand went through the code, and it seems we're using all of these the right way, so no problems :).

PR #56, opened 2016-05-24
lodash needs to be downgraded to be compatible with angular-patternfly

AparnaKarve:
lodash should be downgraded to v3.9.0 to be compatible with `"angular-patternfly": "~2.3.2"`.

It was upgraded in https://github.com/ManageIQ/manageiq-ui-self_service/pull/51/commits/d2813dfe0a14478d5299942d360cf34d87805f4a for the `.sumBy` call.
The `.sumBy` call has now been replaced with `.sum` which is supported by lodash v3.9.0.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -33,7 +33,7 @@
     "angular-ui-router": "~0.2.15",
     "font-awesome": "~4.4.0",
     "jquery": "~2.1.0",
-    "lodash": "~4.12.0",
+    "lodash": "~3.9.0",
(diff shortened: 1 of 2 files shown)


AparnaKarve:
@himdel Please review. Thanks.

h-kataria:
verified in UI :+1:

himdel:
Yup, looks good :) :+1: 

We just need to remember to convert it back when finally moving to lodash v4.

PR #58, opened 2016-05-25
Default sorting filters should always be applied when the list is loaded

AparnaKarve:
When the `Services` or `Requests` list is loaded,
(a) the list should always have the default sorting filters applied
(b) number of items in the list must match the number displayed in the Dashboard or the side navigation bar


AparnaKarve:
@h-kataria Please review/test. Thanks.

h-kataria:
:+1:

PR #60, opened 2016-06-01
List items should be checked against the actual filter value for filtering to work correctly

AparnaKarve:
List items should be checked against the actual filter value as opposed to the displayed Filter value String in the UI

https://bugzilla.redhat.com/show_bug.cgi?id=1322934


AparnaKarve:
@h-kataria Please review.

h-kataria:
verified fix in UI

PR #64, opened 2016-06-15
Simplify creation of blueprint delete confirmation message

mzazrivec:
1. we got rid of __('?'), __('blueprint?') which would be kind of weird to translate
2. we don't construct the confirmation message in two places
3. we reduced the amount of code :-)


mzazrivec:
@AparnaKarve Review please?

AparnaKarve:
@mzazrivec Looks great.

@dclarizio OK to merge.

PR #65, opened 2016-06-16
Fix build step not copying console dependencies and translations

### LINKED ISSUE #61: "The page you were looking for doesn't exist" while trying to open a console (on the appliance)
The url `/bower_components/spice-html5-bower/spiceHTML5/spice_auto.html` doesn't exist, because the file (and its novnc counterpart) are not being copied over to `manageiq/public/self_service` as part of `gulp build`.

@miq-bot assign himdel

himdel:
Right now, neither the gettext translations nor the html templates for spice & vnc consoles are properly copied during the `gulp build` step, so neither actually works in the appliance now.

This fixes `gulp gettext-copy` by fixing the path selector (one more dir level now), and adds `gulp console-copy` to copy the console templates and their dependencies to the right place.

Also adjusted the console code to find the template using the full path, including `/self_service` and updated the devel-mode server to actually cope with such a path too.

Assuming `darga/yes` since without this, consoles and i18n won't work in SSUI..

Fixes #61 
https://bugzilla.redhat.com/show_bug.cgi?id=1347312
https://bugzilla.redhat.com/show_bug.cgi?id=1347319

@AparnaKarve could you review please?

### DIFF
--- a/client/app/services/consoles.service.js
+++ b/client/app/services/consoles.service.js
@@ -78,7 +78,7 @@
     }
 
     function openSpice(results) {
-      var url = '/bower_components/spice-html5-bower/spiceHTML5/spice_auto.html' +
+      var url = '/self_service/bower_components/spice-html5-bower/spiceHTML5/spice_auto.html' +
         '?host=' + $location.host() +
         '&port=' + $location.port() +
         '&path=' + results.url +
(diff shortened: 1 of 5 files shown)


AparnaKarve:
Successfully verified in SSUI - both BZs are resolved with these changes. 👍

PR #66, opened 2016-06-17
npm-shrinkwrap - force gulp-header to 1.8.2 for gulp-angular-templatecache

himdel:
this is to fix tracker1/gulp-header#37:
  * gulp-angular-templatecache depends on gulp-header 1.x
  * gulp-header 1.8.3 tries to lstat before writing and fails

which leads `gulp templatecache` to fail with ENOENT on the file it's trying to write to.. :)

@simaishi , @Fryguy ..

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,8 @@
+{
+  "dependencies": {
+    "gulp-header": {
+      "version": "1.8.2",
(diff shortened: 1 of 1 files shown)


Fryguy:
:+1: LGTM  cc @chessbyte This is marked blocker as builds are broken without it.

PR #68, opened 2016-06-24
Add a spinner for dialog field refreshes

eclarizio:
This simply shows/hides a patternfly icon while the ajax transaction for refreshing a dialog field is happening so that the user can tell something is going on.

### DIFF
--- a/client/app/services/dialog-field-refresh.service.js
+++ b/client/app/services/dialog-field-refresh.service.js
@@ -24,6 +24,12 @@
           }
         });
 
+        allDialogFields.forEach(function(dialogField) {
+          if (_.includes(dialogFieldsToRefresh, dialogField.name)) {
+            dialogField.beingRefreshed = true;
+          }
+        });
+
         if (dialogFieldsToRefresh !== []) {
           refreshMultipleDialogFields(allDialogFields, dialogFieldsToRefresh, url, resourceId);
         }
@@ -45,6 +51,7 @@
         Notifications.error('There was an error refreshing this dialog: ' + result);
       }
 
+      dialogField.beingRefreshed = true;
       fetchDialogFieldInfo(allDialogFields, [dialogField.name], url, resourceId, refreshSuccess, refreshFailure);
     }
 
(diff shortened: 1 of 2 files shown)


dclarizio:
@himdel please review

client/app/services/dialog-field-refresh.service.js
```diff
@@ -24,6 +24,14 @@
           }
         });
 
+        angular.forEach(allDialogFields, function(dialogField) {
+          angular.forEach(dialogFieldsToRefresh, function(fieldToRefreshName) {
+            if (dialogField.name === fieldToRefreshName) {
+              dialogField.beingRefreshed = true;
+            }
+          });
+        });
```
himdel: Maybe we don't have to go through all the fields, and use something like `lodash.find(allDialogFields, { name: fieldToRefreshName })` ?

himdel:
I think it looks good :) but @eclarizio I'd rather you replaced that double nested foreach with only one foreach + lodash.find ..

eclarizio:
@himdel Tried using a `lodash.find` but it didn't exist. Looking at the documentation, it was using `_.find`, so I used that with an `_.includes`. Didn't know we had access to those lodash functions, good call.

himdel:
Hehe, yeah we have lodash in both ssui and the ops ui, really useful sometimes :)

As for `lodash` vs `_` .. well, at some point I want to make SSUI use `_` for consistency, so.. no problem there.. But how it's done right now in SSUI is that you indeed use `lodash` but need to inject it first..

(As in `function DialogFieldRefreshFactory(CollectionsApi, Notifications, lodash) {`)

client/app/services/dialog-field-refresh.service.js
```diff
@@ -24,6 +24,12 @@
           }
         });
 
+        _.find(allDialogFields, function(dialogField) {
```
himdel: Wait, this is wrong.. `find` is supposed to return the first item for which the callback returns true..

You sure you didn't mean something like.. ?

```js
dialogFieldsToRefresh.forEach(function(name) {
  _.find(allDialogFields, { name: name }).beingRefreshed = true;
});
```

eclarizio: Ah, well, I don't actually care about the results of the find, so this actually still works in the UI. The anonymous function that's setting the `beingRefreshed` property is all that matters, though I agree this does look a little odd and should probably be more like your second one.

PR #74, opened 2016-07-01
Add Turkish translation

mzazrivec:
It's 100% translated.


chessbyte:
@dclarizio darga label?

PR #79, opened 2016-07-14
Fix IE11 - CollectionsApi - URI-escape right-side values in query string

himdel:
This makes `CollectionsApi` escape all its query-string arguments - without this, you can get an url like

    /api/services?filters[]=service_id=nil&filter[]=retires_on>=2016-07-14&filter[]=retires_on<=2016-08-13

instead of

    /api/services?filters[]=service_id%3Dnil&filter[]=retires_on%3E%3D2016-07-14&filter[]=retires_on%3C%3D2016-08-13

And while the first url works just fine in chrome, it does return a 400 for IE11, meaning it's impossible to log in from IE..

@dclarizio we *may* want a BZ and darga/yes for this, since we *are* supposed to support IE11 in darga, right?

### DIFF
--- a/client/app/resources/collections-api.factory.js
+++ b/client/app/resources/collections-api.factory.js
@@ -21,7 +21,7 @@
         return MockData.retrieveMockData(collection);
       } else {
         return $http.get(url + buildQuery(options), buildConfig(options))
-            .then(handleSuccess);
+          .then(handleSuccess);
       }
 
       function handleSuccess(response) {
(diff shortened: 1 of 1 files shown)


chessbyte:
@himdel Running `git cherry-pick -x -m 1 7d161bc` on darga causes merge conflict.  Please resolve and make  a PR against darga.  /cc @dclarizio

himdel:
@chessbyte darga PR in https://github.com/ManageIQ/manageiq-ui-self_service/pull/95

chessbyte:
Backported to Darga via https://github.com/ManageIQ/manageiq-ui-self_service/pull/95

PR #80, opened 2016-07-14
Language switcher

himdel:
This adds a language switcher component, both in the user's menu, and on the login screen.

i18n-switch-login

The login screen defaults to `User Default` meaning whatever the user has saved, but any choice is saved for the next time.

i18n-switch-menu

The menu switcher changes the language immediately, and saves..

The names of the languages get generated from the translations, so while testing the PR, short names (en, fr, es, ..) may appear instead.. The screenshots were done using this `available_languages.json`: `
{"de-DE":"Deutsch","es":"Español","fr":"Français","ja":"日本語","ru":"ру́сский язы́к","sk":"Slovensky","tr":"Türkçe","zh-CN":"中文","zh-Hans-CN":"汉族"}`

WIP:
   * [x] load a list of available languages from the json (https://github.com/ManageIQ/manageiq-ui-self_service/pull/77)
   * [x] a menu component showing the list of languages, and switching
   * [x] the same for login screen
   * [x] save the settings via the API (https://github.com/ManageIQ/manageiq/pull/9801)
   * [x] load the settings from the API (when "User Default" selected)
   * [x] detect the browser-default language
   * [x] pick the right language when browser-default (so that `browserPreferred=[ "fr-FR" ]; available=[ "fr" ]` actually matches `"fr"`) (same logic as [fastgettext](https://github.com/grosser/fast_gettext/blob/master/lib/fast_gettext/storage.rb#L130))
   * [x] commit cleanup, etc.
   * [x] trigger a state reload after switching - so that one-time bound translate strings can work too
   * [x] switch language immediately on login screen
   * [x] going back to login screen should reset the language (to English or browser-default)
   * [x] bug: on my services page, switching from something to English hangs the tab
   * [x] unexpanded selectpickers keep the last translation's version (but the options are fine)

https://bugzilla.redhat.com/show_bug.cgi?id=1356254
https://bugzilla.redhat.com/show_bug.cgi?id=1356256

### DIFF
--- a/client/app/components/language-switcher/language-switcher.directive.js
+++ b/client/app/components/language-switcher/language-switcher.directive.js
@@ -20,7 +20,7 @@
     return directive;
 
     /** @ngInject */
-    function LanguageSwitcherController(gettextCatalog, Language, lodash) {
+    function LanguageSwitcherController($scope, gettextCatalog, Language, lodash) {
       var vm = this;
       vm.mode = vm.mode || 'menu';
 
@@ -34,6 +34,14 @@
         delete hardcoded._user_;
       } else {
         vm.chosen = Language.chosen = { code: '_user_' };
+
+        $scope.$watch('vm.chosen.code', function(val) {
+          if (!val || (val === '_user_')) {
+            val = "en";
+          }
+
+          Language.setLocale(val);
+        });
       }
 
       Language.ready
--- a/client/app/components/language-switcher/language-switcher.directive.js
+++ b/client/app/components/language-switcher/language-switcher.directive.js
@@ -8,7 +8,9 @@
   function LanguageSwitcherDirective() {
     var directive = {
       restrict: 'AE',
-      scope: {},
+      scope: {
+        mode: '@?',
+      },
       templateUrl: 'app/components/language-switcher/language-switcher.html',
       controller: LanguageSwitcherController,
       controllerAs: 'vm',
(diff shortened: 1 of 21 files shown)


himdel:
@mzazrivec could you re-test the hang when switching to english please? The last commit *might* fix it but unfortunately I can't reproduce..

himdel:
Updated to fix the cached sort-by field name in stateservices on controller reinit - to show the name in the right language afterwards..

@mzazrivec should be ready, hopefully.. ;)

mzazrivec:
@himdel The following scenario still doesn't seem to work:
1. I login with a french locale
2. in the dashboard screen, I switch to english
3. click on my services
4. note the sort by field is still in french (rather than in english)

himdel:
Fixed the sort field (pfSort would cache the old instance, so updating *that* now, instead of replacing), *and* the login screen now updates the logo when the language changes..

mzazrivec:
@chessbyte Merge into Darga please? I'd like to update the gettext catalog in darga afterwards
for the additional translations to be made.

himdel:
Looks like it won't cherry-pick cleanly - created https://github.com/ManageIQ/manageiq-ui-self_service/pull/91

chessbyte:
Backported to Darga via https://github.com/ManageIQ/manageiq-ui-self_service/pull/91

PR #83, opened 2016-07-15
Add Coveralls Badge

chessbyte:


Fryguy:
We don't have coverage, so can't have coveralls.

PR #87, opened 2016-07-19
Update node requirement to 6.x

himdel:
Previously, we would depend on node 0.12 and (officially) npm 2 (except it wouldn't work for most cases).

The JS ecosystem is slowly but surely moving towards ES6, supported starting in node4 (but poorly).

This pushes the dependency up to node version 6..

(Should not need `npm install -g npm` everywhere either anymore.)


Fryguy:
@himdel Do you have any changes to make for the build process in manageiq-appliance-build?

himdel:
Yes, coming shortly..

bdunne:
@himdel Is there a corresponding update to [the developer setup guide](https://github.com/ManageIQ/guides/blob/master/developer_setup.md)?

himdel:
@bdunne I'm not sure one is needed, given fresh enough fedora (unfortunately that means fc25 now), the instructrions will give you the right version.. (but tbh node 4 still works just fine).

Alternatively, we could update those to install [nvm](https://github.com/creationix/nvm) and use that to pick the right version.

PR #73, opened 2016-07-01
Update patternfly & angular-patternfly dependencies to 3.*

himdel:
Updated patternfly & angular-patternfly to 3.5.1, minor updates to some dependencies..

Removed old bower overrides, and unused moment.js.

   * [x] check changes in patternfly 2.3.0 -> 3.5.1
   * [x] check changes in angular-patternfly 2.3.2  -> 3.5.1
   * separate PR:
      * update angular-bootstrap to 1.* (angular-patternfy forces 0.13)
      * same for lodash
   * [x] check list-view item styling changes .. good enough :)
   * [x] check select default item behaviour changes (schedule service retirement)
   * [x] 3.5.1 -> 3.6.0

PR #97, opened 2016-08-03
[DARGA] Fix missing languages in login screen language switcher

himdel:
The list of available languages is pulled asynchronously, and when it's downloaded only after the language switcher select has been populated, angular updates the <select> element, but angular-patternfly 2.* won't update the visible selectpicker after.

Thus, sometimes, there are no languages in the selector.

This manually triggers a refresh of that selectpicker after the list is downloaded and the <select> is populated.

(This is not needed in master because angular-patternfly 3.* already does the right thing - #73)

https://bugzilla.redhat.com/show_bug.cgi?id=1363753

### DIFF
--- a/client/app/components/language-switcher/language-switcher.directive.js
+++ b/client/app/components/language-switcher/language-switcher.directive.js
@@ -20,7 +20,7 @@
     return directive;
 
     /** @ngInject */
-    function LanguageSwitcherController($scope, gettextCatalog, Language, lodash, $state) {
+    function LanguageSwitcherController($scope, gettextCatalog, Language, lodash, $state, $timeout, jQuery) {
       var vm = this;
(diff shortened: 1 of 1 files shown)


himdel:
@dclarizio .. this should probably get in today..

chessbyte:
View from Darga branch:
```text
$ git log
commit 488abd7cc662d8b459e928329febc52aeedb61b4
Merge: a669412 fdbe11e
Author: Oleg Barenboim <chessbyte@gmail.com>
Date:   Tue Aug 23 17:12:53 2016 -0400

    Merge pull request #97 from himdel/darga-switcher-pf-select-fix
    
    [DARGA] Fix missing languages in login screen language switcher
```

Issue #121, opened 2016-09-09
Migrate from jshint to eslint


AllenBW:
This single issue will encompass all work included in the migration.
Reasons for this change:

1. JSHint team has joined with ESLint
2. Customizability
3. Plugin support (principally https://github.com/Gillespie59/eslint-plugin-angular)
4. ENV support  (amd, jquery, es6)

The sum of tasks to be completed in the fulfillment of this issue (each task encompasses at least one pr, we're going to go slow on this effort, as to avoid any unnecessary messes):
- [x] Add eslint rules and dependencies (extending present state)
- [x] Separate specs from codebase
- [x] Apply eslint changes by rule type across app, a pr for each (large changes) or a few (many small changes) will be made
- [ ] Add eslint task to gulp & travis
- [ ] Remove totality of jshint remnants 

@chriskacerguis feed me (assign) this one, really excited to introduce the ✋ of :goberserk: into our codebase

chriskacerguis:
Assigned!

PR #122, opened 2016-09-09
Adds eslint dependencies and preliminary rules

AllenBW:
Extending jshint, includes angular specific best practices and check as well.

I know everyone is busy, but ideally this is more of a conversation than a pr @himdel (if you have a sec) or anyone in general :)  

It is known that we are using airbnb eslint elsewhere: https://github.com/ManageIQ/manageiq/blob/62186766366eae2e93ba1300abe85fc3d8377482/package.json  so the question has to be asked, are we ok doing something different here? (semi rhetorical question I believe yes, because architecturally this is a different beast)

#121 is one bullet closer to complete with this pr


chriskacerguis:
@Fryguy any issues here?

chriskacerguis:
@himdel any thoughts / issues here?

AparnaKarve:
@AllenBW Am I correct in my understanding that this would be more handy in the future when we upgrade to Angular 2 and as a result we would be coding more in `es6`?

AllenBW:
@AparnaKarve heeeeeeeey you noticed that :P yeah es6 is one of the style envs we're targeting with this, the thought being that the forward leaning will be of great benefit the future.  In general though the move to using eslint, stronger code style checks is of great benefit now.  To no single persons fault ( 👊 collective) ssui is wee bit messy.

himdel:
I'd be definitely for keeping this eslint and manageiq's eslint config similar, or as close to similar as makes sense. But you're right this is a different beast, so.. :)

PR #124, opened 2016-09-09
Isolate unit tests

AllenBW:
Relocate .spec files to `/tests`, reconfigure gulp config to support change

another notch in #121, this task removes the work of adding the mountain of code style exceptions that would be support the `.spec`s interwoven with the app 

@chriskacerguis 🌏


chriskacerguis:
👍

himdel:
What's the darga and euwe status of this PR?

chriskacerguis:
@himdel do we need the euwe tag?  I believe that this was in there before the merge, and so it should have it in there.  I'll add the Darga tag

himdel:
Aah, ok, could be, thanks :) adding euwe/yes to #192 then :)

PR #128, opened 2016-09-13
Resolves the issue of spec files being included in index.html

AllenBW:
Resolves the issue of spec files being included in index.html
Updates spec.html

@chriskacerguis 🉑


chriskacerguis:
👍

PR #131, opened 2016-09-14
Tagging of Blueprints and Items on Canvas

dtaylor113:
Implemented Tagging of Blueprints:

image

Implemented Tagging of Items on Canvas:

image

Reverted eslint changes for Blueprint files.

I originally had this PR as two smaller ones, but ran into rebase/merge issues with the eslint changes, so had to fix everything in one PR.

PR #134, opened 2016-09-15
Add Notifications Drawer and Toast Notifications List

jeff-phillips-18:
This PR replaces the use of inline notifications and the Angular Patternfly Notifications service with a Notifications Drawer, Toast List, and EventNotifications service.

The inline messages are now shown as Toasts in the upper right section of the window and are then also available in the Notifications Drawer until removed by the user. The notifications shown in the drawer are for the current session only, there is no saving/retrieval of past notifications or notifications from other sessions.

Toasts:
image

Notifications Drawer:
image

@serenamarie125 @beanh66 @chriskacerguis @dtaylor113

### DIFF
--- a/client/app/components/notifications/notification-body.html
+++ b/client/app/components/notifications/notification-body.html
@@ -37,7 +37,7 @@
     </div>
     <div class="col-md-4 col-sm-3" ng-if="notificationGroup.notificationType == 'task'">
       <div class="drawer-pf-notification-info expanded-info" ng-click="customScope.markRead(notification, notificationGroup)">
-        <span class="info-title">Started: </span>
+        <span class="info-title" translate>Started: </span>
         <span class="date">{{notification.data.startTime | date:'MM/dd/yyyy'}}</span>
         <span class="time">{{notification.data.startTime | date:'h:mm:ss a'}}</span>
       </div>
@@ -49,7 +49,7 @@
     </div>
     <div class="col-md-4 col-sm-3" ng-if="notificationGroup.notificationType == 'task'">
       <div class="drawer-pf-notification-info" ng-click="customScope.markRead(notification, notificationGroup)">
-        <span class="info-title">Completed: </span>
+        <span class="info-title" translate>Completed: </span>
         <span class="date">{{notification.data.endTime | date:'MM/dd/yyyy'}}</span>
         <span class="time">{{notification.data.endTime | date:'h:mm:ss a'}}</span>
       </div>
--- a/client/app/components/notifications/notification-body.html
+++ b/client/app/components/notifications/notification-body.html
@@ -1,0 +1,69 @@
+<div ng-if="!drawerExpanded" ng-click="customScope.markNotificationRead(notification, notificationGroup)">
+  <div class="dropdown pull-right dropdown-kebab-pf">
+    <button class="btn btn-link dropdown-toggle" type="button" ng-click="customScope.clearNotification(notification, notificationGroup)">
+      <span class="pficon pficon-close"></span>
+    </button>
+  </div>
+  <span ng-if="notification.type" class="pull-left {{customScope.getNotficationStatusIconClass(notification)}}" ng-click="customScope.markRead(notification, notificationGroup)"></span>
+  <span class="drawer-pf-notification-message notification-message"
+        ng-click="customScope.markRead(notification, notificationGroup)"
+        tooltip-popup-delay="500" tooltip-placement="bottom" tooltip="{{notification.message | translate}}">
+    {{notification.message | translate}}
+  </span>
+  <div ng-if="!notification.data.inProgress" class="drawer-pf-notification-info" ng-click="customScope.markRead(notification, notificationGroup)">
+    <span class="date">{{notification.timeStamp | date:'MM/dd/yyyy'}}</span>
+    <span class="time">{{notification.timeStamp | date:'h:mm:ss a'}}</span>
+  </div>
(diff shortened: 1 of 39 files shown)


dtaylor113:
Please merge this after #131 is merged!  Some of the blueprint files were reverted in #131, so are outdated in this PR.  -thanks

client/app/components/notifications/notification-body.html
```diff
@@ -0,0 +1,69 @@
...
+      <span ng-if="notification.type" class="pull-left {{customScope.getNotficationStatusIconClass(notification)}}" ng-click="customScope.markRead(notification, notificationGroup)"></span>
+      <span class="drawer-pf-notification-message notification-message"
+            ng-click="customScope.markRead(notification, notificationGroup)"
+            tooltip-append-to-body="true" tooltip-popup-delay="500" tooltip-placement="bottom" tooltip="{{notification.message | translate}}">
+            {{notification.message | translate}}
+      </span>
+    </div>
+    <div class="col-md-4 col-sm-3" ng-if="notificationGroup.notificationType == 'task'">
+      <div class="drawer-pf-notification-info expanded-info" ng-click="customScope.markRead(notification, notificationGroup)">
+        <span class="info-title">Started: </span>
```
AllenBW: no translate attribute on this span?

client/app/components/notifications/notification-body.html
```diff
@@ -0,0 +1,69 @@
...
+      </div>
+      <div ng-if="notification.inProgress" class="progress">
+        <div class="progress-bar progress-bar-info"
+             ng-style="{width: notification.data.percentComplete + '%'}" tooltip="{{notification.data.percentComplete}} + {{'% Complete' | translate}}">
+        </div>
+      </div>
+    </div>
+    <div class="col-md-4 col-sm-3" ng-if="notificationGroup.notificationType == 'task'">
+      <div class="drawer-pf-notification-info" ng-click="customScope.markRead(notification, notificationGroup)">
+        <span class="info-title">Completed: </span>
```
AllenBW: Also this one, translate attribute?

jeff-phillips-18:
Corresponding Jira Stories:

Design: https://patternfly.atlassian.net/browse/MGMTUX-179
Initial Dev: https://patternfly.atlassian.net/browse/MGMTUX-180
Final Dev: https://patternfly.atlassian.net/browse/CFUX-92

jeff-phillips-18:
@AllenBW I looked at replacing toastr as well but I agree it's better left to another PR since the fields are slightly different and I didn't want to get into it all at once.

dtaylor113:
Hi Jeff, if you have to update this PR, could you remove from client/app/states/designer/blueprints/list/list.state.js:
```
/* This notification 'splice' code doesn't work.  Splice needs a third argument, the items to splice in
     * Not sure what this code is trying to accomplish, but it exists in login, request list, & services list
    if (angular.isDefined($rootScope.notifications) && $rootScope.notifications.data.length > 0) {
      $rootScope.notifications.data.splice(0, $rootScope.notifications.data.length);
    }
*/
```
If it is no longer relevant.  If not, I'll remove it in my next PR.  -thanks

dtaylor113:
LGTM :+1:

serenamarie125:
From a UX perspective, this is good to go!

client/app/components/notifications/notification-body.html
jeff-phillips-18: Do we prefer the translate attribute or the {{"some string" | translate}} or does it matter?

Also, is there a date format translator? That is something better than
  {{notification.timeStamp | date:'MM/dd/yyyy'}}
or
  {{notification.timeStamp | date:'h:mm:ss a'}}

That will format based on locale?

dtaylor113:
For reviewers, try out the new github review process.  Click on "Files Changed" above, then on the upper right "Review Changes" green button.

client/app/components/notifications/notification-body.html
AllenBW: This is a darn good question, I don't know if it matters, i like the idea of only the strings though, looping in @himdel for some opinions here..

chriskacerguis:
@jeff-phillips-18 looks like this needs to be rebased

chriskacerguis:
👍

client/app/components/notifications/notification-body.html
himdel: I'd say prefer translate attribute *if* you can sanely use it, but go for `{{""|translate}}` if you'd have to create elements just to be able to do the attribute :).

As for l10n, we simply currently don't support it. Would be nice though.. And IIRC we do have `moment` in there, with locale support, so maybe we could make a moment filter and set the appropriate moment locale when changing the active language .. and get at least a rudimentary l10n that way... OTOH, I'd say it's better to be consistently wrong than being inconsistently right, so.. if our date picker can't handle different formats reasonably, it'd be more suprising to have one format in edit and one in view than to force the US way in both..

client/app/states/designer/blueprints/list/list.state.js
```diff
@@ -86,13 +86,6 @@
       visibilityNames.push(item.name);
     }
 
-    /* This notification 'splice' code doesn't work.  Splice needs a third argument, the items to splice in
```
himdel: @jeff-phillips-18 Jeff, this code does work, splice doesn't need more than 2 arguments... it clears the array without destroying any references :).

You can think of it as $rootScope.notifications.data = []; except you can be sure that if you've done foo = $rootScope.notifications.data before, then even foo will be empty (and still point to the same array).

PR #150, opened 2016-09-19
Add Delete Profile and View Profile Details features

jeff-phillips-18:
This PR adds functionality to allow the user to delete a profile and the ability to drill down to see more details on the profile.

image

image

@serenamarie125 @beanh66 @chriskacerguis

### DIFF
--- a/client/app/states/designer/profiles/details/details.state.js
+++ b/client/app/states/designer/profiles/details/details.state.js
@@ -1,0 +1,59 @@
+(function() {
+  'use strict';
+
+  angular.module('app.states')
+    .run(appRun);
+
+  /** @ngInject */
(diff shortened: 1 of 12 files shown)


client/app/states/designer/profiles/details/details.state.js
```diff
@@ -0,0 +1,59 @@
...
+      return null;
+    }
+  }
+
+  /** @ngInject */
+  function StateController(profile, sprintf, ProfilesState) {
+    var vm = this;
+
+    function setTitle() {
+      vm.title = sprintf(__("View Profile: %s"), profile.name);
```
jeff-phillips-18: This allows different locales to format the string differently, some languages might have the name before "View Profile".

PR #152, opened 2016-09-19
Updated gulp-clean-css to latest version, 1.0.0 doesn't even exist

AllenBW:
Minor change, when switching the depreciated dependency I overlooked updating the version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "gulp-jscs": "^4.0.0",
     "gulp-jshint": "^2.0.1",
     "gulp-load-plugins": "^1.3.0",
-    "gulp-clean-css": "^1.0.0",
+    "gulp-clean-css": "^2.0.12",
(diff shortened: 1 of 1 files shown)


chriskacerguis:
@AllenBW so, I merged this, but why isn't our testing catching this stuff?

AllenBW:
@chriskacerguis catching what? we don't test any of the gulp tasks and don't have a dependency version hound in travis. nothing to catch here really

chriskacerguis:
I guess, the better question is what **should** the tests fail on.

AllenBW:
@chriskacerguis I would say, tests (excluding style checks and other assisting hooks) should fail whenever functionality of the app deviates from ta previously established nominal operation.  Tests should describe the entirety of actions we expect users to perform and should be seen as a companion to any new functionality moving forward.

PR #155, opened 2016-09-19
added CodeClimate config with a link eslintrc

chriskacerguis:
added CodeClimate config with a link eslintrc


chriskacerguis:
@AllenBW @AparnaKarve can you all please take look here.

AparnaKarve:
@chriskacerguis Was reading up the doc here https://docs.codeclimate.com/docs/eslint
Do we need `channel: "eslint-3"` ?

chriskacerguis:
@AparnaKarve I think the suggestion version was fine, but that brings up the fact we are trying to get everything "current" (and I didn't check the rules @AllenBW put in) so I think that would be a good idea.

chriskacerguis:
@AparnaKarve @himdel you all ok with this?

himdel:
Sure, my only concern is the rename of `eslintrc` - the name without the extension is considered deprecated..

chriskacerguis:
@himdel updated.  Waiting for checks to pass.

PR #159, opened 2016-09-20
Snake case all ui_properties

jntullo:
Updating the values of chartDataModel that are passed to or come from the API to be chart_data_model. This will make the `ui_properties` of a Blueprint more consistent with only `snake_case` variables. 

I did a test of Blueprints including updating / creating Blueprints with these new values (note that I did update all of my existing blueprints in the DB to use `chart_data_model`). Would appreciate an additional smoke test @dtaylor113 :)

### DIFF
--- a/client/app/components/blueprint-details-modal/blueprint-details-modal-service.factory.js
+++ b/client/app/components/blueprint-details-modal/blueprint-details-modal-service.factory.js
@@ -227,7 +227,7 @@
     }
 
     function disableOrderListTabs() {
-      return vm.blueprint.ui_properties.chartDataModel.nodes.length <= 1;
+      return vm.blueprint.ui_properties.chart_data_model.nodes.length <= 1;
     }
 
     function toggleActionEqualsProvOrder() {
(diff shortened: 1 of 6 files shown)


dtaylor113:
Hi, I think we only need to change it in two places:

Once when we get it from API:

client/app/states/designer/blueprints/editor/editor.state.js
```
function StateController($state, $stateParams, blueprint, blueprintTags, serviceTemplates) {
    var vm = this;
    vm.title = 'Blueprint Designer';
    if (blueprint) {
      ...
      vm.blueprint = blueprint;
      vm.blueprint.ui_properties.chartDataModel = blueprint.ui_properties.chart_data_model;
      // if we don't want two copies floating around
      delete(vm.blueprint.ui_properties.chart_data_model);
    }
```

And another time when we set it via the API

client/app/services/blueprints-state.service.js

function getBlueprintPostObj(tmpBlueprint) { 
...
    blueprintObj.ui_properties.chart_data_model = chartDataModel;
...

Would this work?  It involves far less changes to the code base.

dtaylor113:
Just did a smoke test, everything looks fine :-)

PR #180, opened 2016-09-23
Update gettext catalog for master

mzazrivec:
Euwe will be addressed in a different pull request.


mzazrivec:
Marking this as wip, since I've noticed problems in the way the developers mark the strings for
translations (some strings are messed up).

PR #181, opened 2016-09-23
Update gettext catalog euwe

mzazrivec:
* updated zanata.xml
* update catalogs for euwe


mzazrivec:
Marking this as wip, since I've noticed problems in the way the developers mark the strings for
translations (some strings are messed up).

PR #185, opened 2016-09-26
Add substitute filter

mzazrivec:
In certain situations, it's not possible to correctly annotate strings for translation when these
strings are a value of an html element and contain a variable interpolation. For example:

```html
<span tooltip="{{item.v_total_vms}} VMs">
```

This is where the `substitute` filter comes in handy. The above situation would then be resolved as:

```html
<span tooltip="{{ '[[number]] VMs'|translate|substitute:{number: item.v_total_vms} }}">
```
i.e. `[[` and `]]` mark start and end of a variable to be substituted, the actual values would be
then substituted with the context of the `substitute` filter.

With the substitute filter, the string with interpolation contained in an element attribute can:
* be properly substituted
* be properly marked for translation

The pull request also fixes the above problem in couple of html files.

### DIFF
--- a/client/app/components/notifications/notification-footer.html
+++ b/client/app/components/notifications/notification-footer.html
@@ -11,8 +11,8 @@
        ng-class="{'disabled': !notificationGroup.notifications || notificationGroup.notifications.length === 0}"
        confirmation
        confirmation-if="true"
-       confirmation-title="{{'Clear All ' + notificationGroup.heading |translate}}"
-       confirmation-message="{{'Are you sure you would like to clear all ' + notificationGroup.heading + '?'|translate}}"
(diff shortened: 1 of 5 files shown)


mzazrivec:
@AllenBW Yup, I'll also update the documentation, once we agree whether this is good approach, etc.

chessbyte:
Euwe Backport details:
```sh
$ git log
commit 532501b81512a9e1687a1068340708caf1af68fb
Date:   Mon Sep 26 13:01:10 2016 -0500

    Merge pull request #185 from mzazrivec/add_substitute_filter
    
    Add substitute filter
    (cherry picked from commit 889706d25e9267365f81e22b7938f3c38aafb644)
```

mzazrivec:
FYI, the documentation / guides PR for the filter: https://github.com/ManageIQ/guides/pull/156

PR #197, opened 2016-09-28
Enhancement/81 html resource load

### LINKED ISSUE #81: Move resources loading from index.html
The [client/index.html](https://github.com/ManageIQ/manageiq-ui-self_service/blob/master/client/index.html) file includes every JS file and every package we use in SSUI, and everytime some adds a new JS file or package, it quite often causes conflicts in others PR that are also doing some adding of new resources.

It would be nice to have loading of these resources separated from the `index.html` file, to not force other contributors to rebase their PRs everytime someone adds something new to `index.html`.
I'm not sure what would be the best way to do so though.

If somebody has any idea for this, I can try to do the work once I have nothing more with high priority.

@himdel

AllenBW:
closes #81 :feelsgood: 🚑 

So this PR the SUI switching from passing a dependency loaded `index.html` to sideloading (via ejs) two dependency partials at runtime. In the process it was necessary to yugely refactor `server/.app.js` as it was making a mess all over the place 🌋 😷 😲 

I expect in this pr you'll notice no difference from master, except maybe a shorter app boot time and no more mess merge nonsense with `index.html` (hah because its gone, jk its still there, but changed).
- `gulp serve-dev` should behave as expected
- `gulp build` should behave as expected
- app proxy with miq should behave as expected

SO @himdel @dtaylor113 @dclarizio @jeff-phillips-18 @AparnaKarve @chriskacerguis @chalettu 
since all ya'll were mildly involved in this, one way or another gonna humbly demand some 👀 on the branch when you get a chance.  Yes yes review the code, but also if you can pop on the branch and confirm all works as expected.

🙏


AllenBW:

AllenBW:
(ps we might have to talk about code climate running elsint on things not the app)

chriskacerguis:
@AllenBW well, let's start with the codeclimate fails :)

chriskacerguis:
First pass, it looks good to me.  @himdel @martinpovolny thoughts?

client/index.ejs
```diff
@@ -0,0 +1,66 @@
...
+      display: block !important;
+    }
+
+    .splash {
+      display: none;
+      background-color: #428bca;
+    }
+  </style>
+
+  <!-- build:css styles/lib.css -->
```
himdel: Why are these still needed here, since they're in styles.html as well? (Same for js below)

himdel:
@AllenBW `gulp build` doesn't create an `index.html` now - are you sure we want that?

AFAIK this won't work now, since we don't use node to serve it on the appliance yet, and I'm not sure that's something we should be doing in the production mode anyway.. I think the ejs -> html conversion should be a part of the build process too (why run EJS on every request when the file will never change..)

AllenBW:
@himdel ooo goooooooood catch on that!  see i knew waiting for your 👀 would be worth it <3

client/index.ejs
AllenBW: So these are still used by build process, we run wiredep over index.html, literally exactly the same as before.

there is a cleaner way to do it, kill that stuff use [gulp-ejs](https://www.npmjs.com/package/gulp-ejs) to inject those dependency partials, but i'll be honest... didn't want to put the time into figuring out which older version would jive with our node version and not cause issue.

if ya want @himdel i can make an outstanding issue to do that conversion whenever we update everything (the conversion of removing the build comments and relying wholly on ejs for dependency injection during the build process)

AllenBW: @himdel #198 <3

gulpfile.js
```diff
@@ -52,8 +55,9 @@ gulp.task('available-languages', task('available-languages'));
 /**
  * Build tasks
  */
-gulp.task('inject', ['wiredep', 'sass', 'templatecache'], task('inject'));
```
himdel: missed one... `gulp.task('serve-dev', [...'inject'...` ;)

himdel:
No further complaints :+1: :shipit: :)

jeff-phillips-18:
I pulled this branch, built, and ran. Everything worked as expected 👍

AllenBW:
@jeff-phillips-18 very much so appreciated all the same!

PR #201, opened 2016-09-30
Unit Tests for Blueprint List View

dtaylor113:
New stats:
```
PhantomJS 1.9.8 (Linux 0.0.0): Executed 79 of 79 SUCCESS (4.035 secs / 0.445 secs)

=============================== Coverage summary ===============================
Statements   : 38.02% ( 1808/4755 )
Branches     : 12.01% ( 181/1507 )
Functions    : 32.54% ( 412/1266 )
Lines        : 38.02% ( 1808/4755 )
```
Sorry, can't get old stats as I refactored too much in new branch.


dtaylor113:
Yay!  Finally resolved all the code climate issues :-)

.codeclimate.yml
```diff
@@ -8,7 +8,8 @@ engines:
     enabled: true
     config:
       languages:
-      - javascript
+        javascript:
```
chriskacerguis: This should be in a separate PR, and I'm not sure we are done discussing it.

tests/blueprints/blueprints-list.directive.spec.js
```diff
@@ -0,0 +1,692 @@
...
+
+      // Third row has 3 Items on Canvas, so publish button should be enabled
+      disabledPublishButton = angular.element(rows[2]).find('.btn.btn-default.disabled');
+      expect(disabledPublishButton.length).to.eq(0);
+    });
+  });
+
+  function getBlueprints() {
+    return [
+      {
```
chriskacerguis: Thinking out loud here, but would it be better to put this JSON data into a .json file, instead of putting it in code?

tests/blueprints/blueprints-list.directive.spec.js
```diff
@@ -0,0 +1,692 @@
...
+            "href": "http://localhost:8001/api/blueprints/10000000000026"
+          }
+        ]
+      }
+    ];
+  }
+
+  function getServiceCatalogs() {
+    return [{
+      "href": "http://localhost:8001/api/service_catalogs/10000000000001",
```
chriskacerguis: (same) Thinking out loud here, but would it be better to put this JSON data into a .json file, instead of putting it in code?

tests/blueprints/blueprints-list.directive.spec.js
dtaylor113: I like the idea, but it looks like it would take a little more work to implement:

http://stackoverflow.com/questions/17370427/loading-a-mock-json-file-within-karmaangularjs-test

.codeclimate.yml
dtaylor113: A separate PR with just this single change to  .codeclimate.yml?

tests/blueprints/blueprints-list.directive.spec.js
dtaylor113: I like the idea, but it looks like it would take a little more work to implement:

http://stackoverflow.com/questions/17370427/loading-a-mock-json-file-within-karmaangularjs-test

.codeclimate.yml
chriskacerguis: I'd like to merge this PR, but we are not done discussing the change yet in https://github.com/ManageIQ/manageiq-ui-self_service/issues/204

tests/blueprints/blueprints-list.directive.spec.js
chriskacerguis: @dtaylor113 seems like it would be the right thing to do overall?  thoughts?

dtaylor113: Ok, I can take some more time and look into implementing that.

PR #202, opened 2016-09-30
Implemented WebSocket token requesting mechanism for Notifications

skateman:

PR #211, opened 2016-10-03
[EUWE] Implemented WebSocket token requesting mechanism for Notifications

skateman:
This is an Euwe version of #202


himdel:
For the test-failure related discussion, see #222.

PR #212, opened 2016-10-03
updated mass_threshold 

### LINKED ISSUE #204: CodeClimate issues that should not block PR merge
CodeClimate sometimes reports duplicates that are valid, but that make little sense fixing in the PR they're reported in.

Example:  

Another example:  
https://github.com/ManageIQ/manageiq-ui-self_service/issues/201 - current failures

Gitter discussion:  
:point_up: [September 30, 2016 2:53 PM](https://gitter.im/ManageIQ/manageiq-ui-self_service?at=57ee7c600ff4ef7a50f36834)

We may either ignore them when needed (but possibly leads to code rot), disable the similar code rule (but possibly skips valid warnings), or...

This issue should be a discussion on how to handle that...

chriskacerguis:
closes #204

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -8,7 +8,8 @@ engines:
     enabled: true
     config:
       languages:
-      - javascript
+        javascript:
+          mass_threshold: 20
 ratings:
   paths:
   - client/app/**
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -9,7 +9,7 @@ engines:
     config:
       languages:
         javascript:
-          mass_threshold: 20
+          mass_threshold: 100
 ratings:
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -8,7 +8,8 @@ engines:
     enabled: true
     config:
       languages:
-      - javascript
+        javascript:
+          mass_threshold: 20
```
dtaylor113: Also, I'll take out my threshold change in #201

chriskacerguis: /sigh.  Thanks for catching that.  It's updated to 100

chriskacerguis:
@himdel can you please review / merge?

PR #213, opened 2016-10-03
Enhancement/141 use node sass

### LINKED ISSUE #141: Switch to node-Sass / libsass
I think it would be good to remove the dependency of ruby-sass and use node-sass?  @himdel @chalettu @AllenBW thoughts on this?

AllenBW:
closes #141 

Replaces ruby dependency (gulp-ruby-sass) with gulp-sass.

No visual changes result, build process is identical.

@chriskacerguis @dtaylor113 @jeff-phillips-18 if you're curious...

Not sure what to do about the hakiri error, ruby is no longer in the app so... is it necessary?


chriskacerguis:
@himdel @martinpovolny @AparnaKarve @chalettu thoughts?

/cc @Fryguy @dclarizio @chessbyte

gulp/config.js
```diff
@@ -232,20 +232,7 @@ module.exports = (function() {
...
-        bower + 'patternfly-sass/assets/stylesheets/',
-        // Hack so that font-awesome/variables can be found
-        client + 'assets/sass/'
-      ],
-      style: 'compact',
-      noCache: false,
-      compass: false,
-      bundleExec: true,
-      sourcemap: false,
+      outputStyle: 'compact',
```
chriskacerguis: do we want to use "compress" to minify the css?

chriskacerguis:
@Fryguy as this is no longer a "Ruby" app, thoughts about disabling "Hariri"

gulp/tasks/sass.js
```diff
@@ -17,9 +17,8 @@ module.exports = function(gulp, options) {
       browserSync.notify('Compiling Sass');
     }
 
-    return sass(config.src, config.options)
-      .pipe(plumber({errorHandler: options.onError}))
-      .on('error', options.onError)
+    return gulp.src(config.src)
+      .pipe(sass(config.options).on('error', sass.logError))
```
himdel: This changes how error handling works - previously, `errorHandler` from gulpfile.js would get called, respecting `--fail`, etc. Sass logError does something else, right? Might make sense to keep using `options.onError`?

AllenBW: can do 👍

AllenBW:
@chriskacerguis compress is the same as it was before, no issue changing though

chriskacerguis:
I think we should start to compress, also on that note are we minifying the JS?

@martinpovolny @himdel @chalettu thoughts?

gulp/config.js
himdel: Agreed on `compressed` - although the difference is negligible (before 412 KB, compact 504 KB, compressed 500 KB)

chriskacerguis:
FYI - Going to leave this open for comments until 9am Oct 4, 2016 CST.

himdel:
As for minifying JS, I think we should be able to do that without problems, but we're currently not - but maybe a good idea for a separate PR :) (Although, whoever is doing that, please make it compatible with babel :))

chriskacerguis:
@himdel Minify as separate PR = totally 👍

himdel:
@AllenBW `compress` != `compressed` ;) (https://github.com/sass/node-sass#outputstyle)

AparnaKarve:
Wouldn't minifying JS make it harder to debug from the browser?

chriskacerguis:
@AparnaKarve it would in a production environment, but we don't want to do that in a dev one.

AparnaKarve:
@chriskacerguis The production environment already has the JS minified (at least that's what I've seen on a QE appliance)

AllenBW:
@AparnaKarve @himdel @chriskacerguis true story, we already min and uglify built js , one of the changes that happened on a commit when updating the dependencies and after the eslint prs made it in(as this enabling this was being held back by a mix of outta date dependencies and absent di for some components)

PR #220, opened 2016-10-04
Display notifications using toasts and in the drawer

skateman:
Analogous with https://github.com/ManageIQ/manageiq/pull/11448 and because they are very similar, the plan is to join&move them to the [ui-components](https://github.com/ManageIQ/ui-components) repo.

@himdel can you verify that everything is on its proper place?
@jeff-phillips-18 the notifications aren't displayed sometimes (or with a high delay), am I missing sometihng here?

### DIFF
--- a/client/app/core/constants.js
+++ b/client/app/core/constants.js
@@ -1,9 +1,10 @@
-/* global toastr:false, _:false, $:false, sprintf: false */
+/* global toastr:false, _:false, ActionCable:false, $:false, sprintf: false */
 (function() {
   'use strict';
 
   angular.module('app.core')
     .constant('lodash', _)
+    .constant('ActionCable', ActionCable)
     .constant('toastr', toastr)
     .constant('sprintf', sprintf);
 })();
--- a/client/app/resources/collections-api.factory.js
+++ b/client/app/resources/collections-api.factory.js
@@ -10,6 +10,7 @@
       query: query,
       get: get,
       post: post,
+      delete: remove,
     };
 
     return service;
@@ -46,6 +47,20 @@
       // $log.debug("post = " + url + buildQuery(options));
 
       return $http.post(url, data, buildConfig(options))
+        .then(handleSuccess);
+
+      function handleSuccess(response) {
+        return response.data;
+      }
+    }
+
+    // delete is a reserved word in JS
+    function remove(collection, id, options) {
+      var url = API_BASE + '/api/' + collection + '/' + (id || "") + buildQuery(options);
+
+      // $log.debug("post = " + url + buildQuery(options));
+
+      return $http.delete(url, buildConfig(options))
         .then(handleSuccess);
 
       function handleSuccess(response) {
--- a/client/app/services/event-notifications.service.js
+++ b/client/app/services/event-notifications.service.js
@@ -5,7 +5,7 @@
   .factory('EventNotifications', EventNotificationsFactory);
 
   /** @ngInject */
-  function EventNotificationsFactory($timeout, lodash, CollectionsApi) {
+  function EventNotificationsFactory($timeout, lodash, CollectionsApi, Session, logger, ActionCable) {
     var state = {};
     var toastDelay = 8 * 1000;
 
@@ -53,6 +53,25 @@
 
     doReset();
 
+    var cable = ActionCable.createConsumer('/ws/notifications');
+
+    cable.subscriptions.create('NotificationChannel', {
+      disconnected: function () {
+        var vm = this;
+        Session.requestWsToken().then(null, function () {
+          logger.warning('Unable to retrieve a valid ws_token!', null, 'Notifications');
+          // Disconnect permanently if the ws_token cannot be fetched
+          vm.consumer.connection.close({ allowReconnect: false });
+        });
+      },
+      received: function (data) {
+        $timeout(function () {
+          var msg = miqFormatNotification(data.text, data.bindings);
+          service.add('event', data.level, msg, {message: msg}, data.id);
+        });
+      },
+    });
+
     return service;
 
     function doReset() {
@@ -61,7 +80,7 @@
         heading: 'Events',
         unreadCount: 0,
         notifications: [],
-      }
+      };
       state.groups = [events];
       state.unreadNotifications = false;
       state.toastNotifications = [];
@@ -99,6 +118,7 @@
         type: type,
         message: message,
         data: notificationData,
+        href: id ? '/api/notifications/' + id : undefined,
         timeStamp: (new Date()).getTime(),
       };
 
@@ -194,6 +214,7 @@
         var resources = group.notifications.map(function (notification) {
           notification.unread = false;
           service.removeToast(notification);
+
           return { href: notification.href };
         });
         if (resources.length > 0) {
--- a/client/app/services/event-notifications.service.js
+++ b/client/app/services/event-notifications.service.js
@@ -5,7 +5,7 @@
   .factory('EventNotifications', EventNotificationsFactory);
 
   /** @ngInject */
-  function EventNotificationsFactory($timeout, lodash) {
+  function EventNotificationsFactory($timeout, lodash, CollectionsApi) {
     var state = {};
     var toastDelay = 8 * 1000;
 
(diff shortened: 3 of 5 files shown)


jeff-phillips-18:
@skateman I believe the delay you are seeing is due to the notifications being added outside of an angular digest cycle. If you put the add call inside a $timeout (with no delay) you should see the immediate display of the new notification.

client/app/services/event-notifications.service.js
```diff
@@ -42,21 +42,70 @@
       }
     };
 
+    function miqFormatNotification(text, bindings) {
+      var str = __(text);
+      _.each(bindings, function (value, key) {
```
himdel: `_` -> `lodash`, and needs to be dependency-injected..

client/app/services/event-notifications.service.js
```diff
@@ -42,21 +42,70 @@
...
+
     doReset();
 
+    var cable = ActionCable.createConsumer('/ws/notifications');
+
+    cable.subscriptions.create('NotificationChannel', {
+      disconnected: function () {
+        var _this = this;
+        Session.requestWsToken().then(null, function () {
+        console.warn("Unable to retrieve a valid ws_token!");
```
himdel: broken indent

client/app/services/event-notifications.service.js
```diff
@@ -42,21 +42,70 @@
...
+
+      var options = {
+        expand: 'resources',
+        attributes: 'details'
+      };
+
+      CollectionsApi.query('notifications', options).then(function (result) {
+        result.resources.forEach(function(resource) {
+          var msg = miqFormatNotification(resource.details.text, resource.details.bindings);
+          events.notifications.splice(0, 0, {
```
himdel: isn't this `events.notifications.unshift`?

client/app/services/event-notifications.service.js
```diff
@@ -185,16 +242,22 @@
...
     }
 
     function clearAll(group) {
       if (group) {
+        var resources = [];
         angular.forEach(group.notifications, function(notification) {
+          resources.push({ href: notification.href });
           service.removeToast(notification);
         });
+        CollectionsApi.post('notifications', undefined, {}, {action: 'delete', resources: resources});
```
himdel: if resources.length?

client/app/services/event-notifications.service.js
```diff
@@ -185,16 +242,22 @@
...
+          }
           updateUnreadCount(group);
         }
       }
     }
 
     function clearAll(group) {
       if (group) {
+        var resources = [];
         angular.forEach(group.notifications, function(notification) {
```
skateman: @himdel is it possible to use a simple `map` here? I'm not familiar with this `angular.forEach` and I would like to get rid of the `resources.push` line.

client/app/resources/collections-api.factory.js
```diff
@@ -10,6 +10,7 @@
       query: query,
       get: get,
       post: post,
+      delete: delete_,
```
AllenBW: any reason why this isn't just `delete`?

client/app/services/event-notifications.service.js
```diff
@@ -5,7 +5,7 @@
   .factory('EventNotifications', EventNotificationsFactory);
 
   /** @ngInject */
-  function EventNotificationsFactory($timeout, lodash) {
+  function EventNotificationsFactory($timeout, lodash, CollectionsApi, Session, logger) {
```
AllenBW: Does `ActionCable` need to be added here?  (kinda rhetorical question, it does, but if it doesn't need to be to function, we need to declare it as a constant in a similar was as lodash and inject it into the factory)

client/app/services/event-notifications.service.js
```diff
@@ -42,21 +42,72 @@
...
-        {
-          notificationType: 'event',
-          heading: "Events",
-          unreadCount: 0,
-          notifications: [],
-        },
-      ];
+      var events = {
+        notificationType: 'event',
+        heading: "Events",
```
AllenBW: single quotes ❣️

client/app/resources/collections-api.factory.js
skateman: It's a reserved word, can't declare a function with that name :disappointed:

chriskacerguis: maybe something different than "delete_"?

skateman:
@himdel removed WIP and tests are green, good to go from my side

skateman:
@miq-bot add_label euwe/yes

chriskacerguis:
@skateman should this be included in Euwe?

client/app/core/constants.js
```diff
@@ -4,6 +4,8 @@
 
   angular.module('app.core')
     .constant('lodash', _)
+    // eslint-disable-next-line
```
himdel: This one is not like the others :)

Looking at the whole file, we should be consistent - so no `window`, and no comment, but add it to the list on the first line :)

chriskacerguis: I'm curious, what is your reasoning behind disabling the rule vs. fixing it?

client/app/services/event-notifications.service.js
```diff
@@ -42,21 +42,73 @@
...
+      return str;
+    }
+
     doReset();
 
+    var cable = ActionCable.createConsumer('/ws/notifications');
+
+    cable.subscriptions.create('NotificationChannel', {
+      disconnected: function () {
+        // eslint-disable-next-line consistent-this
```
chriskacerguis: @AllenBW thoughts here?  I see this alot, and is this something we want to add to the .eslintrc.json or get people to fix it?

tests/core.module.spec.js
```diff
@@ -7,5 +7,7 @@
   function mock($httpBackend) {
     $httpBackend.when('GET', /available_languages.json/)
       .respond({});
+    $httpBackend.when('GET', /notifications/)
```
himdel: Aah, can you make this `/api\/notifications/`? We do have a few files with `notifications` in their path...

client/app/core/constants.js
himdel: This one can't be fixed. It's a global, which is not defined in this file :) so.. we really have to disable the rule here (or do something ugly like `window["ActionCable"]`)

client/app/services/event-notifications.service.js
himdel: Oh yeah, this one we should look into.. `vm` makes sense as an alias for `this` *only* in the context of angular controllers, where it's a convention that stands for view-model.  But there's a lot of places in JS where you need to save this somewhere else (at least until we move to ES6) - so I'd be for relaxing the rule, or, if we want to be strict, extending the list of options to something like `"vm", "self"` or `"vm", "service", ...`..

AllenBW: Agree with @himdel it is 💯% an angular thing, that being said, in the context of angular I am a **very** strong proponent of `vm = this` always forever :finnadie: 

Consistency throughout will be kindness to the future...

himdel:
LGTM :+1:

AllenBW:
apologies for late feedback, there appears to be a disconnect between what my mind thinks is happening and what my fingers are actually doing.. 😕

client/app/services/event-notifications.service.js
himdel: Consistency good, though if feels weird outside controllers :) (As in, every time I see it, I think I'm in a controller.) (Also in the context of manageiq, `vm` is a bit unfortunate.. as in, at some point, we'll have a VmController to show the list of vms, and we'll be ..uh, hunter2ed :).)

chessbyte:
Euwe Backport conflict:
```sh
$ git cherry-pick -e -x -m 1  a24e991
error: could not apply a24e991... Merge pull request #220 from skateman/async-notifications
hint: after resolving the conflicts, mark the corrected paths
hint: with 'git add <paths>' or 'git rm <paths>'
hint: and commit the result with 'git commit'

$ git status
On branch euwe
Your branch is up-to-date with 'upstream/euwe'.
You are currently cherry-picking commit a24e991.
  (fix conflicts and run "git cherry-pick --continue")
  (use "git cherry-pick --abort" to cancel the cherry-pick operation)

Changes to be committed:

	modified:   bower.json
	modified:   client/app/core/constants.js
	modified:   client/app/resources/collections-api.factory.js
	modified:   tests/core.module.spec.js

Unmerged paths:
  (use "git add <file>..." to mark resolution)

	both modified:   client/app/services/event-notifications.service.js

$ git diff
```

```diff
diff --cc client/app/services/event-notifications.service.js
index 3fd09ec,7325d77..0000000
--- a/client/app/services/event-notifications.service.js
+++ b/client/app/services/event-notifications.service.js
@@@ -5,7 -5,7 +5,11 @@@
    .factory('EventNotifications', EventNotificationsFactory);
  
    /** @ngInject */
++<<<<<<< HEAD
 +  function EventNotificationsFactory($timeout) {
++=======
+   function EventNotificationsFactory($timeout, lodash, CollectionsApi, Session, logger, ActionCable) {
++>>>>>>> a24e991... Merge pull request #220 from skateman/async-notifications
      var state = {};
      var toastDelay = 8 * 1000;
  
@@@ -150,17 -198,15 +206,27 @@@
        }
      }
  
++<<<<<<< HEAD
++=======
+     function markRead(notification, group) {
+       if (notification && notification.href) {
+         CollectionsApi.post('notifications', notification.id, {}, {action: 'mark_as_seen'});
+       }
+       updateNotificationRead(false, notification, group);
+     }
+ 
++>>>>>>> a24e991... Merge pull request #220 from skateman/async-notifications
      function markUnread(notification, group) {
 -      updateNotificationRead(true, notification, group);
 +      if (notification) {
 +        notification.unread = true;
 +      }
 +      if (group) {
 +        updateUnreadCount(group);
 +      } else {
 +        state.groups.forEach(function(group) {
 +          updateUnreadCount(group);
 +        });
 +      }
      }
  
      function markAllRead(group) {
```

chessbyte:
Backported to Euwe via https://github.com/ManageIQ/manageiq-ui-self_service/pull/234

PR #221, opened 2016-10-04
Update gulp optimize to rely on ejs for build dependency injection

### LINKED ISSUE #198: Update Production Build Process to rely soley on ejs
`index.html` presently relies on wiredep for dependency injection when building for production.  This action should be assumed, as it is in dev, by ejs or in this case we would use [gulp-ejs](https://www.npmjs.com/package/gulp-ejs) but not before all other dependencies are updated (to ensure seamless functionality)

AllenBW:
closes #198 

Removes a number of extraneous build tasks
Removes undesired `<--` comments from index
Improves build flow, personally I believe this flow to be more logical, easy to follow, readable

@chriskacerguis @himdel @jeff-phillips-18 @dtaylor113 FYSA nothing noticeably changes, except a few minor things (removing that silly comment `<--` comment n such)

### DIFF
--- a/client/index.ejs
+++ b/client/index.ejs
@@ -1,66 +1,44 @@
 <!DOCTYPE html>
 <html ng-app="app" class="layout-pf layout-pf-fixed layout-pf-fixed-inner-scroll transitions">
 <head lang="en">
-  <title ng-bind="title" translate>Self Service</title>
-  <meta charset="utf-8"/>
-  <meta http-equiv="X-UA-Compatible" content="IE=edge, chrome=1"/>
-  <meta name="viewport"
-        content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no"/>
-  <link rel="shortcut icon" href="/favicon.ico"/>
-  <link rel="icon" href="/favicon.ico"/>
-  <base href="/self_service/">
-  <style>
-    [ng-cloak] {
-      display: none;
-    }
+    <title ng-bind="title" translate>Self Service</title>
+    <meta charset="utf-8"/>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge, chrome=1"/>
+    <meta name="viewport"
+          content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no"/>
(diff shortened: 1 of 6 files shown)


client/index.ejs
```diff
@@ -1,66 +1,44 @@
...
-  <!-- <%- include styles.html %> -->
+    <title ng-bind="title" translate>Self Service</title>
+    <meta charset="utf-8"/>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge, chrome=1"/>
+    <meta name="viewport"
+          content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no"/>
+    <link rel="shortcut icon" href="/favicon.ico"/>
+    <link rel="icon" href="/favicon.ico"/>
+    <base href="/self_service/">
+    <style>
```
chriskacerguis: Is this style section needed?  I thought (I could be wrong) that it was in the CSS files?

AllenBW:
@miq-bot add_label enhancement, euwe/no

client/index.ejs
```diff
@@ -1,66 +1,44 @@
...
-
-  <!-- <%- include styles.html %> -->
+    <title ng-bind="title" translate>Self Service</title>
+    <meta charset="utf-8"/>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge, chrome=1"/>
+    <meta name="viewport"
+          content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no"/>
+    <link rel="shortcut icon" href="/favicon.ico"/>
+    <link rel="icon" href="/favicon.ico"/>
+    <base href="/self_service/">
```
chriskacerguis: @AllenBW Do we want that hard coded?

himdel:
Well, :-1: for randomly changing the indent of `index.html`, but otherwise, wheee, finally no while line! :) :+1:

client/index.ejs
AllenBW: @chriskacerguis not even sure we need this... will make a note to look at 🤘

AllenBW:
`command + l` is my mistress, set at two spaces, she formats all my life!!!

hope **NOTHING** breaks (it shouldn't, this delivers identical end result)

PR #218, opened 2016-10-03
Add unread notification indicator, clean up notifications

jeff-phillips-18:
- Adds unread notification indictor badge to the nav bar
- move indicator badges to top right and change color from red to blue
- Fix notification items icons
- Update visuals for notification drawer action buttons
- Move toast notification list to be below the nav bar
- Update administration section’s icon

Indicators and Toast Positioning:
image

Notification drawer items, Administration section icon:
image

@serenamarie125 @beanh66 @chriskacerguis

PR #222, opened 2016-10-04
[EUWE] Update gettext catalog for euwe

mzazrivec:
Requested today by translators.

The PR also removes `maxlength` check from `.jshintrc` for the tests to pass.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -23,7 +23,7 @@
   "undef"         : true,     // true: Require all non-global variables to be declared (prevents global leaks)
   "unused"        : false,    // true: Require all defined variables be used
   "strict"        : false,    // true: Requires all functions run in ES5 Strict Mode
-  "maxparams"     : 10,       // {int} Max number of formal params allowed per function
+  "maxparams"     : false,    // {int} Max number of formal params allowed per function
   "maxdepth"      : 4,        // {int} Max depth of nested blocks (within functions)
   "maxstatements" : 64,       // {int} Max number statements per function
   "maxcomplexity" : 12,       // {int} Max cyclomatic complexity per function
--- a/client/gettext/po/manageiq-ui-self_service.pot
+++ b/client/gettext/po/manageiq-ui-self_service.pot
@@ -21,7 +21,7 @@ msgstr ""
 msgid "'%s' was succesfully saved."
 msgstr ""
 
-#: app/layouts/application.html:49
+#: app/layouts/application.html:75
 msgid "(Current Group)"
 msgstr ""
 
@@ -39,6 +39,10 @@ msgstr ""
 
 #: app/components/retire-service-modal/retire-service-modal-service.factory.js:62
 msgid "4 Weeks"
+msgstr ""
+
+#: app/layouts/application.html:52
+msgid "About"
 msgstr ""
 
 #: app/states/about-me/about-me.html:1
@@ -72,7 +76,8 @@ msgstr ""
 msgid "Add Profile"
 msgstr ""
 
-#: app/states/marketplace/details/details.html:29
+#: app/states/marketplace/details/details.html:36
+#: app/states/marketplace/details/details.html:41
 msgid "Add to Shopping Cart"
 msgstr ""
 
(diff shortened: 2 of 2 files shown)


AllenBW:
I vote we up ` "maxparams" : 10, // {int} Max number of formal params allowed per function` to **12** to fix this build and #211 and #218

As jshint goes away next version, no point in fighting the linters here 
@chriskacerguis @mzazrivec @himdel

AparnaKarve:
@AllenBW If`.jshintrc` was removed here : https://github.com/ManageIQ/manageiq-ui-self_service/pull/127/commits/0cb745a9a6997feaa9471ae769eede24eb978902, how come the `"maxparams" : 10` failure is still showing up?

AllenBW:
@AparnaKarve that wasn't flagged as a go for euwe, code here (and in all the other failing euwe prs outstanding) still use jscs 
 https://github.com/mzazrivec/manageiq-ui-self_service/tree/0bcc44dbc897035f5d6c4a6f2e76ab733f011c3e

himdel:
Agreed, no point in fighting it, I'd actually just set to 0.

himdel:
Uh, not 0, false :-)

AparnaKarve:
@AllenBW For some reason I assumed this commit : 0cb745a was backported to euwe as well since the backporting is usually like-for-like.
Any particular reason why we are excluding that commit for euwe and keeping it for upstream only?

AllenBW:
@AparnaKarve i am the not right person to ask why it was excluded :P @chriskacerguis ?

chriskacerguis:
Looks like it got lost in the euwe/conflicts that happened with @ZitaNemeckova commit.  Can we just re-add what we need?

chriskacerguis:
@mzazrivec could you address that jshint issue, not sure what happened.

AparnaKarve:
@chriskacerguis Did we ever plan on removing `.jshintrc` ? Or was it unintentionally removed in 0cb745a? (I'm still not clear on that)

- If it was an unintentional delete, then we'll need an upstream PR to put `.jshintrc` back in, with the `"maxparams" : 10"` rule relaxed. The upstream PR should then be backported to euwe. Once we have taken these steps,  the `"maxparams"` related failures here in this PR would be resolved.

- If it was meant to be deleted, then we'll need a euwe-specific PR to delete `.jshintrc`

Either of the above 2 actions should sync-up things between upstream and euwe-backporting.

AllenBW:
@AparnaKarve hmm something odd is going on, all jshint and jscs has been intentionally removed from sui master here: https://github.com/ManageIQ/manageiq-ui-self_service/commit/67475e886cbd7cd470932c051ed536b53d8be6b4

if we aren't backporting eslint commits, it should definitely be restored

chriskacerguis:
@AparnaKarve no, it should not have been removed.  

@AllenBW no worries, can we just add the files back real quick?

himdel:
It was not removed, that's why we're hitting the failure here. 0cb745a had a rebase artifact removing it, but that change never got to #205 (correctly) so `.jshintrc` *is* in euwe and apparently supposed to be (since https://github.com/ManageIQ/manageiq-ui-self_service/pull/184 is `euwe/no`).

And yes, to make the tests pass here, we'll need to fix jshintrc in darga. That limit of 10 params is simply untenable and needs to be removed - it was never designed to work with angular's DI anyway.

@mzazrivec Could you please change that offending line (`maxparams`) in `.jshintrc` to `false`?

(BTW, @chriskacerguis this is a perfectly good example of what I'm afraid of in #204 - a perfectly good change blocked by an unrelated style check failure (granted, complicated by the fact that the old style checks are run as part of the specs but still))

himdel:
@chessbyte it's green now :)

chriskacerguis:
@himdel I guess I'm not following your comment.  (1) because that is talking about Code Climate, and this was a Travis failure (going to assume you mean build failures in general), but (2) I far as I see the system worked.  We got a build failure, we investigated, and we resolved in a rather quick process.  AND once we identified the issue, I was more than happy to do a merge with it red, but I wanted to make sure that others in different timezones had the opportunity to comment.  Seems to me that the system worked fine.  If you don't like the way that the system is working (which is totally cool/good to question it...maybe can you log an issue with your proposal (I don't feel that #204 is the place for that, but feel free to discuss in there) and we can discuss with the team at large)?

himdel:
@chriskacerguis well, OK, if that's what you take from that, I'm happy with how things are :) I was just really afraid of having a policy of "jshint fails, can't merge" and wanted to make that clear :).

PR #233, opened 2016-10-05
Restores eslint rules

### LINKED ISSUE #228: Restore all relevant eslint policies
So we rolled back from node 6 to node 0.10 and a number of eslint policies had to be disabled (because they didn't existing in the older version of eslint).  This is an issue designed to restore those rules and include any fixing required for `gulp eslint` to pass.

@miq-bot add_label 'technical debt'

AllenBW:
Good job team, no rules violated in between the eslint refactoring and now! 🙇‍♀️ 🙏 👏  

These rules were commented out to satisfy an older version of eslint.
Now that dependencies have been updated they can be restored.

@chriskacerguis 

closes #228

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,9 +12,9 @@
   "plugins": [
   ],
   "globals": {
-    "flowchart": true,  // Blueprints support lib
-    "__": true,  // Zanata parsing cue
-    "N_": true   // Zanata parsing cue
+    "flowchart": true, // Blueprints support lib
+    "__": true, // Zanata parsing cue
+    "N_": true // Zanata parsing cue
   },
   "rules": {
     "accessor-pairs": "error",
@@ -42,7 +42,7 @@
         "properties": "always"
       }
     ],
-//    "class-methods-use-this": "error",
+    "class-methods-use-this": "error",
     "comma-dangle": [
       "error",
       "always-multiline"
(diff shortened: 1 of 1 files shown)


chriskacerguis:
@AllenBW doesn't this need to go in the Euwe branch, not master?

AllenBW:
@chriskacerguis I thought euwe was a ⛔️ for eslint and friends? (defer to you)

chriskacerguis:
It is, but you are doing a PR to the "master" branch.  You need to do it against the Euwe branch 😄

AllenBW:
mmm ewue doesn't suffer from what this pr cures, https://github.com/ManageIQ/manageiq-ui-self_service/blob/euwe/.eslintrc.json

AllenBW:
@chriskacerguis actually this reveals a larger concern, so we include eslint config, but no dependencies in `package.json` furthermore, this pr reverses what I would expect to reflect the status of eslint config for the lover version of node

chriskacerguis:
@AllenBW ok, so the question I have right now is...we are using ESLint now NOT JShint, correct?

AllenBW:
@chriskacerguis Correct, On master there is no jshint. On euwe it was my understanding we would continue to use jshint, (because its still in place) but what do i know :P

chriskacerguis:
So, I'm thinking we need to delete the jshint file on MASTER, and update the one in Ewue.

AllenBW:
transposed? jshint isn't on master eslint is and neeeeds to be updated

chriskacerguis:
Ok, got my jshint / eslint confused.

PR #235, opened 2016-10-06
Sample skin + docs

### LINKED ISSUE #115: How to skin the application is not clear
Previously, if someone dropped files into the skin directory, those files would be used for images and text.  Now, I believe the place is assets (even though I search the code and I still see skin)?  I'm a little confused what is the right way to skin the application.

So this is a request for some documentation in the README on how to skin the application, and perhaps a set of tests that validate that if a file is placed in a particular place, then it is actually used (and thus the test serves as an example as well)

himdel:
This adds a `skin-sample` folder, containing a simple sample skin, and a README describing how the skinning works, where to look, and what to override.

If everybody agrees with the wording of the README, I'll do a similar PR with just the readme to ManageIQ/guides so that it's part of the devel docs as well.

Applying the skin: `skin-sample/link.sh` (optionally with the path to the ssui folder)
Unapplying: `skin-sample/unlink.sh` (or `rm client/skin`)

Closes #115


himdel:
@epwinchell if you ever have some extra time, maybe this sample skin deserves some love so that it actually looks good. But it is just a *sample* skin, so..

himdel:
The CC issues can't be fixed IMO...

BTW @AllenBW should the scss linter check *css* files too? At least 2 of the errors don't even make sense for css..

AllenBW:
@himdel I would say yessssssss we should be checking css files  ALL TEH FILES

seriously though, if these things cant be changed, welp tough luck codeclimate we're not changing them!  

<3

himdel:
@chriskacerguis Nope, sorry, I [cannot](https://github.com/ManageIQ/manageiq-ui-self_service/pull/235#issuecomment-251960339) :). Not without breaking it..

chriskacerguis:
@himdel not arguing, but for the sake of completeness could you maybe shed some light on to why you feel that way?

himdel:
Yup, already on it.. :)

> Selector splash__message should be written in lowercase with hyphens
Not without breaking everybody's skin.. - it's overriding styling for this class (actaully, we're using BEM AFAIK so .. different convention even - will investigate)

> Begin pseudo elements with double colons: ::
Not in CSS

> Color literals like #404 should only be used in variable declarations; they should be referred to via variable everywhere else.
Not in CSS.

> Similar code found in 1 other location (mass = 111)
Well, yes, it's an override.

chriskacerguis:
Thanks @himdel makes sense to me!

chriskacerguis:
Merging due to @himdel code additions being "valid".

PR #238, opened 2016-10-08
Add a new view service dialog list to SUI

dgutride:
This adds a new menu option for dialogs under the designer section (along with blueprints).  The page only consists of a list with filter by name/last updated and sorting by the same two columns.

New unit tests for the list view state, loading and sorting were also added. 

The other major change is from code climate errors when run locally.  The state services in the simpler states consist of boilerplate code that is repeated regularly (for setting up filtering and sorting functions).  This was pulled out into a list configuration service that shares the code now.  Code climate did not report any errors on the services folder after this.
@chriskacerguis @serenamarie125 @dclarizio 

screen shot 2016-10-08 at 7 11 23 am
designer-dialogs

### DIFF
--- a/client/app/services/list-configuration.service.js
+++ b/client/app/services/list-configuration.service.js
@@ -1,0 +1,40 @@
+/* eslint camelcase: "off" */
+(function() {
+  'use strict';
+
+  angular.module('app.services')
+    .factory('ListConfiguration', ListConfigurationFactory);
+
+  /** @ngInject */
+  function ListConfigurationFactory() {
+    var configuration = {};
+
+    configuration.setupListFunctions = function(list, currentField) {
+      list.sort = {
+        isAscending: true,
+        currentField: currentField,
+      };
+
+      list.filters = [];
+
+      list.setSort = function (currentField, isAscending) {
+        list.sort.isAscending = isAscending;
+        list.sort.currentField = currentField;
+      };
+
+      list.getSort = function () {
+        return list.sort;
+      };
+
+      list.setFilters = function (filterArray) {
+        list.filters = filterArray;
+      };
+
+      list.getFilters = function () {
+        return list.filters;
+      };
+    };
+
+    return configuration;
+  }
+})();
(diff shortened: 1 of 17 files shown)


chriskacerguis:
@dgutride can you please take a look at those CodeClimate issues?

dgutride:
@chriskacerguis  - I reviewed the new code climate issues and while it says the code is similar, this is navigation code that will have a similar structure but should not be refactored out into a helper function.  Is there a way to add an ignore for this error to indicate I reviewed it and feel that certain code needs to be descriptive and verbose and not combined?  Too much optimization in this particular case will make the code unreadable and won't help future developers in my opinion.

client/app/services/list-configuration.service.js
```diff
@@ -0,0 +1,40 @@
+/* eslint camelcase: "off" */
+(function() {
+  'use strict';
+
+  angular.module('app.services')
+    .factory('ListConfiguration', ListConfigurationFactory);
```
dtaylor113: Hi, couldn't these helper functions been added to the already existing ListView.js?

chriskacerguis:
@dgutride I don't know of any way to "ignore" issues via comments.  You might look at the CC docs to see, that said, I think that discussion like this is sufficient to move forward with a merge.

tests/dialogs/dialogs-list.directive.spec.js
```diff
@@ -0,0 +1,86 @@
...
+  };
+
+  beforeEach(function () {
+    Session.create({
+      auth_token: 'b10ee568ac7b5d4efbc09a6b62cb99b8',
+    });
+    $httpBackend.whenGET('').respond(200);
+
+    $scope.dialogs = [
+      {
```
dtaylor113: Note:  We started adding this mock data to the 'tests/mock' directory as seperate json files.

dtaylor113:
Hi, this PR was merged while I was reviewing it.  I appreciate the fast turn around, but it was just put up on Friday afternoon (I believe) and I'm reviewing it first thing Monday morning and no other developer has reviewed it.  I don't think PRs should be merged without at least one developer giving a :+1: .  -thanks!

chriskacerguis:
@chriskacerguis <- apparently, no longer a developer :(

dtaylor113:
Sorry @chriskacerguis , didn't mean to imply that.  I see you as more of an overseer of these PRs.  I meant a developer working in the SUI space.  No worries :-)

PR #252, opened 2016-10-17
Adds yarn, updates travis to install using yarn

### LINKED ISSUE #245: Migrate to yarn for packages?
Seems there has been some talk about suing Yarn, so I'm going to open an issue to discuss.

Info about Yarn: https://code.facebook.com/posts/1840075619545360
Site: https://yarnpkg.com/en/docs/install

I'm not advocating we switch (yet), but it brings up a lot of good points / issues so I thought it would be good to at least discuss (as I'm sure others will ask about it).

AllenBW:
Yarn does not yet jive with bower, when it does this build step will be removed: https://github.com/yarnpkg/yarn/pull/896

Added gulp-imagemin optional dependencies 

closes #245 

@chriskacerguis @himdel


AparnaKarve:
@chriskacerguis Looks like this a done deal.
So going forward would we be using `yarn` instead of `npm`?

chriskacerguis:
Hi @AparnaKarve yes, that is correct (the readme is updated).

AllenBW:
@AparnaKarve you're correct, `yarn` or `yarn install` will now be the way to setup dependencies

We aren't exactly using yarn instead npm.   npm is still very much involved in the app, you could think of yarn as an package.json manager.

himdel:
@AllenBW I'm seeing a `error npm-shrinkwrap.json found. This will not be updated or respected. See [TODO] for more information.` - we should probably reflect what's in shrinkwrap to yarn.lock and remove that shrinkwrap...

And another thing, should we remove `scripts.install` & `scripts.postinstall` from `package.json` now?  Feels like we shouldn't be using npm-specific scripts if we're not using npm..

AllenBW:
@himdel Thanks for the feedback!! So first point, do we need the shrinkwrap anymore? (After removing the file and running yarn install again)`gulp templatecache` operates as expected, updating the dependencies likely resolved this isssssssssssuuuuueeeeeeee. 🌮

☝️ so you mind if i axe the file in question? 

Second point, SO a cool thing yarn does is `yarn run foo` where foo can be the name of any script we have in our `package.json`.... the bummer is right now it doesn't quite work (see https://github.com/yarnpkg/yarn/issues/1156) , but _IDEALLY_ I would like to keep the scripts there, and references them in the future rather than rewrite them all the places they are required

himdel:
Right.. looking at the issue that caused us to have npm-shrinkwrap in the first place (https://github.com/tracker1/gulp-header/issues/37), it's resolved and probably only affected gulp-header 1.8.3, so 1.8.8 should be fine.

So feel free to completely and utterly destroy npm-shrinkwrap.json ;).

And if the scripts are planned to be supported, then sure, let's keep them :)

PR #261, opened 2016-10-18
Enhancement/251 remove bower components

AllenBW:
Updates zatana file names
Updates dev url
Updates documentation 

### all to service ❗️ 🌮

@chriskacerguis @Fryguy @himdel


himdel:
Uh-oh.. I'm almost sure there'll be bugs related to changing the URL.. IIRC some of our css references paths with `../` which may or may not have to become `../../`.

Has anybody tested this in production?

himdel:
Also.. this needs links to PRs in manageiq-appliance, manageiq-appliance-build and manageiq to fix all the code that depends on the wrong `public/` location or the wrong URL.

AllenBW:
@himdel looping in @Fryguy as that might have been included in his efforts as they relate to this, if not then yes! if so then no 🤘

Issue #259, opened 2016-10-18
Update gulp serve to start MIQ as well


chriskacerguis:
We need to update the gulp serve task to start MIQ start use the "rails s"

chriskacerguis:
Thinking about what @himdel said.  I'm thinking we should revisit this.  Le's adjust gulp serve to NOT start MIQ, and instead let's add a new task that will start both.

himdel:
Agreed, that makes sense :) Either that, or a way to disable the functionality, via some sort of ENV variable.. but I like a separate task even better :).

And.. the task should probably fail when spawning rails fails..

AparnaKarve:
> but I like a separate task even better

I second that. A separate gulp task would be better. No confusion whatsoever.

chalettu:
I resubmitted this as a new PR https://github.com/ManageIQ/manageiq-ui-service/pull/268.   Feel free to take a look.  It now is a separate gulp task and it accepts a ENV variable that will allow you to override the port that you start Manage IQ on.  I ran our ES Lint rules against the new file and corrected the issues it found. Please feel free to take a look at it.  Thank you all for your input.

chriskacerguis:
#268 has been merged.

PR #265, opened 2016-10-18
Gulp ruby integration

chalettu:
This commit is for issue
https://github.com/ManageIQ/manageiq-ui-service/issues/259 which asked
to be able to automatically launch a miq backend when gulp serve is
run.


gulp/tasks/manageiq-server.js
```diff
@@ -0,0 +1,13 @@
...
+module.exports = function (gulp, options) {
+    var manageiq_dir = require('../config')[options.key || 'manageiq_dir'];
+
+    return task;
+
+    function task() {
+
+          var cmd = spawn('rails', ['s'], {stdio: 'inherit',cwd:manageiq_dir});
+    }   
+};
```
martinpovolny: missing endline, please, fix you editor setting.

himdel:
Please don't. SSUI should never try to start the manageiq app. It should depend on the app already running.

Most devels already work with manageiq started, sometimes on diffferent ports, this is not something SSUI should ever do.

Also, so many style checks must have failed on that file.. was there any need to rush this so much?

martinpovolny:
Honestly I have to say that I don't like the speed of merge here.

martinpovolny:
By the way what about hose Merge commits here? I am quite sure we have previously refused to merge PR with such issues.

chriskacerguis:
@chalettu @AllenBW Martin brings up a good point here about style checks.  It looks like it passed Travis / CodeClimate, but there appear to be some style issues.  Can you please take a look at our eslint rules, travis job, etc to ensure that we have everything setup as needed?

@chalettu Can you please address the style comments in a new PR, and please run ESlint to make sure that all are good.

@himdel RE: the MIQ app already running...since gulp serve is just run in development, I'm not sure that it really matters who starts what.  That said, I can think of 1 use case where that could be an issue.  What are your concerns with gulp serve starting MIQ?  The SUI team met here in Austin, TX and we liked the idea of having MIQ started by gulp.

@himdel RE: Style checks.  Every test passed, so if something was missed we should look at that (see comment to @chalettu and @AllenBW)

@martinpovolny RE: merge commits.  Yep, my bad there.  I spoke to @Fryguy about that, and we decided to just leave it (as @Fryguy said "it happens").  I have had @chalettu update his settings to ensure that doesn't happen again.

@martinpovolny RE: speed of merge.  The team is here in Austin (with @Fryguy and @chessbyte), so we were all discussing things and so we were able to move a bit faster (but, see comment above, I did make a mistake by merging the "merges" in the PR.

AllenBW:
@himdel @chriskacerguis code climate excludes checking the following folders
- tests/**/*
- node_modules/**
- gulp/**
- bower_components/**
- reports/**

furthermore  we only run eslint on app files, not build tasks

thinking we might want this changed ?

chriskacerguis:
@AllenBW Yeah, since we are going to move to Node, then we should check the Node server side, and the gulp tasks.

martinpovolny:
> Most devels already work with manageiq started, sometimes on diffferent ports,

I +1 that. I personally run different ManageIQ's on different ports from different directories so that I can comfortably switch versions and also switch between doing review/testing and working on my stuff.

Now I will have a thing that will try to run one of my copies of ManageIQ by default. I really don't like that.

I quite like to run SUI as the docs suggest:

```
PROXY_HOST=127.0.0.1:3000 gulp serve
```

that allows me to select the desired MIQ server by providing the port number.

Now I have the SUI reach out of it's directory and try to run MIQ in ../manageiq/ even if I provide port for a different MIQ server. Even one that is on a different host.

Maybe this fit's someones simple usecase, but for me this is counterintuitive. 

Best way to resolve all the issues with this PR would be reverting it. 

Minimal mitigation should include fixing the style problems and either: 
 * running the MIQ server only if the PROXY_HOST was not given or
 * provide a 2nd 'serve' task maybe 'run" that does not start the MIQ server.

PR #268, opened 2016-10-19
Created task that allows you to start ManageIQ server from a gulp task

chalettu:
This new task allows you to specify an env variable to start MIQ
services.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,8 +48,11 @@ http://localhost:3000 order and serve up the REST API.
 - From the `manageiq-ui-service` directory, start the development version of
   the service UI, which will initiate the UI listening on _http://localhost:3001_, and talking to the REST API at
   _http://[::1]:3000_.  This command will also open a browser page to  _http://localhost:3001/ui/service/login_.
-  - `gulp serve`
-
+  - `gulp serve` 
+- If you have a local copy of Manage IQ Server installed and would like to start it up at the same time you bring up the service ui web server, run   
+	- ``` gulp start-manageiq-server ```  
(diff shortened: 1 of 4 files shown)


chriskacerguis:
@himdel @martinpovolny does this work for you two?

chriskacerguis:
@chalettu Can you add some usage info to the readme?

chalettu:
@chriskacerguis I updated the readme to reflect how to configure and use the new gulp task.  Please let me know if you would like any revisions to the wording.

martinpovolny:
Cool, thanks, this surely works for me. I see a bit of disconnect between the variables `MANAGEIQPORT` and `PROXY_HOST`, but it works.

himdel:
Thanks, much better! :)

I wonder if the new task should fail if it fails to start manageiq, .. but, up to you :).

chriskacerguis:
@chalettu I like @himdel suggestions of failing if it can't start MIQ, can you make that happen?

chriskacerguis:
@chalettu ping ^

chalettu:
I looked into it and since its an ongoing process I didnt find an easy way to bubble back up a failure to the gulp task.  If anyone has a solid example of that type of pattern available I would be more than happy to modify it to support this feature.

himdel:
@chalettu Yeah, I was wondering about it too, and maybe the only way to do that would be to start, timeout for a bit, and then check if it's still running .. which doesn't sound that clean, so idk.. It's the halting problem all over again :)

PR #276, opened 2016-10-24
[EUWE] Change My Requests subsections from tabs to secondary nav

jeff-phillips-18:
This PR removes the tabs from the My Requests section and adds secondary nav items "Requests" and "Order History"

Jira Story: https://patternfly.atlassian.net/browse/CFUX-142

(cherry-picked and resolved conflicts for euwe branch)

Backport of #270 

@chriskacerguis @chessbyte

PR #277, opened 2016-10-24
Minor fixes for I18N and status icon display in services list.

jeff-phillips-18:
This fixes an issue brought up in #270 as well as a small icon
placement issue I found in the services list while testing the same.

@chriskacerguis @himdel 

@miq-bot add_label euwe/no

### DIFF
--- a/client/app/components/blueprints/blueprints-list.directive.js
+++ b/client/app/components/blueprints/blueprints-list.directive.js
@@ -130,25 +130,25 @@
       return [
         ListView.createFilterField(
           'name',
-          'Name',
-          'Filter by Name',
+          __('Name'),
+            __('Filter by Name'),
           'text'),
         ListView.createFilterField(
           'visibility',
-          'Visibility',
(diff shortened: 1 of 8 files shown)


himdel:
@jeff-phillips-18 not like this.

You really need to `N_()` the string **literal**, not any variable .. literal. You can do `__(foo)` when that foo is either a literal already or is a variable which has come from `N_("literal")`.

So you really need to fix *all* the *callers* of `createFilterField` and `createSortField` and put the `N_` there. Then (but only then) the body of those functions can indeed use `__(variable)`.

jeff-phillips-18:
@himdel ah, I see. Thanks, I will fix it.

himdel:
Thanks! this looks right :)

himdel:
Just adding a note here for clarity.. 

This is correctly `euwe/no`, but *technically* it's `euwe/backported` because these changes were part of https://github.com/ManageIQ/manageiq-ui-service/pull/276 .

PR #284, opened 2016-10-24
Add sass-lint guidelines, swaps warnings for errors

### LINKED ISSUE #282: Builds should fail on Linting warnings
In order to produce cleaned or standardized code I think that on linting errors (Sass lint, Eslint) and warnings Travis or CodeClimate (which ever does the check) should fail.

AllenBW:
Warnings don't fail our build, are rarely acknowledged, but errors will raise eyebrows

closes #282 

@chriskacerguis @himdel 

sorry everybody :-(  IT HAD TO BE DONE 😑

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -1,0 +1,91 @@
+options:
+  formatter: stylish
+files:
+  include: '**/*.s+(a|c)ss'
+rules:
+  # Extends
+  extends-before-mixins: 2
+  extends-before-declarations: 2
+  placeholder-in-extend: 2
+
+  # Mixins
+  mixins-before-declarations: 2
+
+  # Line Spacing
+  one-declaration-per-line: 2
+  empty-line-between-blocks: 2
+  single-line-per-selector: 2
+
+  # Disallows
+  no-attribute-selectors: 0
+  no-color-hex: 2
+  no-color-keywords: 2
+  no-color-literals: 2
+  no-combinators: 0
+  no-css-comments: 2
+  no-debug: 2
+  no-disallowed-properties: 0
+  no-duplicate-properties: 2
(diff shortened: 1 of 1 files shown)


chriskacerguis:
Build failing due to changes.  @AllenBW do you want to address, or just let others?

himdel:
@chriskacerguis , @AllenBW should we even still be running the linters as part of `gulp test`? Or rather, should we be running `gulp testonly` on travis instead? (So we'd use travis only for unit tests and codeclimate for linting.) WDYT?

AllenBW:
@himdel I like this rather brilliant idea, only checking changed code for errors (saves the sanity of many), though flip side, ideally we don't want any code that would error in the repo, so having that almighty **check all teh things** kinda works out right now 🐑 🛡   

@chriskacerguis I can address, it'll take my mind off this webpack nonsense :shipit: or we can push linting solely to cc and distribute this burden amongst the many

chriskacerguis:
@himdel GREAT idea.  In looking at things, using something like Hound CI would help here too.  @AllenBW make that happen.  

Great collaboration!

chriskacerguis:
Closing this, going to use @himdel 's suggestion (and Hound CI)

himdel:
@chriskacerguis .. we will still need the linter config even if we use CC or something for that, won't we? Meaning we probably do want the PR merged..?

chriskacerguis:
@AllenBW is that the right file name?  .sass-lint.yml?  Should be it .scss-lint.yml

AllenBW:
@chriskacerguis sass-lint is the correct name 🤘

chriskacerguis:
Build fails due to sass linting errors (which this PR didn't cause).

PR #291, opened 2016-10-25
Removed Task for eslint and sasslint from Test

chalettu:
As requested, I removed the listing tasks from the test Gulp task.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -62,7 +62,7 @@ gulp.task('build-specs', ['templatecache'], task('buildSpecs'));
 /**
  * Testing tasks
  */
-gulp.task('test', ['vet', 'templatecache'], task('test', {singleRun: true}));
+gulp.task('test', ['templatecache'], task('test', {singleRun: true}));
 gulp.task('testonly', ['templatecache'], task('test', {singleRun: true}));
 gulp.task('autotest', task('test', {singleRun: false}));
 


chriskacerguis:
@jeff-phillips-18 @dtaylor113 @martinpovolny @himdel I asked Chris to remove the linting from the build system, and instead let Hound CI handle that.  thoughts?

dtaylor113:
What does that mean for the build process?  Build Will pass, PR will show
errors?

AparnaKarve:
So essentially with this PR `gulp test` and `gulp testonly` now do the same thing, which is to validate the PhantomJS specs. Would that be correct?

chriskacerguis:
So the build process will work as follows:

Travis - Runs unit tests and integration (phantomJS) tests
CodeClimate - Code Security, test coverage, and various dup and linting checks
Hound - Will provide comments in your PR about issues (similar to how the MIQ bot works)

So, Travis will only fail if there is a unit / integration test fail.

himdel:
<3 Loving this :)

jeff-phillips-18:
Will there be a way to run the lint tests locally?

himdel:
@jeff-phillips-18 `gulp eslint` ;) .. also `gulp sasslint` for sass... (Or `gulp vet` for both)

jeff-phillips-18:
👍

jeff-phillips-18:
Looks like 'Hexadecimal colors should not be used' was just added. Any idea why? What is the rub on using hex?

chriskacerguis:
@jeff-phillips-18 not following, I don't see anything like that is this PR?

jeff-phillips-18:
@chriskacerguis Correct. It was added in #284 . Sorry. I can add this as an issue when I put up the PR to attempt to fix existing warnings (now errors).

chriskacerguis:
@jeff-phillips-18 ahhh...cools.  Sorry, I need more ☕️ 

Not sure, if it wasn't there, and now it is (I think Allen just added the file), feel free to do a PR where you change the .sass-lint.yml and we can talk / discuss 😄

AparnaKarve:
We have 2 identical tasks now:
```
  gulp.task('test', ['templatecache'], task('test', {singleRun: true}));
  gulp.task('testonly', ['templatecache'], task('test', {singleRun: true}));
```
Any reason to keep them both? Like I said above, we could remove `gulp testonly`

AllenBW:
@chriskacerguis @jeff-phillips-18 totally rolled with recommended values from sass-lint, we use hex everywhere so down for continuing to use hex everywhere

Side note colors in the sass should be defined in one place and referenced by name everywhere else .
`_colors.sass` for hex values `_app_colors.sass` adds functional names to the colors, these functional names should be the only reference we make, if you need a color, add a color, but at **all costs** avoid plopping in yet another color to hunt down.

chriskacerguis:
@chalettu see https://github.com/ManageIQ/manageiq-ui-service/pull/291#issuecomment-256395415

PR #293, opened 2016-10-26
Fix sass lint errors

jeff-phillips-18:
This fixes the existing sass lint errors. While most of the errors were fixed in the sass files, there were some adjustments to the lint rules as well.

lint rule adjustments:

1) no-color-hex - I see no reason to enforce this rule. I don't see any benefit
2) nesting-depth: 2 - I feel 2 is not enough, this makes it very hard to override inherited styling or to specify styles on pseudo elements. Allowing 3 would remove existing errors.
3) no-important - while I understand why we might want to disallow this, i think it is unrealistic. Without !important it is very hard to override styles especially given only 2 (or 3) nesting depth
4) class-name-format - added exceptions for overrides of styles on components from dependencies (angular-drag-and-drop-lists)

I moved a lot of color definitions from various sass files into the _colors.sass file. However, it would be good (as @AllenBW mentioned in #291) to define application colors for each and use only the application defined colors in the various sass files. This PR does not address that issue.

@chriskacerguis @AllenBW @dtaylor113


AllenBW:
👋 @jeff-phillips-18 thanks for the mention!!  also for moving color definitions <3 you're a 👑 among 🍠 for undertaking this work...
 1. agreed absolutely 
 2. agreed absolutely
 3. disagree absolutely, there are presently 12 of these to resolve, if need be I volunteer myself as tribute 🕊  `!importat` is horrible and disgusting and the worst thing ever and should never be allowed to exist
screen shot 2016-10-26 at 1 15 50 pm
 4. agree absolutely

jeff-phillips-18:
@AllenBW That's a bit harsh on the !important 😉   I agree it should be used sparingly and as you state there are only 12 instances in the code base now. Sometimes it is necessary to use !important especially given the nesting level limits.

I will take another shot at removing them and report back what I find.

AllenBW:
Truth, very harsh, you'll have to forgive my pre-lunch thoughts on important 😆  (but still its super bad).

.sass-lint.yml
```diff
@@ -29,7 +29,7 @@ rules:
   no-empty-rulesets: 2
   no-extends: 0
   no-ids: 2
-  no-important: 2
+  no-important: 0
```
chriskacerguis: The no-important seems to be a point of disagreement.  I'm inclined to agree with @AllenBW on this.  Question for @jeff-phillips-18 why would we be overriding things, if your team is doing the UX (it would seem that in that use case, there would be no need to override)?  Maybe I'm missing something.

AllenBW: :one: 🏩 ⚓️

chriskacerguis: @AllenBW please translate 1️⃣ 🏩 ⚓️?

AllenBW: @jeff-phillips-18  is my one love, my anchor ❗️ excising `!important` is dreadful ☠️ work.

jeff-phillips-18:
@AllenBW I was able to remove current uses of !important and added the rule back.

AllenBW:
# @jeff-phillips-18  YOU ARE THE MAN 🕴 🕴 🕴

serenamarie125:
Thanks @AllenBW & @chriskacerguis for accepting the adjustments to items 1, 2 & 4,  @jeff-phillips-18 thanks for bringing those to their attention and fixing those issues associated with !important.  If similar issues come up in the future, let's discuss again.

PR #315, opened 2016-11-02
Pull and compile most recent translations

jjlangholtz:
The language selection for the login page and the user settings
'Switch Language' dropdown were listing language options using the
cryptic language codes instead of human friendly locale names.

This change pulls down the latest translations from Zanata and
recompiles the json translation files along with
`available_languages.json` which is used to populate the language
switching choices.

https://bugzilla.redhat.com/show_bug.cgi?id=1389224


chriskacerguis:
@jjlangholtz can you add to the docs how to do this?

jjlangholtz:
@chriskacerguis I mostly followed the guidelines that @mzazrivec started [on the guides repo](https://github.com/ManageIQ/guides/blob/master/i18n.md#manageiq-self-service-ui). There is already a link there from the `Code Tour` doc.

There were a few minor adjustments that I had to make. I didn't do the extract/push to Zanata because I'm fairly sure I need to be credentialed for that. Also, for some reason, when pulling the files from Zanata they were named incorrectly (self_service instead of service) so I had to rename them.

Should I just make a PR for those updates to the guides repo?

chriskacerguis:
if you could do a separate PR to update the docs, that'd be great.

simaishi:
Backported to Euwe via #324

PR #319, opened 2016-11-07
Reconfigure console service to use CollectionsApi service

AllenBW:
Requires additional testing, but resolves the 404 issue when requesting console open from sui.

Cors changes the `GET` to `OPTIONS`, causes the request to fail, using `CollectionApi` service avoids this headache. 

Related bz: https://bugzilla.redhat.com/show_bug.cgi?id=1392100

@bascar @martinpovolny hey you wonderful folks, I could use some help testing this if ya can spare the cycles

@chriskacerguis if ya could hold off on merging till I get some verification on this fix... 🌮 💘 👍

### DIFF
--- a/client/app/services/consoles.service.js
+++ b/client/app/services/consoles.service.js
@@ -5,7 +5,7 @@
     .factory('Consoles', ConsolesFactory);
 
   /** @ngInject */
-  function ConsolesFactory($window, CollectionsApi, $http, $timeout, logger, $location, EventNotifications) {
+  function ConsolesFactory($window, CollectionsApi, $timeout, logger, $location, EventNotifications) {
     var service = {
       open: openConsole,
     };
@@ -15,10 +15,10 @@
     function openConsole(vmId) {
       CollectionsApi.post('vms', vmId, {}, {
         action: 'request_console',
-        resource: { protocol: "html5" },
+        resource: {protocol: "html5"},
       })
-      .then(consoleResponse)
-      .catch(consoleError);
+        .then(consoleResponse)
+        .catch(consoleError);
     }
 
     function consoleResponse(response) {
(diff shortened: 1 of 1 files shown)


AllenBW:
@miq-bot add_label euwe/yes

martinpovolny:
Following the discussion on Gitter:

The root cause of the bug is that the $http service connects to a **wrong port** as seen here:
wrong port
 
This PR fixes the issue because `CollectionApi` uses different mechanism to determine the port even thought it uses the $http service under the hood.

 1. I'd conclude that we **can merge this as a fix**.
 1. we should make sure that we don't have more places with the same issue
 1. and/or we should fix the way the port is determined in case of $http

@AllenBW , @himdel : thoughts?

himdel:
Agreed, :+1: 

> and/or we should fix the way the port is determined in case of $http

I wonder if we can get rid of that double redirect - it's happening only because of browsersync in devel mode.

AllenBW:
@martinpovolny agreed, unexpected behavior is never a good thing, unfamiliar devs will want to use $http and unknowingly introduce bugs, stopping that in the sui early on is a good call

I checked, nowhere else is there an instance that might offend the in this way, so i would say thats all folks! 🤘  (for now 👹 )

chessbyte:
Euwe Backport details:
```sh
$ git log -1
commit 43eaa054b7641f43a9bed49f4da8faec8b1f491a
Date:   Tue Nov 8 08:10:45 2016 -0600

    Merge pull request #319 from AllenBW/bug/1392100-failed-console-operations
    
    Reconfigure console service to use CollectionsApi service
    (cherry picked from commit 84b91dc0add8c0551030acd1d3e7436ed7e45bed)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1392100
```

PR #320, opened 2016-11-07
Allow approver to approve/deny service requests

jjlangholtz:
This change adds a couple of buttons to the request details state to
allow for approving or denying the selected request. The buttons should
only be visible if the current user has the correct permissions and when
the `approval_state` for the request is `pending`.

The yarn.lock file is also updated here because for some reason the
dependencies are requiring two versions of node-sass.

Also cleans up a few linting issues.

Currently working on adding tests.

### DIFF
--- a/client/app/states/requests/requests/details/details.html
+++ b/client/app/states/requests/requests/details/details.html
@@ -24,6 +24,18 @@
                 <h2>{{ ::vm.request.description }}</h2>
                 <h4>{{ ::vm.request.options.long_description || vm.request.description }}</h4>
               </div>
+
+              <div ng-if="vm.pendingApproval()">
+                <input type="text" ng-model="vm.approvalReason" placeholder="{{'Reason' | translate}}">
+                <div>
+                  <button class="btn btn-primary" type="button"
+                    ng-disabled="!vm.approvalReason"
+                    ng-click="vm.updateApprovalState('approve')" translate>Approve</button>
+                  <button class="btn btn-primary" type="button"
+                    ng-disabled="!vm.approvalReason"
+                    ng-click="vm.updateApprovalState('deny')" translate>Deny</button>
+                </div>
+              </div>
             </div>
           </div>
         </div>
--- a/client/app/components/blueprints/_blueprints.sass
+++ b/client/app/components/blueprints/_blueprints.sass
@@ -1,14 +1,14 @@
 .blueprints-list
 
   .list-view-pf-checkbox
-    margin-top: 0px
-    margin-bottom: 0px
+    margin-bottom: 0
+    margin-top: 0
 
   .list-view-pf
     padding-bottom: 68px // Adding so dropdown menus on the last row don't get cut off
 
     .list-view-pf-actions
-       margin: 0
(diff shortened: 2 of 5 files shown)


client/app/states/requests/requests/details/details.html
```diff
@@ -24,6 +24,18 @@
                 <h2>{{ ::vm.request.description }}</h2>
                 <h4>{{ ::vm.request.options.long_description || vm.request.description }}</h4>
               </div>
+
+              <div ng-show="vm.pendingApproval()">
```
AllenBW: `ng-if` preferred to `ng-show`  

reason, show essentially just plays with `display` if removes element from dom <3

chriskacerguis:
Can you add a screenshot so we can see what it looks like?

jjlangholtz:
Sure thing! Here is what it looks like for now:

screen shot 2016-11-07 at 9 40 27 am

AllenBW:
@jjlangholtz remember I <3 you

not a huge fan of shoving it in the title heading area, looks kinda busy, any way to stuff another `<hr` above the approval ?  essentially break it out into another section (like create vm or that details block)

also, maybe make it a text area or span the whole length of the screen, make it LOUD AND SCREAM FOR ATTENTION

also, does forging acting on the approval or denial of the service roadblock any other action? (should we be disabling buttons till a user addresses the request? )

🤔

jjlangholtz:
@AllenBW thanks for the feedback!

I didn't really spend any time figuring out where the buttons should go because I knew I would get it wrong 😄. For consistency with the rest of the interface I was thinking of either putting the buttons in the corner like the service details page. I know it isn't quite the same since this actually has input and the service details page just has buttons:


screen shot 2016-11-07 at 10 07 56 am



> also, does forging acting on the approval or denial of the service roadblock any other action? (should we be disabling buttons till a user addresses the request? )

AFAIK nothing needs to be disabled while the request is pending. In my testing the "My Services" tab correctly shows services once the request has been approved. When the request is denied or pending the requesting user is unable to take any actions related to the service.

chriskacerguis:
@jjlangholtz yeah, we can we work on that UI a little bit?

chriskacerguis:
@jjlangholtz can you add a new screenshot?

jjlangholtz:
The UI has not been updated yet, because there are at least three more buttons that need to be added (copy, edit, and delete requests) I think it is worthwhile having a discussion on the UX.

AllenBW:
To be honest, unless this is a **need to have it now** kinda feature, we should hold off on the ui until we have that holistic approach for requests parity.

chriskacerguis:
@AllenBW @jjlangholtz what's the verdict here?

jjlangholtz:
@chriskacerguis I think because this involves adding a significant amount of functionality that does not yet exist in the SUI (approve, deny, delete, copy, edit requests), this should be put on hold until after a design/UX story.

AllenBW:
ditto 👑

On Tue, Nov 15, 2016 at 11:27 AM, Joshua Langholtz <notifications@github.com

PR #326, opened 2016-11-08
508 compliance gulp

chalettu:
added a gulp task that tests for 508 compliance


chriskacerguis:
@chalettu ^

chalettu:
Testing /Users/chrishale/Code/manageiq-ui-service/client/app/components/ownership-service-modal/ownership-service-modal.html

ERROR Section508.D.HeadingOrder
Line:1 Col:0
The heading structure is not logically nested. This h4 element appears to be the primary document heading, so should be an h1 element.
--------------------
<h4 class="modal-title" id="myModalLabel" translate="">Set Service Ownership</h4>

NOTICE Section508.N.Errors
Line:1 Col:0

chalettu:
The output seems to just be regular text , The format can be tweaked in a few different ways if we would like to change it @AllenBW

PR #299, opened 2016-10-28
Address gettext on primary and secondary navigation items

AparnaKarve:
Navigation text and tooltip for primary and secondary navigation items shows up in English on non `en` locales.

PR https://github.com/ManageIQ/manageiq-ui-service/pull/277 probably meant to resolve this issue, but it did not address it completely.

https://bugzilla.redhat.com/show_bug.cgi?id=1389655

Before:
screen shot 2016-10-28 at 4 39 36 pm

After:
screen shot 2016-10-28 at 4 39 11 pm

PR #327, opened 2016-11-09
Reload the whole page after switching the language

himdel:
When moving to vertical menu, we've lost the ability to translate menu items on the fly, the patternfly component can't do it by itself.

This got fixed by #299, but it only works once, because it mutates the original object, so changing from language A to language B tries to find the language A string in the catalog (and thus only works if A=English). (Steps to reproduce - have the language set to English, switch to German (the menu changes), switch to French (the menu stays in German without this PR).)

Unfortunately we can't really *not* mutate the original object and show a copy, because then it stops getting the badge count updates every minute (only the original gets updated).

So.. triggering a full browser reload after switching the language, meaning the state gets recreated anew, in English, and is only ever translated once. The side effect is that switching the language takes a bit longer because the whole page reloads, but..

https://bugzilla.redhat.com/show_bug.cgi?id=1391742

### DIFF
--- a/client/app/components/language-switcher/language-switcher.directive.js
+++ b/client/app/components/language-switcher/language-switcher.directive.js
@@ -52,7 +52,9 @@
       vm.switch = function(code) {
         Language.setLocale(code);
         Language.save(code);
-        $state.reload();
+
+        // FIXME(#328) this should be $state.reload(), once we refactor Navigation.items so that we can work with a translated copy and still get badge count updates
+        window.document.location.href = window.document.location.href;
       };
     }
   }
--- a/client/app/components/language-switcher/language-switcher.directive.js
+++ b/client/app/components/language-switcher/language-switcher.directive.js
@@ -20,7 +20,7 @@
     return directive;
 
     /** @ngInject */
-    function LanguageSwitcherController($scope, gettextCatalog, Language, lodash, $state) {
+    function LanguageSwitcherController($scope, gettextCatalog, Language, lodash, $state, $window) {
       var vm = this;
       vm.mode = vm.mode || 'menu';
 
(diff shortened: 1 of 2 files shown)


himdel:
wip because it still breaks when coming from a translated login screen..

himdel:
@AllenBW definitely not for euwe but.. I think we may want to refactor `Navigation.items` so that we can work with a translated copy instead of rewriting the strings in the original, and still get the count updates, WDYT?

client/app/components/language-switcher/language-switcher.directive.js
```diff
@@ -52,7 +52,9 @@
       vm.switch = function(code) {
         Language.setLocale(code);
         Language.save(code);
-        $state.reload();
+
+        // FIXME this should be $state.reload(), once we refactor Navigation.items so that we can work with a translated copy and still get badge count updates
```
AllenBW: down with this, we need it, will do the job for now

AllenBW:
@himdel agree with this. translations should behave the same everywhere.  Lemme noodle on an approach and write up a bug for posterity?

himdel:
@AllenBW sure, thanks!
Going home for the day, I'll try to finish the transition from login screen tomorrow..

himdel:
Updated the transition from the login screen to do the same thing, fixed `window` to use `$window` instead .. removing wip :)

@AllenBW can you re-review please? :)

himdel:
(Added `blocker` because this also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1391142)

chessbyte:
Euwe Backport details:
```sh
$ git log -1
commit 84e8b93dd6367ace6d27757a04f446d0bd88ec98
Date:   Thu Nov 10 08:31:26 2016 -0600

    Merge pull request #327 from himdel/bz1391742-left-menu-i18n
    
    Reload the whole page after switching the language
    (cherry picked from commit c4638ff380da1daf5f3e9604d55a55c03ad2db0d)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1391142
    https://bugzilla.redhat.com/show_bug.cgi?id=1391742
```

PR #338, opened 2016-11-14
[WIP] Count retired/retiring services for dashboard client-side

jjlangholtz:
This change fixes a bug where the dashboard would fail to render when a
service was scheduled to retire between today and 30 days from today.
The error was due to a request to to service collections API with filter
parameters that are not supported. Currently, filter expressions using
the `>=` and `<=` operators must have numeric operands. The dashboard
has a widget that counts the number of services scheduled to retire soon
and uses filter parameters with datetime strings, causing the issue.

### DIFF
--- a/client/app/states/dashboard/dashboard.state.js
+++ b/client/app/states/dashboard/dashboard.state.js
@@ -22,9 +22,7 @@
           requireUser: true,
         },
         resolve: {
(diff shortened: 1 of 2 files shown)


jjlangholtz:
There is a workaround that allows filtering to work properly under certain conditions, closing in favor of #341

PR #341, opened 2016-11-15
Fix links in dashboard for service counts and filters

jjlangholtz:
Fixes a bug for the dashboard view caused by a 500 error when making a
request to the API using an unsupported filter expression. An issue is
already filed with the ManageIQ repo. In addition, this change includes
various enhancements to the dashboard for services.

* Make links to filtered service views more apparent with a cursor
* Request services retiring soon in UTC time to prevent inconsistencies
across time zones
* Fix `current` services filter by removing `retired === false` check.
Services retiring soon are `retired = false` and would show up with that
check.

https://github.com/ManageIQ/manageiq/issues/12648

### DIFF
--- a/client/app/states/dashboard/_dashboard.sass
+++ b/client/app/states/dashboard/_dashboard.sass
@@ -43,6 +43,9 @@
       border-left: 0
       margin: 0
 
+      a
+        cursor: pointer
+
       .pficon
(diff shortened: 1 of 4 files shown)


chriskacerguis:
@jjlangholtz can / should we add tests?

PR #343, opened 2016-11-15
Update remove service to use delete request

jjlangholtz:
Fix the service delete action in the service detail view. Instead of
making a post request with the `retire` action, use the delete request
for the associated resource.


chriskacerguis:
@jjlangholtz do we need tests here? (not sure if they are already tests for this)

PR #350, opened 2016-11-22
Prevent selecting invalid warn options

jjlangholtz:
When scheduling a retirement date in the future, a group of options are
available for selecting when a warning event should trigger (1, 2, 3, or
4) weeks prior the scheduled retirement date.

This change limits the selectable options to only values that make
sense. For example, you should not be able to select a warning 4 weeks
prior to retirement if you set retirement to tomorrow.

@miq-bot add_label enhancement, euwe/no

### DIFF
--- a/client/app/components/retire-service-modal/retire-service-modal-service.factory.js
+++ b/client/app/components/retire-service-modal/retire-service-modal-service.factory.js
@@ -8,6 +8,7 @@
   function RetireServiceFactory($modal) {
     var modalService = {
       showModal: showModal,
+      FactoryController: RetireServiceModalController,
     };
 
     return modalService;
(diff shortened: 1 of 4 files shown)


client/app/components/retire-service-modal/retire-service-modal-service.factory.js
```diff
@@ -62,10 +63,18 @@
       { value: 21, label: __('3 Weeks') },
       { value: 28, label: __('4 Weeks') },
     ];
+    vm.visibleOptions = [];
 
     activate();
 
     function activate() {
+      $scope.$watch('vm.modalData.resource.date', function(date) {
+        var daysBetween = moment(date).diff(moment(), 'days');
```
chriskacerguis: Do we need to translate the "days" string?

jjlangholtz: No this is just an identifier for moment's `diff` method to indicate unit of measure:

http://momentjs.com/docs/#/displaying/difference/

It is not a string the user sees.

PR #353, opened 2016-11-28
Update api datetime filter to use acceptable operators

jjlangholtz:
The filter API was recently updated to support `>` and `<` operators for
datetime filters. The previously used `>=` and `<=` now correctly result
in a bad request.

https://github.com/ManageIQ/manageiq/pull/12660

@miq-bot add_label bug

### DIFF
--- a/client/app/states/dashboard/dashboard.state.js
+++ b/client/app/states/dashboard/dashboard.state.js
@@ -98,10 +98,10 @@
       return undefined;
     }
     var currentDate = new Date();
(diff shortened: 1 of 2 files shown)


simaishi:
@miq-bot remove_label euwe/no
@miq-bot add_label euwe/yes

This will be backported via https://github.com/ManageIQ/manageiq-ui-service/issues/411

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq-ui-service/pull/411

PR #360, opened 2016-11-29
Restore notification count to icon in lieu of number

AllenBW:
@chriskacerguis @serenamarie125

PR #347, opened 2016-11-21
Resolve issue of notification badge count missing

AllenBW:
https://bugzilla.redhat.com/show_bug.cgi?id=1396887

@chriskacerguis @chessbyte 
## Before
ssui-notifications-number

## After
screen shot 2016-11-21 at 9 35 43 am

PR #361, opened 2016-11-29
Revert notification count

AllenBW:
similar to https://github.com/ManageIQ/manageiq-ui-service/pull/360
except for master rather than euwe

@chriskacerguis


Fryguy:
Reverts https://github.com/ManageIQ/manageiq-ui-service/pull/347

PR #366, opened 2016-12-01
[EUWE] Hide Power Status and Power Operation buttons in the List View

AparnaKarve:
Hide Power Status and Power Operation buttons in the List View until we have a robust backend solution to display the Power Status and the Power Op buttons.

Before:
screen shot 2016-12-01 at 10 46 04 am

After:
screen shot 2016-12-01 at 10 44 27 am

While I was at it (hiding Power Op status & buttons), I have restored the `Created` field width back to 2 so it does not get truncated, like it did before with the Power Op buttons displayed.

https://bugzilla.redhat.com/show_bug.cgi?id=1402993

### DIFF
--- a/client/app/states/services/list/list.html
+++ b/client/app/states/services/list/list.html
@@ -1,12 +1,14 @@
 <div pf-toolbar id="serviceToolbar" config="vm.toolbarConfig"></div>
 
 <div class="list-view-container">
+  <!-- TODO: The Power Operation buttons would be hidden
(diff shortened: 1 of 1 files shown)


AparnaKarve:
@miq-bot add_label bug,ui,euwe/yes,blocker

AparnaKarve:
@chessbyte This is the equivalent Upstream PR : https://github.com/ManageIQ/manageiq-ui-service/pull/368

syncrou:
@AparnaKarve -  👍 LGTM

PR #367, opened 2016-12-01
Remove eot fontawesome webfonts

jjlangholtz:
The Embedded OpenType (EOT) font file type is only used by IE versions 8
and up. However, IE10 and higher support the same open format that other
browsers use. Because we support IE10 and higher these fonts do not need
to be loaded.

This removes the console warning message for attempting to download
an unknown font file type on both Chrome and Firefox.


chriskacerguis:
@jjlangholtz what's the PT card for this story?

jjlangholtz:
@chriskacerguis Added a chore story here https://www.pivotaltracker.com/story/show/135359297

PR #342, opened 2016-11-15
List all services

AllenBW:
## Does not include (work left to do in other prs): 
* Configuration, Policy, Life cycle buttons (and functionality) 
* Propose a (more so) permanent solution for a footer toolbar

## Includes (what it does)
* new list view fields and layout that scale with browser viewport 
* Adds pagination and limit functionality (tests for pagination component)
* Load screen component
* Power state component
* Update badge count query, no longer requests extraneous info
* Update Collections api to handle additional functionality
* General refactoring and deduplication
* Lots of 🌮 🌮 🌮 

https://www.pivotaltracker.com/story/show/133840391

## The Good Stuff
list

### Smaller
screen shot 2016-11-22 at 4 00 16 pm

### Smallest
screen shot 2016-11-22 at 4 00 36 pm

**Hey you beautiful 🍑s would appreciate a look, thoughts, suggestions, inputs**
@chriskacerguis @jjlangholtz @himdel @serenamarie125 @jntullo @chalettu

PR #368, opened 2016-12-01
Hide Power Status and Power Operation buttons in the List View

AparnaKarve:
Hide Power Status and Power Operation buttons in the List View until we have a robust backend solution to display the Power Status and the Power Op buttons.

Before:
screen shot 2016-12-01 at 2 20 55 pm

After:
screen shot 2016-12-01 at 1 56 13 pm

https://bugzilla.redhat.com/show_bug.cgi?id=1391685

### DIFF
--- a/client/app/components/explorer/explorer.html
+++ b/client/app/components/explorer/explorer.html
@@ -21,10 +21,12 @@
 
 <div ng-if="vm.viewType == 'listView'" class="col-md-12 list-view-container">
   <loading status="vm.loading"></loading>
+  <!-- TODO: The Power Operation buttons would be hidden
+       TODO: while we wait on a robust backend solution on how to retrieve Power Status for Services in the List View-->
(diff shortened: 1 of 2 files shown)


AparnaKarve:
This is the Upstream equivalent of the Euwe PR: https://github.com/ManageIQ/manageiq-ui-service/pull/366.
Upstream code has significantly changed since https://github.com/ManageIQ/manageiq-ui-service/pull/342 which has "componentized" most UI elements and has also added pagination to the Service List view -- which means going forward, Upstream and Euwe PRs cannot be identical for most cases.

chessbyte:
Euwe BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1402993

chessbyte:
Backported to Euwe via https://github.com/ManageIQ/manageiq-ui-service/pull/366

PR #330, opened 2016-11-10
Power operations in Service Detail view

AparnaKarve:
https://github.com/ManageIQ/manageiq-ui-service/pull/271 added the Service Power Operation feature in the Services List view.

This PR adds the same feature in the Service Detail view.

- A new Service `PowerOperations` was created that would be injected in both the Service list controller as well as the Service detail controller.
- Buttons in the top right corner were rearranged to avoid making the page look overly busy (with too many buttons). 
- Also added a `Power State` field to indicate the current Power status

Screenshots
-------------
Before:
screen shot 2016-11-10 at 11 02 30 am

After:
screen shot 2016-11-10 at 11 00 28 am

screen shot 2016-11-10 at 10 57 08 am

screen shot 2016-11-10 at 10 59 20 am

PR #371, opened 2016-12-02
[EUWE] Power operations in Service Detail view

AparnaKarve:
This PR adds the Power Operations feature in the Service Detail view.

- A new Service PowerOperations was created that would be injected in both the Service list controller as well as the Service detail controller.
- Buttons in the top right corner were rearranged to avoid making the page look overly busy (with too many buttons).
- Also added a Power State field to indicate the current Power status

This is the Euwe equivalent of the upstream PR https://github.com/ManageIQ/manageiq-ui-service/pull/330

https://bugzilla.redhat.com/show_bug.cgi?id=1391685

### DIFF
--- a/client/app/services/poweroperations.service.js
+++ b/client/app/services/poweroperations.service.js
@@ -20,7 +20,7 @@
       powerOperationTimeoutState: powerOperationTimeoutState,
       powerOperationStartTimeoutState: powerOperationStartTimeoutState,
       powerOperationStopTimeoutState: powerOperationStopTimeoutState,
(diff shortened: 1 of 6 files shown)


AparnaKarve:
@chriskacerguis For some strange reason `Hound` pointed out a few linting violations on this Euwe PR, but for the equivalent upstream PR https://github.com/ManageIQ/manageiq-ui-service/pull/330, with identical code changes there were no violations.
Not sure what to make of it.

AparnaKarve:
@miq-bot add_label bug,blocker,euwe/yes

PR #381, opened 2016-12-09
[WIP] Create tagEditor component and associated service

jjlangholtz:
`tagEditor` to be used in modals for single and batch editing of tags.
The component will need to:
- [ ] Have a binding for common tags of resources that need tags edited
- [ ] Manage a tag assignments collection
- [ ] Cache/hold available tag categories and options
- [ ] Provide a callback after dismissing tags from UI
- [ ] Two-way bind tag assignments for both single and multiple value
categories

Add `taggingService` for higher-level interactions with CollectionsApi.
The service will need to:
- [ ] Get all tag categories and associated value options
- [ ] Post tag assignments (involves assigning and unassigning tags)

@miq-bot add_label wip, euwe/no

### DIFF
--- a/client/app/components/tag-editor/tag-editor.component.html
+++ b/client/app/components/tag-editor/tag-editor.component.html
@@ -1,0 +1,25 @@
+<div class="row">
+  <div class="col-md-6">
+    <input type="text" class="form-control" placeholder="Select a Category"
+      ng-model=""
+      uib-typeahead=""
+      typeahead-min-length="0">
+  </div>
+  <div class="col-md-6">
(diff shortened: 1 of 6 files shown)


dtaylor113:
Hi @jjlangholtz ,

I'm not sure if we're doing the same component.  There is a tagging directive at: client/app/components/tagging

It is used to tag blueprints and service items (on the canvas):

image

image

It handles adding and removing tags, as well as handles single-value tags.

jjlangholtz:
Closing in favor of #402

PR #377, opened 2016-12-06
Pull in the real power_state column off of service

syncrou:
This PR pulls in the service.power_state directly from the virtual power_state column vs the cached version previously exposed in the options hash.

Dependent on: ManageIQ/manageiq#12963

https://bugzilla.redhat.com/show_bug.cgi?id=1396728

PR #382, opened 2016-12-09
[EUWE] Pull in the real power_state column off of service

AparnaKarve:
This PR pulls in the service.power_state directly from the virtual power_state column vs the cached version previously exposed in the options hash.

https://bugzilla.redhat.com/show_bug.cgi?id=1396728

This is Upstream equivalent of https://github.com/ManageIQ/manageiq-ui-service/pull/377

### DIFF
--- a/client/app/services/poweroperations.service.js
+++ b/client/app/services/poweroperations.service.js
@@ -24,58 +24,58 @@
     };
 
     function powerOperationUnknownState(item) {
-      return item.powerState === "" && item.powerStatus === "";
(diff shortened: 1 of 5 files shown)


AparnaKarve:
It's OK to ignore those hound violations (they are not valid. I always see them only on the Euwe branch)

@chessbyte This is the Euwe equivalent of the Upstream PR https://github.com/ManageIQ/manageiq-ui-service/pull/377

chessbyte:
@chriskacerguis please review before I merge

PR #383, opened 2016-12-09
Add Designer -> Catalogs view

jeff-phillips-18:
This PR adds the Designer Catalogs view to show a listing of the
current service catalogs along with each catalogs list of catalog items
(service templates).

Finishes 133964531


jeff-phillips-18:
Screenshots:

image

image

image

jeff-phillips-18:
@beanh66 @serenamarie125

tests/catalogs/catalogs-list.component.spec.js
```diff
@@ -0,0 +1,253 @@
...
+  };
+
+  beforeEach(function () {
+    Session.create({
+      auth_token: 'b10ee568ac7b5d4efbc09a6b62cb99b8',
+    });
+    $httpBackend.whenGET('').respond(200);
+
+    $scope.designerCatalogs = [
+      {
```
chriskacerguis: This looks to be mock data, can you put the mock data in tests/mock/ ?

serenamarie125:
@jeff-phillips-18 is it possible to line up the service icon with the catalog name ( horizontally left aligned) ?  And then have both descriptions start at the same place ?  Although i remember you saying there may be a PF bug ... is this what's causing this?

Also wondering if you can do the compacted spacing ( less top and bottom margin ) like you have done in the Monitoring/Alerts area of the Ops UI.

beanh66:
@serenamarie125 I agree, just one clarification....you would want the catalog description to start where the service description starts, correct?

jeff-phillips-18:
@serenamarie125 We don't use compact styling anywhere else in SUI only on a couple in Ops UI. I think we should stay consistent. If we go compact here, we should look at going compact in the other list views as well.

serenamarie125:
@jeff-phillips-18 arghhh ... compact styling should be the PatternFly default IMO.  I'll create a story about changing everything to compact styling throughout SUI then.  Leave this as is for now.

jeff-phillips-18:
@chriskacerguis Moved mock data to mock json files

@serenamarie125 @beanh66 Updated catalog items field alignment:
image

Also, fixed pre-existing eslint and sasslint errors. gulp vet is now clean.

chriskacerguis:
Just waiting on Hound, and then I will merge

chriskacerguis:
Hound seems stuck.  Merging.

PR #388, opened 2016-12-13
[EUWE] Lock is-number-like transient dependency

jjlangholtz:
This transient dependency is currently released with a broken version and it needs to be locked to a previously working version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "bower": "^1.3.0",
-    "browser-sync": "^2.8.1",
+    "browser-sync": "2.8.3",
     "chai": "^2.1.1",
     "chai-as-promised": "^4.3.0",
     "chalk": "^1.0.0",


chriskacerguis:
Looks fine with me if passes

chriskacerguis:
@jjlangholtz build is failing, can you please address

jjlangholtz:
@chriskacerguis Build is passing now, still testing it out locally though.

jjlangholtz:
@chriskacerguis Ran into several problems trying to lock dependencies but eventually found this to be the simplest solution. Unfortunately, this fix does not prevent the same kind of issue happening with another transient dependency.

chriskacerguis:
@jjlangholtz ok.  I made a card for this (https://www.pivotaltracker.com/story/show/136080951) and we can chat about it tomorrow.  I'll wait for the build to finish, then will ping Oleg.

chriskacerguis:
@chessbyte ready for merge.

PR #390, opened 2016-12-14
Reload login screen to close the modal dialog

AparnaKarve:
If a modal dialog is left active until the user session times out, we get a login screen with the modal still visible as shown below --
screen shot 2016-12-14 at 11 38 17 am

The PR addresses the above issue. 

https://www.pivotaltracker.com/n/projects/1914499/stories/134431669

### DIFF
--- a/client/app/states/login/login.state.js
+++ b/client/app/states/login/login.state.js
@@ -36,6 +36,10 @@
       password: API_PASSWORD,
     };
 
+    if ($window.location.href.includes("?timeout")) {
+      Notifications.message('danger', '', __('Your session has timed out.'), true);
+    }
+
     if (Session.privilegesError) {
       Notifications.error(__('User does not have privileges to login.'));
     }
--- a/client/app/config/authorization.config.js
+++ b/client/app/config/authorization.config.js
@@ -11,7 +11,7 @@
     $httpProvider.interceptors.push(interceptor);
 
     /** @ngInject */
-    function interceptor($injector, $q) {
+    function interceptor($injector, $q, $window) {
       return {
         response: response,
         responseError: responseError,
@@ -50,14 +50,14 @@
           }
 
           Session.destroy();
-          $state.go('login');
+          $window.location.href = $state.href('login');
         }
       }
     }
   }
 
   /** @ngInject */
-  function init($rootScope, $state, Session, $sessionStorage, logger, Language, ServerInfo, ProductInfo) {
+  function init($rootScope, $state, Session, $sessionStorage, logger, Language, ServerInfo, ProductInfo, $window) {
     var unregisterStart = $rootScope.$on('$stateChangeStart', changeStart);
     var unregisterError = $rootScope.$on('$stateChangeError', changeError);
     var unregisterSuccess = $rootScope.$on('$stateChangeSuccess', changeSuccess);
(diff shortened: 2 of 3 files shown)


chriskacerguis:
@AllenBW could you please take a look?

.eslintrc.json
```diff
@@ -14,7 +14,8 @@
   "globals": {
     "flowchart": true, // Blueprints support lib
     "__": true, // Zanata parsing cue
-    "N_": true // Zanata parsing cue
+    "N_": true, // Zanata parsing cue
+    "$window": true
```
jjlangholtz: I think you can avoid declaring `$window` as a global if you dependency inject it at `client/app/config/authorization.config.js:60`.

AparnaKarve: Ok, that might address the CC error `$window is not defined`  in `client/app/config/authorization.config.js`
What about the same CC error in `client/app/states/logout/logout.state.js`?

jjlangholtz: Ah, I cannot figure out how to show the old CC errors, was it the same message for both issues?

Hmm I don't see why it would say `$window` not defined in the `logout.state`, running it locally it looks like both ESLint issues were in `authorization.config.js`.

AllenBW:
👀

.eslintrc.json
AparnaKarve: let me remove that commit -- 0911e7c134d4efe823f44dbab17b133c7444d30e so we know the exact CC error

AllenBW:
cc is whining about the absence of `$window` from line 60.. the `function init(`

inject it there for cc 💚 !

.eslintrc.json
AllenBW: this shouldn't be necessary, $window is an angular service, if properly injected it'll be defined, cc wont complain

AparnaKarve:
>$window needs to be properly injected, otherwise it won't result in the desired change, adding it as an eslint global is not the solution

@jjlangholtz @AllenBW Yes, I misunderstood the CC error -- I thought that it was complaining about an undefined `$window` despite it being injected appropriately. 
Turns out that it needed dependency injection in two places in `client/app/config/authorization.config.js `, whereas I only had it in one place.
All taken care of now. Thanks!

/cc @chriskacerguis

AparnaKarve:
Added another commit to ensure that the timeout message appears on the login screen when the user session times out.
screen shot 2016-12-15 at 8 16 47 am

chriskacerguis:
@AllenBW thoughts?

client/app/states/login/login.state.js
```diff
@@ -36,6 +36,10 @@
       password: API_PASSWORD,
     };
 
+    if ($window.location.href.includes("?timeout")) {
+      Notifications.message('danger', '', __('Your session has timed out.'), true);
```
AllenBW: `EventNotifications.error(__('Your session has timed out.'));`

should be used here, while you're touching this file might as well remove the other use of `Notifications` (depreciated in favor of `EventNotifications`)

AparnaKarve: Makes sense! Thanks @AllenBW

AparnaKarve: @AllenBW Could `Notifications` and `EventNotifications` be two separate things? Reason I ask is because `EventNotifications` probably expects the authentication token to be already present.
When I replace all `Notifications` calls with `EventNotifications` in that file, I get this at login 
screen shot 2016-12-15 at 9 17 01 am

AllenBW: Oh snap, really?  Yeah yeah, this makes sense.  Ignore my previous comment then, all good with this, great job! 🌮 

@chriskacerguis  👍

PR #392, opened 2016-12-15
Updated detail-reveal component template and css based on UI team feedback

chalettu:


chalettu:

chriskacerguis:
@chalettu can you please add screenshots?

jeff-phillips-18:
Please run 'gulp vet' to catch sass lint issues.

chalettu:
closed_stateClosed State. 
Changes Implemented: Only show arrows and arrow cell border if a row has more details.  
open_state
Open State .  Changes implemented.  Grey background on row when row selected.  Close Button X implemented.  
Implemented indent of content inside of detail component

chriskacerguis:
@serenamarie125 thoughts?

serenamarie125:
@chalettu thanks for making the changes!  Couple of nits ...

- is it possible to add borders around the expanded panel as shown here http://www.patternfly.org/pattern-library/content-views/list-view/#/_design?
- is there any way to have the width of the first column in the wider so that the headings aren't wrapping?
- I'm assuming the ".." and "..." is fake data ?

chalettu:
sample screenshot
@serenamarie125 I added more of a pronounced border to when the component is expanded.  The content that was in the other sample image is actually not part of this PR and was just mock data.   The content that goes inside this component can be changed to fit whatever is needed inside it.  I will be sure when we do start doing PR's for detail content that we do have the tabular data more stretched out appropriately.

chriskacerguis:
@cshinn Could you please review?

cshinn:
@chalettu I don't know that the drop shadow etc. is necessary. In our current implementations of expandable lists, the border is still just a simple 1px solid border, but it is changed to blue as seen here. Similar styling in this case would lead to more consistency within our products. 
screen shot 2016-12-16 at 12 39 46 pm

chalettu:
Ok, I can work on doing a blue border.  The link Serena sent, I looked at the sample and it had a shadow on the border on the bottom so I just tried to mimick that.  I'll put in a blue border and add another snapshot.

cshinn:
@chalettu Gotcha, in this case, there's an inconsistency between the design and implementation pages for the pattern. At the moment, I'd say that conforming to the current implementation is more important. I'll look into getting that inconsistency resolved.

chalettu:
sample blue border
@cshinn Here is the updated view to support a blue border around the active revealed component.  I color matched the code that patternfly is using for the listview as well.

cshinn:
@chalettu Border looks great! The expanded section could use some more left padding to make it appear subordinate to the main row. Should be about 20px indented from the "name" in this case.

chalettu:
final
@cshinn Does this work?

cshinn:
Spectacular! Looks great

PR #397, opened 2016-12-16
Enhancement/134032877 approve deny requests

AllenBW:
https://www.pivotaltracker.com/story/show/134032877

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -9,7 +9,7 @@ engines:
     config:
       languages:
         javascript:
-          mass_threshold: 100
+          mass_threshold: 130
 ratings:
   paths:
   - client/app/**
--- a/client/app/components/request-explorer/request-explorer.component.js
+++ b/client/app/components/request-explorer/request-explorer.component.js
@@ -51,9 +51,9 @@
         },
         configuration: [
           {
-            title: __('Configuration'),
-            actionName: 'configuration',
-            icon: 'fa fa-cog',
+            title: __('Lifecycle'),
+            actionName: 'lifecycle',
+            icon: 'fa fa-recycle',
             actions: [
               {
                 icon: 'fa fa-check',
--- a/client/app/components/request-explorer/request-explorer.component.js
+++ b/client/app/components/request-explorer/request-explorer.component.js
@@ -56,17 +56,17 @@
             icon: 'fa fa-cog',
             actions: [
               {
-                icon: 'pf pficon-ok',
-                name: __('Approve Selected'),
+                icon: 'fa fa-check',
+                name: __('Approve'),
                 actionName: 'approve',
-                title: __('Approve Selected'),
+                title: __('Approve'),
                 actionFn: approveRequests,
                 isDisabled: false,
               }, {
-                icon: 'pf pficon-error-circle-o',
-                name: __('Deny Selected'),
+                icon: 'fa fa-ban',
+                name: __('Deny'),
                 actionName: 'deny',
-                title: __('Deny Selected'),
+                title: __('Deny'),
                 actionFn: denyRequests,
                 isDisabled: false,
               },
@@ -74,6 +74,7 @@
             isDisabled: false,
           },
         ],
+        listActionDisable: listActionDisable,
       });
 
       vm.fetchData = fetchData;
@@ -242,6 +243,10 @@
 
     function denyRequests() {
       ProcessRequestsModal.showModal(vm.selectedItemsList, "deny");
+    }
+
+    function listActionDisable(config, items) {
+      items.length <= 0 ? config.isDisabled = true : config.isDisabled = false;
     }
   }
 })();
--- a/client/app/components/request-explorer/request-explorer.component.js
+++ b/client/app/components/request-explorer/request-explorer.component.js
@@ -9,7 +9,8 @@
     });
 
   /** @ngInject */
-  function ComponentController($state, CollectionsApi, RequestsState, ListView, $filter, lodash, Language, EventNotifications) {
+  function ComponentController($state, CollectionsApi, RequestsState, ListView, $filter, lodash, Language, EventNotifications,
+                               ProcessRequestsModal) {
     var vm = this;
 
     vm.$onInit = function() {
@@ -22,12 +23,15 @@
         selectedItemsList: [],
         limitOptions: [5, 10, 20, 50, 100, 200, 500, 1000],
         listConfig: {
-          selectItems: false,
-          showSelectBox: false,
+          multiSelect: true,
           selectionMatchProp: 'id',
+          onCheckBoxChange: selectItem,
           onClick: handleRequestClick,
         },
         toolbarConfig: {
+          actionsConfig: {
+            actionsInclude: true,
+          },
           filterConfig: {
             fields: getRequestFilterFields(),
             resultsCount: vm.listDataCopy ? vm.listDataCopy.length : 0,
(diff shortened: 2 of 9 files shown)


AllenBW:
@serenamarie125 @cshinn @beanh66

ss of current state, gonna write some tests if this the below is okish 
screen shot 2016-12-19 at 12 14 22 pm
screen shot 2016-12-19 at 12 15 09 pm
screen shot 2016-12-19 at 12 14 52 pm
screen shot 2016-12-19 at 12 14 37 pm

client/app/components/request-explorer/request-explorer.component.js
```diff
@@ -46,6 +50,31 @@
...
+                actionName: 'deny',
+                title: __('Deny Selected'),
+                actionFn: denyRequests,
+                isDisabled: false,
+              },
+            ],
+            isDisabled: false,
+          },
+        ],
       });
```
jeff-phillips-18: Are Approve / Deny actions really considered configuration actions?

client/app/components/request-explorer/request-explorer.component.js
```diff
@@ -22,12 +23,15 @@
         selectedItemsList: [],
         limitOptions: [5, 10, 20, 50, 100, 200, 500, 1000],
         listConfig: {
-          selectItems: false,
-          showSelectBox: false,
+          multiSelect: true,
```
jeff-phillips-18: Don't need to set multiSelect when using check boxes.

client/app/components/request-explorer/request-explorer.component.js
AllenBW: @jeff-phillips-18 you thinking more lifecycle?

beanh66:
@AllenBW Overall looks great, but I have a few nits as well. 

(1) We should remove color from the icons under the button groups. I also wanted to confirm these are the same icons used in the Ops UI today because I couldn't tell on my instance. Today are these options under the Lifecycle or Configuration group? 

(2) Instead of using "OK," can we have the button either say "Approve" or "Deny" depending which modal is presented? 

(3) For invalid requests, it doesn't look like you have any buttons at the bottom. I think we need a "Close" button to get rid of the modal. I'm not positive how we want to format the warning here.... @serenamarie125 do you have a thought on that?

serenamarie125:
@AllenBW can I give you feedback by tomorrow morning ?  I'm off today but would like to check on some artifacts to make sure that I reply per consistency.

AllenBW:
@serenamarie125 sure!

@beanh66 
1. same icons, not sure if they exist without color : https://www.patternfly.org/styles/icons/#_
2. hesitant to do, breaks consistent behavior throughout app
3. close is in the upper right, warning is pf-icon

serenamarie125:
@AllenBW right now both the Ops and SUI UIs are consistently inconsistent.  In the beginning of January, let's have a pow wow that we mentioned last week gathering everyone to talk about the direction we would like to go, review our suggestions, see where we have consensus and move towards consistency throughout :)  I think that @beanh66 's  points above are addressing many of those.

Per colleen's #3, We want an action button in addition to the "x" button in the upper right.

Additionally (i'll continue with numbers),
4. for button labels, please remove "Selected"
5. I'm going to reply tomorrow with some suggested working for the Invalid Request message.  The suggestion is to stray away from saying "please" in prompts and messages, and keep them as concise as possible.
6.  Do you think that it might be better to use a smaller version of the service icon and list services on rows rather than in this large format ?  Since requests really don't have the ability to be a "real" quad icon, i feel like this is a waste a space.

beanh66:
@AllenBW That makes sense. I think @serenamarie125 has a good plan for getting through the inconsistencies together so we can wait on some of these wording or icon changes until we meet 👍

AllenBW:
@serenamarie125 @beanh66   Updated as per feeeeeeeeeeeeeedback 😋 

screen shot 2016-12-20 at 10 18 19 am
screen shot 2016-12-20 at 10 17 58 am
screen shot 2016-12-20 at 10 17 39 am
screen shot 2016-12-20 at 10 17 27 am
screen shot 2016-12-20 at 10 17 18 am

jeff-phillips-18:
Couple of things here. 

We should use the same format as the delete confirmation dialog. Show the items to be approved or denied but after 5 use a collapsable panel. I would also prefer to see list view rather than card view here. The icon isn't really the most important piece is it? the name is truncated and I would expect that to be the most relevant piece of info.

For the Approve dialog, 'Reason' seems inappropriate, maybe "Comment' would be better?

Still don't like Configuration, Life Cycle would be better.

AllenBW:
@jeff-phillips-18 I asked if lifecycle was more appropriate to no response thanks for providing it now, will make the change.

Icon list has truncated name revealed in entirety by tooltip, used here for consistency.  Fine with implementing a list view, but it should then be done across all explorer action modals, consistency n all.  

Reason is the verbiage used by the api and is a mandatory field for both approve and deny actions.

jeff-phillips-18:
@AllenBW Sorry, I didn't respond I was waiting for other feedback as well. I'd agree with using the list view in the action modals across the board. Regardless of the API verbiage, I think it would be better for the user to see "Comment" rather that "Reason" for an approval. Maybe other UXD thoughts here?

AllenBW:
@miq-bot remove_label wip

beanh66:
@jeff-phillips-18 I agree a list of the services seems more practical. We would need to do this in a number of places, including the tags area Josh is working on. I also agree that "Comment" makes more sense here, my only concern is that comments are typically optional and @AllenBW makes a good point about the Reason field being required in both cases. @serenamarie125 or @cshinn what do you think?

chriskacerguis:
@AllenBW is this ready to go?

AllenBW:
@chriskacerguis yeah lets do it, the changes mentioned in feedback exceed the scope of this pr

.codeclimate.yml
```diff
@@ -9,7 +9,7 @@ engines:
     config:
       languages:
         javascript:
-          mass_threshold: 100
+          mass_threshold: 130
```
chriskacerguis: @AllenBW Why are we upping the threshold?

AllenBW: action modals all look the same, but they can't not (quite yet) look the same, hoping the work imma doing today is gonna address that

chriskacerguis:
@serenamarie125 @beanh66 @cshinn good from UX team?

serenamarie125:
@AllenBW sorry it took me so long!  Please use Note rather than Reason, and if it is required input, indicate it as such with a red asterisk.  You can see the guideline here for mandatory input https://github.com/cshinn/patternfly-design/blob/690ef41ac6afaa7b15dff3f148699e48e27873e4/pattern-library/forms-and-controls/forms/design/design.md

serenamarie125:
looks like you decided to address other issues later per button labels, etc ?

AllenBW:
@serenamarie125 so making the asterisk red isn't going to be a simple task (how funny is that) we presently use pf-form-group which constricts the labels to being all strings **which** means unless you want the asterisk and label to be red, no red for anyone 

(you can check the lib here http://angular-patternfly.rhcloud.com/#/api/patternfly.form.directive:pfFormGroup)

serenamarie125:
ok thanks for trying @AllenBW

serenamarie125:
@AllenBW per the asterisk, might you be able to send a code snippet to the PatternFly team so that they can see the issue ? I mentioned the issue to them this morning and they ask for some sample code or a https://plnkr.co/ link?

AllenBW:
@serenamarie125 no code snippet I can send  as per my comment above:

> you can check the lib here http://angular-patternfly.rhcloud.com/#/api/patternfly.form.directive:pfFormGroup one possible change is to render html in label, not just a string

This is one approach, using https://docs.angularjs.org/api/ng/directive/ngBindHtml#!, instead of a current state,

PR #403, opened 2016-12-20
Add Ability to edit a request

chalettu:
The PV ticket for this is https://www.pivotaltracker.com/story/show/134071981


chalettu:
@ui-bot add_label wip

client/app/states/requests/details/details.state.js
```diff
@@ -32,10 +32,60 @@
...
+    });
+
+    function saveRequest() {
+      var dialogData = dataForSubmit();
+      var data = { action: 'edit', dialogData };
+      CollectionsApi.post(
+        'service_requests',
+        request.id,
+        {},
+        { action: 'edit', options: { dialog: dialogData } }
```
AparnaKarve: @chalettu @chriskacerguis  I'm wondering if the `service_requests` are supposed to be editable or not...
The API does seem to have a resource action for `edit`, however in the OPS UI there is no option to edit or copy a Request of type `ServiceTemplateProvisionRequest`; only Requests of type `MiqProvisionRequest` can be edited or copied.

Are we now supporting edits for `service_requests`?

chriskacerguis: Yes, we are now doing that.

AparnaKarve: Ok, cool!

chalettu: BTW, this ticket is most definitely a work in progress and is not ready for review.  I am committing to it merely as a backup for my laptop. Thanks

chalettu:
new_buttons
@serenamarie125 @cshinn @beanh66 .  I need some UX guidance on this page.  I am working on the ability to edit a request that is "Pending approval".  If you look at the picture attached I created two buttons.  One that allows editing and one that Allows for saves to what is being edited.  I need some guidance on a couple of fronts.  I was wanting to see if we should have a "Cancel" button as well and I also need advice on what the actual interaction states for the Save button, should be.  Should the "Save" button be disabled until the "Edit" button has been pressed or should we hide the button altogether before the "Edit" button has been pressed.

beanh66:
@chalettu The edit button makes sense in the top right corner where you have it. It should only be visible for requests that are pending, correct? I was also wondering if there should be a delete button for pending requests. Once the user clicks "Edit," the Request should go into edit mode and the Save and Cancel buttons belong at the bottom left section of the page. This would be consistent with the way we are handling edit mode for Catalogs and Profiles. I attached a screenshot for clarity. 
edit_mode

AparnaKarve:
> I was also wondering if there should be a delete button for pending requests.

@beanh66 On that note, we would also need a "Copy" button down the line.
Can you suggest a layout to have all these buttons on this form without it looking too busy? Should we resort to the kebab menu layout like we have in the Service detail page?

chalettu:
On the topic of a "Delete" or "Copy" buttons being added, those are things that stretch outside the scope of this PR.  @chriskacerguis and @serenamarie125 I think we need to have a discussion about these other actions and come to an agreement so we can break them out into other PR's if we decide they are necessary.  Thoughts?

chalettu:
image_2
image_1
I attached the updated image of how the request detail page now looks when you go to try and edit a "Pending" request.

chalettu:
@miq-bot remove_label wip

chriskacerguis:
@chalettu is this ready to go?  you removed the WIP label, but not the prefix?

chalettu:
Should be ready to go.

PR #405, opened 2016-12-21
Refactor gulp build tasks

jjlangholtz:
Refactor gulp build tasks to improve parity between development and production builds. Also isolate a point in the build process for a Babel transpilation step.

This PR consolidates several image processing and tasks that differ between development and production environments. Instead of having a separate task for each environment that differ only in output directories, a `build-copy` task will handle copying all assets to the correct build directory.

The new `bundle` task concatenates the application JS into a single bundle file along with the corresponding sourcemaps. This is where the transpilation step will be added in a follow up PR.

https://www.pivotaltracker.com/story/show/136032077


chriskacerguis:
@AllenBW @chalettu thoughts?

gulp/config.js
```diff
@@ -328,7 +281,6 @@ module.exports = (function() {
     ngAnnotateOptions: {
       single_quotes: true
     },
-    devHost: 'http://localhost:3000'
```
jjlangholtz: Ah, removed because it doesn't appear to be used anywhere.

AllenBW: its not 🤘

chalettu: Cool. Looks good to me.

chalettu:
So this PR looks like a good cleanup effort.  It seems to clarify and consolidate things and I am all for it.  Good work!

gulp/config.js
```diff
@@ -194,48 +195,7 @@ module.exports = (function() {
...
-
-  config.devSkinImages = {
-    src: [
-      client + 'skin/images/**/*.*'
-    ],
-    build: temp + 'images',
-    minify: false
-  };
-
-  config.imgs = {
```
AllenBW: I don't know that we ever actually use any of these, but including the contents of `.tmp/img` does appear to be an omission from the refactored build config...

jjlangholtz:

PR #407, opened 2016-12-21
Add babel transpile step

jjlangholtz:
Add Babel and associated plugins to build processes. This allows to now use the latest Javascript syntax (https://babeljs.io/learn-es2015/). We should now begin taking advantage of language level benefits such as `const/let` instead of `var`, arrow functions, template strings, spread operator, and object/array destructuring.

Note as part of this change BrowserSync will now re-compile the application on each javascript file change. This will make refreshes **temporarily** slower, however, this will be drastically improved when we switch over to webpack-dev-server.

Many test files had to be slightly updated because as part of running these files through the transpile process they must be evaluated in strict mode. In strict mode these tests were failing because of their use of global/undeclared variables.

`app.module.js` and `tests/app.module.js` were updated to use ES6 syntax as guinea pigs to verify that transpilation works correctly.

ESLint rules will likely need to updated to account for ES6.

https://www.pivotaltracker.com/story/show/136032077


gulp/tasks/serve.js
```diff
@@ -99,8 +99,11 @@ module.exports = function(gulp, options) {
     }
 
     // Watches less, compiles it to css, browser-sync handles reload
-    gulp.watch([config.sass], ['sass'])
-      .on('change', changeEvent);
+    gulp.watch([config.sass], ['sass']).on('change', changeEvent);
+
+    // Watches source JS files and rebuilds bundle on changes
+    gulp.watch([config.js], ['compile']).on('change', changeEvent);
```
AllenBW: YES YES YES YES YES YES YES YES YES

client/app/app.module.js
```diff
@@ -1,39 +1,45 @@
...
-
-    vm.keyDown = function(evt) {
-      $scope.$broadcast('bodyKeyDown', {origEvent: evt});
-    };
-
-    vm.keyUp = function(evt) {
-      $scope.$broadcast('bodyKeyUp', {origEvent: evt});
-    };
-  }
+  ]).controller('AppController', AppController);
```
AllenBW: CLEAN <3

chriskacerguis:
This is a significant new paradigm.  Would love feedback from @AllenBW @himdel @AparnaKarve @h-kataria @chalettu @martinpovolny @jeff-phillips-18 @dtaylor113 @dgutride 

/cc @dclarizio @Fryguy

AparnaKarve:
> Note as part of this change BrowserSync will now re-compile the application on each javascript file change. This will make refreshes temporarily slower, however, this will be drastically improved when we switch over to webpack-dev-server.

@jjlangholtz When you say -- temporarily slower, any idea how long the refreshes would take? Are we talking about a few extra seconds?
And with webpack-dev-server would they be as fast as they are now?

This is great and would be a game changer ... also one more step towards mainstream Angular2 adoption. 👍

jjlangholtz:
@AparnaKarve From my own testing I'd estimate 1-2 seconds longer than what it used to be. With webpack-dev-server the refreshes will instantly become much faster because the module bundling will only need to re-compile the files that actually change, so the overhead would primarily be on the page refresh time.

Eventually we could work towards [HMR](https://webpack.github.io/docs/hot-module-replacement-with-webpack.html) which would allow us to make changes without needing to refresh at all 🤖

himdel:
Looks like a great first step :+1: :)

The only bit that surprised me was the order in `'client/app/**/*.js': ['coverage', 'babel'],`, @jjlangholtz you're sure coverage can run before babel?

gulp/tasks/serve.js
AllenBW: We gotta get this in, autocompile is kinda clutch

tests/dashboard.state.spec.js
```diff
@@ -34,6 +34,7 @@ describe('Dashboard', function() {
   describe('controller', function() {
     var controller;
```
AllenBW: why not change the rest of these `var` to `let`?

jjlangholtz: For this PR I tried limiting the test file changes to fix test failures due to 'strict' mode requirements. I was thinking we could change `var` to `let` holistically later on with a regex or code mod.

jjlangholtz:
@himdel I tried running the coverage reporter with `babel` running first and last and it seemed to work in both cases. It might be because the specs are being run with Node 6+ which already supports most ES6 features. Either way I switched the order so that `coverage` runs after `babel`, thanks!

himdel:
@jjlangholtz aah could be that, yeah, thanks to you too :)

chalettu:
Reviewed the changes and it looks like a solid step forward for us !  Great work.

chriskacerguis:
@jjlangholtz is this ready to go?

jjlangholtz:
Yes I believe I have addressed all the feedback 👍

jjlangholtz:
@chriskacerguis The workflow doesn't really change with this PR, just the ability to use ES2015/ES6 JavaScript. I don't think any of the documentation needs to be updated for now but certainly will with the next PR when webpack-dev-server is introduced.

PR #415, opened 2017-01-05
Initial set-up for webpack-dev-server

jjlangholtz:
This change is the beginning of the migration from a gulp-based build
system to a webpack-based one. Ideally, the existing gulp tasks should
continue to work until all desired functionality is replicated to the
webpack-based build for each environment (dev, test, prod). From this
point forward, devs should use `yarn start` instead of `gulp serve` when
starting the project locally.

`yarn start` will now kickoff `webpack-dev-server` which should **mostly**
work the same as the previous dev server. However, there is bound to be
some issues that will be discovered/fixed over the coming weeks.

@miq-bot add_label euwe/no, enhancement, wip

**TODO:**

- [x] Some styles are not getting picked up by webpack, figure out which and why
- [x] Some icons/fonts are missing as well, figure out which and why
- [ ] ~webpack-dev-server is currently serving from `/` instead of `/ui/service/`~ This can be addressed when configuring webpack for production usage as the base url doesn't affect development
- [x] Add loaders for .js files (babel-loader)
- [x] Add loaders for .html files (ngtemplate-loader, html-loader)
- [x] Clean up and document webpack configuration
- [x] Fix `sprintf is not a function` when opening modals
- [x] Configure scripts/styles loaders to output sourcemaps


chriskacerguis:
@jjlangholtz this looks really good.  super excited to get this integrated.  Could you please update the readme file with that yarn start vs gulp serve?

client/index.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html ng-app="app" class="layout-pf layout-pf-fixed layout-pf-fixed-inner-scroll transitions">
+  <head lang="en">
+    <title ng-bind="title" translate>Self Service</title>
```
chriskacerguis: This should be "Service UI"

jjlangholtz:
@miq-bot remove_label wip

chriskacerguis:
@AllenBW can you give this a once over?

chriskacerguis:
@jjlangholtz what needs to be done for production usage?  Do we need to do anything to the prod build scripts?

AllenBW:
Oh yeah! totally dropped in on this, it felt real nice. tests run, app builds (and reloads) for development nicely, faster than before (though there might be an issue with `Error: [ManageIQ] transition superseded` on login looking at @jjlangholtz for guidance)

We need to update base url for production (currently its set at `/` rather than `/ui/service/`, but with that in can use gulp build till the effort is put into something less clunky, but we need to decide when that'll be done because maintaining deps in three places is kinda insane (one for tests, one for build, one for webpack; eventually should just be webpack for all)

jjlangholtz:
@chriskacerguis We will need additional webpack configuration for production, there are many plugins designed for optimizing, compressing, minimizing etc. It shouldn't be too much additional effort but I believe it can be done in a follow-up PR.

We would also need a new `package.json` build script that we can kick off with something like `yarn run build` that will output the bundles to the manageiq public directory.

chriskacerguis:
@jjlangholtz can you please make 2 cards?  One for the prod changes, and one for the package.json thing you described?

package.json
```diff
@@ -90,11 +96,13 @@
     "gulp-uglify": "2.0.0",
     "gulp-useref": "3.1.2",
     "gulp-util": "3.0.8",
+    "html-loader": "^0.4.4",
```
AllenBW: Should we keep this absolute?

AllenBW:
👯‍♂️ 👯 💃 🕺

PR #422, opened 2017-01-10
Remove gulp (mostly)

jjlangholtz:
This branch is for tracking the removal of gulp since it is no longer necessary after switching to webpack. There are still several existing gulp tasks that need to be evaluated for removal and/conversion to a webpack equivalent.

https://www.pivotaltracker.com/story/show/137366919

@miq-bot add_label wip, euwe/no, refactoring


jjlangholtz:
Follow on story for moving translation tasks: https://www.pivotaltracker.com/story/show/137725283

@AllenBW @chalettu 
@miq-bot remove_label wip

PR #424, opened 2017-01-11
For angular patternfly downgrading lodash

AllenBW:
For whenever we can go back to 4.x, `sum` will have to be changed back to `sumBy`


AllenBW:

chriskacerguis:
Why are we downgrading?

jeff-phillips-18:
Angular Patternfly still requires lodash 3.10. Soon Angular Patternfly 4.x will release with upgraded dependencies including lodash 4.x.

PR #425, opened 2017-01-12
Various tweaks to allow js minification for prod build

jjlangholtz:
Enable the `ng-strict-di` directive on `html` so that exceptions are
thrown when dependency injection is not handled properly.

Swap the babel plugin ng annotation tool with a loader variant. The
babel-focused version was missing several annotations which prevented
the build bundle from working properly.

Lock ngProgress to the latest version (released on GH but not npm). The
old version we are currently does not handle dependency injection
correctly and therefore cannot be minified.

https://github.com/VictorBjelkholm/ngProgress/issues/116


jjlangholtz:
https://www.pivotaltracker.com/story/show/137196541

@AllenBW

chriskacerguis:
@jjlangholtz can you please fix the conflict?

AllenBW:
Thems some quick responses tex 🥇

chriskacerguis:
Indeed...but now tests are failing 🤕

PR #426, opened 2017-01-13
[EUWE] BZ: 1409959

AllenBW:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1409959

For some reason, some other pr that merged into euwe invalidated the filter `service_id=nil` (which was being used to grab parent services).  This pr replace all instances of the invalid filter that was preventing state resolves.  The filter `ancestry=null`  is presently being used on master, though there was no single pr that resulted in this change so we are making a new pr here.

`ancestry=null` achieves the same result intended by `service_id=nil`

@simaishi @chessbyte @chriskacerguis 

@miq-bot add_label euwe/yes, bug

### DIFF
--- a/client/app/config/navigation.config.js
+++ b/client/app/config/navigation.config.js
@@ -147,7 +147,7 @@
     function fetchServices() {
       var options = {
         expand: false,
-        filter: ['service_id=nil'],
+        filter: ['ancestry=null'],
         auto_refresh: true,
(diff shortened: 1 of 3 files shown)


simaishi:
Euwe BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1410516

PR #428, opened 2017-01-13
SUI: Canvas should be zoomable

dlabrecq:
Added functionality to make blueprint canvas zoomable.

https://patternfly.atlassian.net/browse/CFUX-278
https://www.pivotaltracker.com/n/projects/1914499/stories/135548487

Zoom Level: 100%
maxzoom

Zoom Level: 75%
midzoom

Zoom Level: 50%
minzoom

### DIFF
--- a/client/app/components/blueprints/blueprint-editor/blueprint-editor.directive.js
+++ b/client/app/components/blueprints/blueprint-editor/blueprint-editor.directive.js
@@ -214,20 +214,11 @@
     // Zoom
     //
     vm.maxZoom = function() {
-      if (vm.chartViewModel && vm.chartViewModel.zoom) {
-        return vm.chartViewModel.zoom.isMax();
-      } else {
-        return false;
-      }
-    };
-
+      return (vm.chartViewModel && vm.chartViewModel.zoom) ? vm.chartViewModel.zoom.isMax() : false;
+    }
     vm.minZoom = function() {
-      if (vm.chartViewModel && vm.chartViewModel.zoom) {
-        return vm.chartViewModel.zoom.isMin();
-      } else {
-        return false;
-      }
-    };
+      return (vm.chartViewModel && vm.chartViewModel.zoom) ? vm.chartViewModel.zoom.isMin() : false;
+    }
 
     vm.zoomIn = function() {
       $scope.$broadcast('zoomIn');
--- a/client/app/components/blueprints/blueprint-editor/blueprint-editor.directive.js
+++ b/client/app/components/blueprints/blueprint-editor/blueprint-editor.directive.js
@@ -209,5 +209,32 @@
         })[0];
       }
     };
+
+    //
+    // Zoom
+    //
+    vm.maxZoom = function() {
+      if (vm.chartViewModel && vm.chartViewModel.zoom) {
+        return vm.chartViewModel.zoom.isMax();
+      } else {
+        return false;
+      }
+    };
+
+    vm.minZoom = function() {
+      if (vm.chartViewModel && vm.chartViewModel.zoom) {
+        return vm.chartViewModel.zoom.isMin();
+      } else {
+        return false;
+      }
+    };
+
+    vm.zoomIn = function() {
+      $scope.$broadcast('zoomIn');
+    };
+
+    vm.zoomOut = function() {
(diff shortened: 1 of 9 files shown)


dtaylor113:
Hi @dlabrecq , nice job!
Just wondering if the + - icons display when viewing a published blueprint?   Published blueprints are read-only so the duplicate and delete buttons are hidden (which might affect the alignment of the + - icons/buttons).  -thanks

dlabrecq:
Yes, the icons should be available for a published blueprint. I saw the read-only layout in the HTML and tested it with and without the duplicate and delete buttons. The + and - icons still align right as expected.

dtaylor113:
Awesome @dlabrecq , trying to test this branch, but can't log into SUI :frowning:  I did:
```
yarn cache clean
rm -rf node_modules
yarn run build
yarn start
```
Everything builds, runs fine.  But can't login.  Did you use these build steps?

dtaylor113:
Got it running, some webpack dev settings.

Noticed that if a zoom out to max., then drop an item on the canvas all way over on right, then zoom back in, the item is no longer shown (as expected), however there is no vertical scroll bar to be able to scroll/pan over to see the offscreen item.

dtaylor113:
Another issue -which I think was somewhat pre-existing, is that when I resize the browser, I don't see horiz. or vert. scrollbars.   I thought I had this working in the past, however wasn't ideal because as one resized browser the vert. scroll bar got smaller and smaller  -but it was there :smile:

dtaylor113:
Code looks great!  :+1:

dtaylor113:
Hmm...I somewhat take back my first comment about scrollbars when zooming in & out.  Turns out I had dragged the browser tab to a new screen, which has a lower resolution, so the bottom (horiz) scrollbar wasn't onscreen.   If I turn my display to 1280x720, screen only shows a small portion of SUI and both horiz and vert. scrollbars are not shown/offscreen.  
Think we need to determine what resolutions we support.

dlabrecq:
Regarding the max zoom-out... Yes, that is expected. The canvas is the same size, but the underlying SVG is a bit larger to accommodate zooming out.

When you start out at 100%, you may notice that you can scroll. Depending on the OS, the scroll bars may not appear until you actually begin scrolling (e.g., with the Mac mouse).

As you mentioned, if you place a node to the far right (when zoomed out by 50%), then zoom in, you'll need to scroll to see it at the 100% zoom level. 

When you're max zoomed out by 50%, the canvas and SVG are the same size, so there is no scrolling at this zoom level.

That said, I reviewed this with Chris and he was ok with it. However, we can tweak this later if necessary. An alternative is to let the canvas shrink when zooming out. However, that doesn't provide any more room to work with.

dtaylor113:
LGTM :-)

client/app/components/blueprints/blueprint-editor/blueprint-editor.directive.js
```diff
@@ -209,5 +209,32 @@
...
+    //
+    vm.maxZoom = function() {
+      if (vm.chartViewModel && vm.chartViewModel.zoom) {
+        return vm.chartViewModel.zoom.isMax();
+      } else {
+        return false;
+      }
+    };
+
+    vm.minZoom = function() {
```
AllenBW: comment at line 217 also applies

client/app/components/blueprints/blueprint-editor/blueprint-editor.directive.js
```diff
@@ -209,5 +209,32 @@
         })[0];
       }
     };
+
+    //
+    // Zoom
+    //
+    vm.maxZoom = function() {
+      if (vm.chartViewModel && vm.chartViewModel.zoom) {
```
AllenBW: Recommend simplifying to: 
```
vm.maxZoom = vm.chartViewModel && vm.chartViewModel.zoom ? vm.chartViewModel.zoom.isMax() : false;
```

dlabrecq:
Simplified vm.maxZoom and vm.minZoom, but kept them as functions because the buttons are updated dynamically.

chriskacerguis:
I checked Code Climate, and it passed, however the status in Github is still "pending" as of this comment.  Going to merge.

PR #430, opened 2017-01-13
Updates to Service Explorer, show child services

jeff-phillips-18:
This PR changes the service explore fields being displayed and adds showing the child services for services that have them.

PT: https://www.pivotaltracker.com/story/show/137174911
Jira: https://patternfly.atlassian.net/browse/CFUX-294

Finishes 137174911

image

image

image

@beanh66 @serenamarie125 @chriskacerguis 

@miq-bot add_label euwe/no
@miq-bot add_label ux


beanh66:
Thanks @jeff-phillips-18 the changes look great!

serenamarie125:
@jeff-phillips-18 some of those screens still show the icon near Retires ... but other show Never.  Can you confirm that the icon was replaced with Never ?

jeff-phillips-18:
@serenamarie125 sorry, those were older screenshots. I've updated them.

client/app/components/service-explorer/service-explorer.html
```diff
@@ -0,0 +1,250 @@
...
+           tooltip-placement="bottom"></i>
+        <i class="pficon fa fa-question-circle" ng-if="item.power_state !== 'on' && item.power_state !== 'off'"
+           uib-tooltip="{{'Power State: Unknown'|translate}}"
+           tooltip-append-to-body="true"
+           tooltip-popup-delay="1000"
+           tooltip-placement="bottom"></i>
+        <a uib-tooltip="{{item.name}}"
+           tooltip-placement="bottom"
+           tooltip-append-to-body="true"
+           tooltip-append-to-body="true"
```
AllenBW: Recommend removing duplicate attribute.

client/app/components/service-explorer/service-explorer.html
```diff
@@ -0,0 +1,250 @@
...
+        <div class="list-view-stacked-item"
+             uib-tooltip="{{item.retires_on ? (item.retires_on | date : 'medium') : ('Never' | translate)}}"
+             tooltip-append-to-body="true"
+             tooltip-popup-delay="1000"
+             tooltip-placement="bottom">
+          <strong translate>Retires</strong>&nbsp;
+          <div ng-if=(item.retires_on)>
+            {{ item.retires_on | date : 'shortDate' }}
+          </div>
+          <div translate>Never</div>
```
AllenBW: Never will always be shown, might want to throw an ng-if on it as well.

client/app/components/service-explorer/service-explorer.html
```diff
@@ -0,0 +1,250 @@
...
+              <strong>Created</strong>&nbsp;
+              <div>
+                {{ item.created_at | date : 'shortDate' }}
+              </div>
+            </div>
+          </div>
+          <div class="col-lg-2 col-md-2 col-sm-2 hidden-xs">
+            <div class="list-view-stacked-item"
+                 uib-tooltip="{{item.retires_on ? (item.retires_on | date : 'medium') : ('Never' | translate)}}"
+                 tooltip-append-to-body="true"
```
AllenBW: These three options:
```
tooltip-append-to-body="true"
tooltip-popup-delay="1000"
tooltip-placement="bottom"
```
are repeated many times, recommend  using the $uibTooltipProvider to reconfigure the default options in each controller
http://angular-ui.github.io/bootstrap/#/modal

jeff-phillips-18: @AllenBW  Agreed, but I was reluctant as it would effect all screens. Fast tooltips on list views become annoying as you mouse over rows and the tooltips fly by. Is this something we want throughout the application? Only alternative would be to (as you suggest) configure it in each and every controller.

AllenBW: @chriskacerguis @serenamarie125  looking for yo 💭 

@jeff-phillips-18 I kinda wanta gingerly say throughout application, one view shouldn't yield a drastically different tooltip experience than the other, so if they are too fast as is, yeah lets slow em all down!  and given these options appear to be applicable more often than not, we can put the burden of overriding them on each controller rather than risk the inconsistency of attempting to apply them that way...

ALL of this might be beyond the scope of this pr though (your discretion of course). If so, saul good here 🌮

jeff-phillips-18: the only parameter that ought to go global would be the popup delay. The tooltip placement is component dependent as is the append-to-body.

I'm thinking the delay setting application wide would be for another PR as individual settings  should then be removed from several places.

jeff-phillips-18:
Updated per comments.

And yes, this is to follow the latest design from https://github.com/ManageIQ/manageiq-design/blob/master/UX/ui-service/MyServices/design.md.

I can add the link to Jira and PT.

client/app/components/service-explorer/service-explorer.html
AllenBW: Ah append-to-body yeah, good call on that, tooltip placement though is set to `top` by default, more places than not we use it on the bottom though, I would think it to would make sense to make the default whatever we use more often throughout the app and override when necessary. 

Will make a pt encompassing this work and start on it today (if you're on board and nothing else flares up)

chriskacerguis:
@AllenBW can you check the update?

AllenBW:
@chriskacerguis THIS THING IS 🌮

chriskacerguis:
@AllenBW can you change your approval status?

PR #432, opened 2017-01-16
Minify css bundle output from `yarn build`

jjlangholtz:
Add an additional webpack plugin to minify the output css bundle when
building for production.

In addition adjust source map settings for versions that are faster and
smaller. Tested with latest version of Chrome and the source maps worked
pretty well, with other browsers/versions your mileage may vary.

https://www.pivotaltracker.com/story/show/137196541


jjlangholtz:

PR #442, opened 2017-01-18
Add several missing ngInject annotations for blueprints

jjlangholtz:
When explicit ngInject annotations are missing, dependency injection
does not work properly when the JS is minified. These annotations were
missing leading to the behavior observed in this PT ticket.

https://www.pivotaltracker.com/story/show/137820477


jjlangholtz:

dtaylor113:
Hi, sorry, I wasn't aware that we needed those ngInject comments.  Thanks!

PR #444, opened 2017-01-18
Bug Fix for Service Reconfigure page not pre-populated

chalettu:
This is the fix for a bug reported that the service reconfigure page does not populate dialog or other fields on the reconfigure page.  Pivotal Ticket https://www.pivotaltracker.com/story/show/136522459


chalettu:
@miq-bot add_label bug

PR #448, opened 2017-01-19
Remove toastr as a global and import missing toastr css

jjlangholtz:
Toastr messages do not display properly because the css associated with
the package is not being imported. This change removes Toastr as a
global (it is only used in the logger module) and imports the missing
css.

Before:
screen shot 2017-01-19 at 10 01 17 am

After:
screen shot 2017-01-19 at 10 00 30 am

Also modify the eslintrc.json file to support ES6 module import/export
syntax.

Finally remove the Toasts factory as it is not used anywhere.


jjlangholtz:
https://www.pivotaltracker.com/story/show/137965363

@miq_bot add_label euwe/no, bug

AllenBW:
@jjlangholtz Why not remove toastr all together and use `EventNotifications`?  Do we really need to have two notification libs?

jjlangholtz:
@AllenBW Totally on board with that option as well, actually we have three notification libs: `toastr`, `EventNotifications` (SUI exclusive), and `Notifications` from angular patternfly. I'll put some feelers out to what the level of effort is to consolidate on one option.

AllenBW:
`EventNotifications` ties in with the nav bar and I believe the one we would want to keep
Roping in @jeff-phillips-18 as he is da 🏄 on this

jrafanie:

PR #450, opened 2017-01-19
Add explicit import and export, convert app scripts to modules

jjlangholtz:
Refactor all application scripts to use import and export instead of
IIFEs. Everything should work exactly the same as before, however, this
refactoring will facilitate many future improvements.


chriskacerguis:
Great googly-moogly.  That's a massive PR (yes, I understand why).  😄

chriskacerguis:
@jjlangholtz PT link?

client/app/blocks/exception/exception-handler.provider.js
```diff
@@ -1,69 +1,55 @@
...
+}
+
+/**
+  * Extend the $exceptionHandler service to also display a toast.
+  * @param  {Object} $delegate
+  * @param  {Object} exceptionHandler
+  * @param  {Object} logger
+  * @return {Function} the decorated $exceptionHandler service
+  * @ngInject
+  */
```
chriskacerguis: Great commenting.  That is some good 🌮

client/app/core/constants.js
```diff
@@ -1,10 +0,0 @@
-/* global _:false, ActionCable:false, $:false, sprintf: false, moment: false */
```
AllenBW: Why kill the standalone constants file?

jjlangholtz:
Some more context on why these changes are necessary.

## Background
In order to take advantage of the amazing features a module bundler (`webpack`) provides, we must think about our application as a collection of modules that import what they need and export what they expose. Contrast that with the way the application is composed prior to this change: a collection of scripts that operate on some global data, namely the `angular` object.

When initially setting up webpack, much care had to be taken to load each script in the same order that it was being loaded before. The application has a strict lexical ordering requirement which is not conducive to the way modules work. For example, the current working bundle is assembled with this mechanism:

```javascript
requireAll(require.context('./app', true, /\.module\.js$/));
```

Which webpack expands to something like this, requiring each file in turn:

```javascript
var map = {
	"./blocks/exception/exception-handler.provider.js": 228,
	"./blocks/exception/exception.js": 229,
	"./blocks/logger/logger.js": 231,
	"./blocks/router/router-helper.provider.js": 233,
	"./components/autofocus/autofocus.directive.js": 235,
	"./components/blueprints/blueprint-canvas/blueprint-canvas.directive.js": 236,
	"./components/blueprints/blueprint-delete-modal/blueprint-delete-modal-service.factory.js": 237,
	"./components/blueprints/blueprint-details-modal/blueprint-details-modal-service.factory.js": 238,
	"./components/blueprints/blueprint-details-modal/blueprint-order-list.service.js": 239,
// ...a hundred more entries
```

## Module Advantages

When the scripts are loaded in such a fashion webpack has no way of building up a *dependency graph*. This graph is necessary to use webpack how it is intended to be used. It opens the doors for a number of things:

* *Code spliltting* - The ability to tactically choose which portions of the application are requested by the client and when. Reducing initial bundle sizes and forcing modularity.
* *Hot Module Replacement* - The ability to instantly update a running module on the fly without a page refresh.
* *Future Angular Compatibility* - Future versions of Angular use webpack and module syntax.
* *Explicitness* - It is very clear where identifiers are coming into a module and what identifiers can be used in other modules.

## Why this PR

This is an iterative step to break up the lexical ordering requirements of the application. It makes it very clear that we are organizing the application by types of angular entities (components, services, etc) as opposed to grouping them by how feature/tabs (services, requests, catalogs, designer). As it stands we cannot split/chunk when bringing in the components module brings in every component used across the app:
https://github.com/ManageIQ/manageiq-ui-service/pull/450/files#diff-3069fc5513b72684a639340cd92d00a5

Now that it is explicit what each module contains, it is now possible to break apart the large modules and create chunks based on how the application is used.

Imagine an initial chunk that contains the only the essentials necessary to render the initial dashboard view. Then as you navigate the app, the necessary chunks are loaded on the fly. When you navigate to the designer tab, you load the designer code. Furthermore, when it comes time to upgrade Angular, the upgrade can happen chunk by chunk, view by view, instead of rewriting the whole thing.

Just sharing a few thoughts, let me know what you think!

@ManageIQ/core-ui-service

client/app/core/constants.js
jjlangholtz: Since there wasn't really anything to export from this file (everything in it is global already), I just moved the constant registrations to the `app.core` module itself. That way it is very easy to see exactly what belongs to each module.

chriskacerguis:
@AllenBW any new thoughts on this?

AllenBW:
@chriskacerguis Itza gotta be done eventually, no time like the present!  

@jjlangholtz your thoroughness of explanation is mucho 💃

jjlangholtz:
https://www.pivotaltracker.com/story/show/138043239

PR #452, opened 2017-01-23
[Euwe] Add clear button to retire service modals

jjlangholtz:
Adds a clear button to the bootstrap datepicker for the retire service
modals. This allows a user to remove a previously set retirement date.

Note that this will only work if you actually click the clear button, if
you simply delete the previous input and then click save, the previous
selected value from the datepicker will be submitted.

The modifications to `client/index.html` come from running
`npm install`. It appears that somewhere along the line those changes
weren't committed. It shouldn't have any side effects seeing as they are
automatically injected when running `npm install` anyways.

### DIFF
--- a/client/app/components/retire-service-modal/retire-service-modal-service.factory.js
+++ b/client/app/components/retire-service-modal/retire-service-modal-service.factory.js
@@ -50,6 +50,7 @@
 
     vm.dateOptions = {
       autoclose: true,
+      clearBtn: true,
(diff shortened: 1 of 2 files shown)


chriskacerguis:
@simaishi approved.  👍

simaishi:
Euwe BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1415754

PR #453, opened 2017-01-23
Bug fix for creating rules

jeff-phillips-18:
Need to trim selections from pf-select component.

@miq_bot add_label euwe/no, bug


jeff-phillips-18:
Sorry about the invalid @miq_bot, I copied and pasted from another PR without checking :(

PR #454, opened 2017-01-23
Add improvements to exception handling messages

jjlangholtz:
Minor improvements to the exception service:
* split exception.catcher into a function per use case
* add documentation for recommended usage
* rename to `exception.service.js` to follow naming conventions

Add a guard clause to the `$exceptionHandler` decorator to ensure an
error object.

Add a missing `.catch` clause to the login function.

Note that you will still error messages when closing modals, but the output is less noisy. This is due to `angular-ui-bootstrap` not catching up to angular 1.6.1 yet. Looks like it is actively being fixed from https://github.com/angular-ui/bootstrap/issues/6360


jjlangholtz:
https://www.pivotaltracker.com/story/show/138055019
https://www.pivotaltracker.com/story/show/138055237

PR #458, opened 2017-01-23
Explicitly call yarn test as part of travis script

jjlangholtz:
Looks like travis stopped its default behavior of running `npm test`
once the custom script commands were added. Add `yarn test` explicitly.


jjlangholtz:
Same behavior merged in https://github.com/ManageIQ/manageiq-ui-service/pull/459

PR #459, opened 2017-01-24
Disable known failing tests

AllenBW:
Add yarn test explicitly to travis
Remove precision that occasionally fails service details test


simaishi:
@chriskacerguis There is no `yarn` in Euwe branch, so I don't think this PR should be backported as is. If you need a part of this PR in Euwe branch, please create a separate Euwe specific PR.

simaishi:
@AllenBW @chriskacerguis ping ^

chriskacerguis:
@simaishi thanks for pointing that out :)

AllenBW:
hokay, this will be fixxxxxxxxedddddddd

AllenBW:
@simaishi @chriskacerguis I don't believe this to be applicable to euwe (none of the files are present), is there any reason why we're trying to backport this to euwe?

chriskacerguis:
@AllenBW I removed the euwe/yes tag, so there is no need to back port.

PR #473, opened 2017-01-27
Update webpack config to be aware of skinning

jjlangholtz:
Allow images from `client/app/skin/images` to override default images. Include all *.js and *.css files from `client/app/skin/**` as the last step of the bundle.

I believe this will fix the skinning issues provided productization follows the steps outlined in the skinning README. To test locally:

1. `skin-sample/link.sh` from the root dir of the SUI.
2. `yarn start` (restart if already running)
3. Verify that images, styles, and selected text have been skinned.
4. `skin-sample/unlink.sh` to restore.

Still trying to figure out how to prevent the warning emitted when running without a skin.

@AllenBW @himdel 

### DIFF
--- a/client/app.js
+++ b/client/app.js
@@ -58,3 +58,10 @@ require('./assets/sass/styles.sass');
 
 // Angular templates
 requireAll(require.context('./app', true, /\.html$/));
+
+// Skin overrides
+try {
+  requireAll(require.context('./skin', true, /\.(js|css)$/));
+} catch (e) {
+  // Skin dependencies are not linked
+}
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -1,3 +1,4 @@
+const fs = require('fs');
 const path = require('path');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 const ExtractTextWebpackPlugin = require('extract-text-webpack-plugin');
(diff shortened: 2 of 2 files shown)


jjlangholtz:
https://www.pivotaltracker.com/story/show/138512313

chalettu:
Hey @jjlangholtz do you think we could create another yarn task to run in "skinned" mode?  like yarn start:skinned ?   This is just thinking out loud and opening up for discussion.

jjlangholtz:
@chalettu I like the idea, would need a `yarn build:skinned` as well. One of the goals for this particular change was to not make any changes that required changes downstream. I'm actually not sure which commands are run downstream to link the skin dependencies.

chalettu:
@chriskacerguis do you think maybe we should have a follow on ticket to see about simplifying the skinning process to a new set of yarn commands??

chriskacerguis:
@chalettu I think it's a good idea, however the process is supposed to be transparent downstream, so I think just leave it as-is.

@jjlangholtz Looking at the build log, what are your thoughts about adding the ./skin as a dir in the repo (add a .keep file) so it avoids the error about not finding the dir?

jjlangholtz:
@chriskacerguis That is one of the methods I tried. The skinning is done with bash scripts that create a symlink at `client/skin`, so if a directory already exists at that target then the link is created within it, which is not what we want.

I think if we modified the scripts to not use symlinks and to copy/remove files instead it would work. I'm not sure the exact mechanism that is used downstream though so I was trying to avoid modifying anything in the `skin-sample` dir.

chriskacerguis:
@jjlangholtz thanks for the explanation.

PR #478, opened 2017-01-30
Added dependencies for the Dialog Editor components

romanblanco:
@karelhala


client/app.js
```diff
@@ -10,6 +10,7 @@ window.$ = window.jQuery = require('jquery');
 window._ = require('lodash');
 window.moment = require('moment');
 window.sprintf = require('sprintf-js').sprintf;
+require('jquery-ui-bundle');
```
AllenBW: should go under line 15

package.json
```diff
@@ -111,9 +111,13 @@
     "yargs": "6.6.0"
   },
   "dependencies": {
+    "angular-bootstrap-switch": "^0.5.1",
+    "angular-dragdrop": "^1.0.13",
+    "angular-ui-sortable": "^0.16.1",
```
AllenBW: Are these being instantiated in the app at all? 
(might want to checkout components.module.js)

package.json
```diff
@@ -111,9 +111,13 @@
...
     "express": "4.14.0",
     "http-proxy": "1.16.2",
-    "manageiq-ui-components": "0.0.11",
+    "jquery-ui-bundle": "^1.11.4",
+    "manageiq-ui-components": "romanblanco/ui-components#dialog_editor",
```
AllenBW: wondering what this does for anything else that has a ui-components dep...

jjlangholtz: Do these dependencies need to be added here? It would be great if we can keep them encapsulated in UI Components.

karelhala: The idea behind `ui-components` is that any MiQ application can use some kind of this library so some shared component (dialog editor, table, tile...) will behave same across all applications. However you don't need to use every component which is in `ui-components`, you might use 5 in SUI 7 in `ui-classic` etc. For this purpose we did not added any 3rd party library to `ui-components`. So if you want to use dialog editor from `ui-components` you have to install 3rd party libraries which are required by components corresponded for dialog editor (`jquery-ui`, `bootstrap-switch`, `angular-bootstrap-switch`, `angular-ui-sortable`, `angular-dragdrop`), but if you are not using these components you don't have to install these libraries.

karelhala:
@romanblanco please add `bootstrap-switch` to `package.json`
```shell
yarn add bootstrap-switch
```

And add it to `app.js` as well, so changing title of dialog will not throw error with missing dependency.
```Javascript
require('bootstrap-switch');
```

package.json
jjlangholtz: Thanks for the explanation @karelhala! I'm wondering if there is a possibility for `ui-components` to export individual components in addition to the entire library. That way dependencies can be contained within the library and consumers can import only those that they need. I have seen examples of this in libraries like `lodash` and `rxjs` and in some react component libraries.

romanblanco:
@karelhala, @AllenBW thanks, corrected

package.json
chriskacerguis: I'm thinking that @jjlangholtz thought is a good one.  Can we move to export those to individual components?

himdel:
>  Can we move to export those to individual components?

So.. the point of the ui-components repo is to have all those common components, we are definitely not splitting the repo.

We *could* theoretically produce multiple npm packages from that repo though, in addition to the bundle (what is now `manageiq-ui-components`).

So we'd have `manageiq-ui-components`, `manageiq-ui-components-gtl`, `manageiq-ui-components-toolbar`, `manageiq-ui-components-siteswitcher`, `manageiq-ui-components-dialogeditor` and `manageiq-ui-components-common`, upon which all the other 4 depend.

Then, those 5 new packages could either be compiled to es5, the same as before, or maybe left in typescript so that each dependent can compile it themselves.

@karelhala what do you think, any technical challenges there? Does it make sense to open the discussion in such a direction?


**But** on the other hand, that seems like a lot of work just so that SUI can depend on 3 of the 5 packages anyway. @chriskacerguis I fail to see the benefit now, is there any?

himdel:
@karelhala, @romanblanco  this is still broken unfortunately :(.

I have no local `ui-components`.
I have no `manageiq-ui-service/node_modules`.
I check out this branch, do `yarn; yarn start`. (Even deleting `yarn.lock` before then doesn't help.)
I go to Dialogs, click a random one, click Edit.

`angular.js:10837 Error: [ManageIQ] [$injector:unpr] Unknown provider: DialogEditorProvider <- DialogEditor`

himdel:
```diff
--- a/client/app/states/states.module.js
+++ b/client/app/states/states.module.js
@@ -36,6 +36,7 @@ export const AppRoutingModule = angular
   .module('app.states', [
     'app.core',
     'app.components',
+    'miqStaticAssets.dialogEditor',
   ])
   .run(NotFoundState)
   .run(AboutMeState)
```

This change makes the page not load, and complain that 

```
Uncaught Error: [$injector:modulerr] Failed to instantiate module app due to:
Error: [$injector:modulerr] Failed to instantiate module app.states due to:
Error: [$injector:modulerr] Failed to instantiate module miqStaticAssets.dialogEditor due to:
Error: [$injector:nomod] Module 'miqStaticAssets.dialogEditor' is not available! 
```

So.. either `ui-components` is still not incuded, or the component is not there.
(I did check that `yarn.lock` points to the right place.)

himdel:
Verified that the above issue is not in this PR, it's in how `yarn` (0.16.1) doesn't understand branches :(.

Installing via `npm` instead works just fine :). (And curiously enough reveals that SUI doesn't actually depend on `patterfly` at all, only `patterfly-sass`, which it doesnt use at all :).)

Which means I can merge ManageIQ/ui-components#55, release a proper version and this problem should go away.

AllenBW:
@himdel so latest yarn is v0.19.1, does using this version allow for correct instillation of the dep?

himdel:
@AllenBW tested again, yup, 0.19.1 works :).

But I've also released `manageiq-ui-components` `0.0.12` so we can depend on that now.

package.json
```diff
@@ -112,9 +112,14 @@
...
   "dependencies": {
+    "angular-bootstrap-switch": "^0.5.1",
+    "angular-dragdrop": "^1.0.13",
+    "angular-ui-sortable": "^0.16.1",
+    "bootstrap-switch": "^3.3.2",
     "express": "4.14.0",
     "http-proxy": "1.16.2",
-    "manageiq-ui-components": "0.0.11",
+    "jquery-ui-bundle": "^1.11.4",
+    "manageiq-ui-components": "romanblanco/ui-components#dialog_editor",
```
himdel: Should be 0.0.12 now :)

package.json
```diff
@@ -112,9 +112,14 @@
     "yargs": "6.6.0"
   },
   "dependencies": {
+    "angular-bootstrap-switch": "^0.5.1",
```
AllenBW: ok last feedback, if ya can, keep these versions absolute (remove the `^`)

romanblanco: thanks! 👍

karelhala:
@himdel @chriskacerguis @jjlangholtz Good idea about splitting components to some smaller ones. But I am strongly against splitting it into multiple repos/packages. Let's rework build process of these components in a way that the final package will have `manageiq-ui-components-gtl.js`, manageiq-ui-components-dialog-editor.js`, `manageiq-ui-components.js`... so you will install it from one package, but you can include only some JS files or main `manageiq-ui-components.js`.

Not sure if this is right place to discuss this, since it's closed PR, so probably move this to gitter?

chriskacerguis:
@karelhala Agree.  Splitting components = good, splitting repos = no bueno

PR #483, opened 2017-02-01
Vertical navigation support dynamic translations and badge counts

chalettu:
This enables the page to be refreshed without doing a window refresh.  
PV https://www.pivotaltracker.com/story/show/134436111

### DIFF
--- a/client/app/core/navigation.config.js
+++ b/client/app/core/navigation.config.js
@@ -69,7 +69,7 @@ export function navConfig(NavigationProvider) {
         {
           count: 0,
           tooltip: badgeTooltip,
-          originalToolTip: badgeTooltip
+          originalToolTip: badgeTooltip,
         },
       ];
     }
--- a/client/app/core/navigation.config.js
+++ b/client/app/core/navigation.config.js
@@ -3,42 +3,43 @@
 /** @ngInject */
 export function navConfig(NavigationProvider) {
   const dashboard = createItem(
-    N_('Dashboard'),
+    'Dashboard',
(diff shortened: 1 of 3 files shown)


chalettu:
@miq-bot add_label bug

chalettu:
Ultimately the reason we had issues translating is that we changed the source title's language in our navigation provider, since the provider stayed in its current state after a state reload, the system attempted to translate the translated title.  Solution was to create a variable of the original untranslated string that we can then use in our navigation controller to translate.

chalettu:
FYI - Only our Dashboard title changes at the moment.  We need these other terms to be translated in our menu.  @chriskacerguis where do I file a ticket requesting translations?

chriskacerguis:
@chalettu in BZ, set the "Component" to "Internationalization" (sp?).

client/app/core/navigation.config.js
```diff
@@ -3,42 +3,43 @@
...
 export function navConfig(NavigationProvider) {
   const dashboard = createItem(
-    N_('Dashboard'),
+    'Dashboard',
     'dashboard',
     'fa fa-dashboard'
   );
+
   const services = createItem(
-    N_('Services'),
```
chriskacerguis: Do we need the N_ for the translations (or did we change something I missed)?

chalettu: Translation happens in the navigation controller.  It was already doing it there as well.  It is better in this instance to do it in the navigation controller.

chalettu:
dmo

PR #484, opened 2017-02-02
Hide service dialog fields that are set as not visible

romanblanco:
https://bugzilla.redhat.com/show_bug.cgi?id=1417267

Setting the dialog field as not visible in the classic UI:
screenshot from 2017-02-02 11-33-46

Before:
screenshot from 2017-02-02 11-34-30

After:
screenshot from 2017-02-02 11-34-58

### DIFF
--- a/client/app/components/dialog-content/dialog-content.html
+++ b/client/app/components/dialog-content/dialog-content.html
@@ -20,7 +20,7 @@
                 ng-repeat="dialogField in ::dialogGroup.dialog_fields "
                 ng-init="inputTitle = dialogField.description"
                 class="form-horizontal">
-                <div class="form-group">
+                <div ng-show="dialogField.visible" class="form-group">
                   <label class="control-label col-sm-2">{{ ::dialogField.label }}</label>
 
                   <div ng-switch on="dialogField.type" class="col-sm-5">
@@ -144,7 +144,7 @@
                 </div>
               </form>
               <!-- Form for read only -->
-           
+
             </div>
           </div>
         </div>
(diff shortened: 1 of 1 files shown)


romanblanco:
@AllenBW, can you check it, please?

client/app/components/dialog-content/dialog-content.html
```diff
@@ -20,7 +20,7 @@ <h2 class="text-capitalize no-wrap">{{ ::vm.dialog.label }}</h2>
                 ng-repeat="dialogField in ::dialogGroup.dialog_fields "
                 ng-init="inputTitle = dialogField.description"
                 class="form-horizontal">
-                <div class="form-group">
+                <div ng-show="dialogField.visible" class="form-group">
```
AllenBW: prefer ng-if to ng-show (if removes hidden elements from dom, show puts an equivalent of a display none, can sometimes interfere with elements on the page)

client/app/components/dialog-content/dialog-content.html
```diff
@@ -144,7 +144,7 @@ <h2 class="text-capitalize no-wrap">{{ ::vm.dialog.label }}</h2>
                 </div>
               </form>
               <!-- Form for read only -->
-           
+
```
AllenBW: kick the whitespace! <3

client/app/components/dialog-content/dialog-content.html
romanblanco: @AllenBW I considered the option, but thought it was a purpose, having an invisible field in the dialog, but still sending its value. I'm not 100% sure about it though... I'm OK with changing it, if it's correct.

romanblanco:
@chriskacerguis done

simaishi:
@romanblanco @chriskacerguis can this be `euwe/yes` (as per flag on the BZ)?

chriskacerguis:
@simaishi I'm ok with that as long as @romanblanco is

simaishi:
Removing `euwe/yes` as this is already fixed in Euwe branch (via https://github.com/ManageIQ/manageiq-ui-service/pull/587)

PR #486, opened 2017-02-02
PT#138821445 - Webpack-build passes with symlinked folders

AllenBW:
https://webpack.js.org/configuration/resolve/#resolve-symlinks
So the way I understand this, we *were* resolving the symlinked folder at its location, doesn't have all da config goodness we need for things to not 💥, setting this to false, we resolve the symlinked folder where the 🐑 go to 🛌 (ie root of app).

Fixes skinning, breaks nothing, all 💃 🌮 here!

@jjlangholtz @chriskacerguis @simaishi 

@miq-bot add_label bug, euwe/no

https://www.pivotaltracker.com/story/show/138821445

### DIFF
--- a/config/webpack.dev.js
+++ b/config/webpack.dev.js
@@ -142,6 +142,10 @@ module.exports = {
     }),
   ],
 
+  resolve: {
+    symlinks: false,
+  },
+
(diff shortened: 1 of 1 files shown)


chalettu:
Looks like a good fix

PR #488, opened 2017-02-03
Removed pending approval filter default on Request explorer

chalettu:
PV story https://www.pivotaltracker.com/story/show/138363059


chalettu:
@miq-bot add_label bug
@miq-bot add_label euwe/no

PR #489, opened 2017-02-03
Catalog editor save button update

chalettu:
Moved the buttons to be right aligned instead of left aligned.  PV #https://www.pivotaltracker.com/story/show/138881749

### DIFF
--- a/client/app/components/catalogs/_catalogs.sass
+++ b/client/app/components/catalogs/_catalogs.sass
@@ -1,2 +1,2 @@
 .catalog-save-buttons
-        float: right
+  float: right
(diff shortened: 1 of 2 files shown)


client/app/components/catalogs/_catalogs.sass
```diff
@@ -0,0 +1,2 @@
+.catalog-save-buttons
```
AllenBW: rather than creating a new class, recommend using an existing one, `.pull-right` does just this thing.

chalettu:
@miq-bot add_label bug
@miq-bot add_label euew/no

chriskacerguis:
@chalettu needs changes and rebase.

chalettu:
catalog_new

PR #496, opened 2017-02-06
Fix for refresh of SUI $state.FeatureNav issue

chalettu:
Fix for Pivotal ticket https://www.pivotaltracker.com/story/show/138056447.

### DIFF
--- a/client/app/core/authorization.config.js
+++ b/client/app/core/authorization.config.js
@@ -72,30 +72,29 @@ export function authInit($rootScope, $state, Session, $sessionStorage, logger, L
 
       return;
     }
-    syncSession().then(rbacReloadOrLogin(toState, toParams));
+    syncSession().then(rbacReloadOrLogin(toState, toParams)).catch(badUser);
   }
   function syncSession() {
-    var deferred = $q.defer();
-        // trying saved token..
-    Session.create({
-      auth_token: $sessionStorage.token,
-      miqGroup: $sessionStorage.miqGroup,
+    return new Promise(function (resolve, reject) {
+      // trying saved token..
+      Session.create({
+        auth_token: $sessionStorage.token,
+        miqGroup: $sessionStorage.miqGroup,
+      });
+
+      Session.loadUser()
+        .then(Language.onReload)
+        .then(ServerInfo.set)
+        .then(ProductInfo.set)
+        .then(function () {
+          resolve();
+        })
+        .catch(badUser);
     });
-
-    Session.loadUser()
-      .then(Language.onReload)
-      .then(ServerInfo.set)
-      .then(ProductInfo.set)
-      .then(function() {
-        deferred.resolve("true");
-      })
-      .catch(badUser);
-
-    return deferred.promise;
   }
   function rbacReloadOrLogin(toState, toParams) {
     return function() {
-      if (RBAC.activeNavigationFeatures()) {
+      if (RBAC.navigationEnabled()) {
         $state.go(toState, toParams);
       } else {
         Session.privilegesError = true;
--- a/client/app/core/authorization.config.js
+++ b/client/app/core/authorization.config.js
@@ -45,11 +45,15 @@ export function authConfig($httpProvider) {
 }
 
 /** @ngInject */
-export function authInit($rootScope, $state, Session, $sessionStorage, logger, Language, ServerInfo, ProductInfo, $window) {
+export function authInit($rootScope, $state, Session, $sessionStorage, logger, Language, ServerInfo, ProductInfo, $window, $q, RBAC) {
   var unregisterStart = $rootScope.$on('$stateChangeStart', changeStart);
   var unregisterError = $rootScope.$on('$stateChangeError', changeError);
   var unregisterSuccess = $rootScope.$on('$stateChangeSuccess', changeSuccess);
 
+  $sessionStorage.$sync();  // needed when called right on reload
+  if ($sessionStorage.token) {
+    syncSession();
+  }
   function changeStart(event, toState, toParams, fromState, fromParams) {
     if (toState.data && !toState.data.requireUser) {
       return;
@@ -61,7 +65,6 @@ export function authInit($rootScope, $state, Session, $sessionStorage, logger, L
 
     // user is required and session not active - not going anywhere right away
     event.preventDefault();
-
     $sessionStorage.$sync();  // needed when called right on reload
     if (!$sessionStorage.token) {
       // no saved token, go directly to login
@@ -69,8 +72,11 @@ export function authInit($rootScope, $state, Session, $sessionStorage, logger, L
 
       return;
     }
-
-    // trying saved token..
+    syncSession().then(rbacReloadOrLogin(toState, toParams));
+  }
+  function syncSession() {
+    var deferred = $q.defer();
+        // trying saved token..
     Session.create({
       auth_token: $sessionStorage.token,
       miqGroup: $sessionStorage.miqGroup,
@@ -80,13 +86,16 @@ export function authInit($rootScope, $state, Session, $sessionStorage, logger, L
       .then(Language.onReload)
       .then(ServerInfo.set)
       .then(ProductInfo.set)
-      .then(rbacReloadOrLogin(toState, toParams))
+      .then(function() {
+        deferred.resolve("true");
+      })
       .catch(badUser);
-  }
 
+    return deferred.promise;
+  }
   function rbacReloadOrLogin(toState, toParams) {
     return function() {
-      if (Session.activeNavigationFeatures()) {
+      if (RBAC.activeNavigationFeatures()) {
         $state.go(toState, toParams);
       } else {
         Session.privilegesError = true;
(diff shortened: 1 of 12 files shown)


chalettu:
@miq-bot add_label bug
@miq-bot add_label euwe/no

client/app/core/layouts.config.js
```diff
@@ -19,25 +19,31 @@ function getLayouts() {
 }
 
 /** @ngInject */
-function enterApplication(Polling, lodash, $state, NavCounts, Navigation) {
+function enterApplication(Polling, lodash, $state, NavCounts, Navigation, Session, $timeout) {
```
AllenBW: Unused `Session` can be removed

client/app/core/layouts.config.js
```diff
@@ -19,25 +19,31 @@ function getLayouts() {
...
-    angular.forEach(value.secondary, function(secondaryValue, secondaryKey) {
-      lodash.merge(secondaryValue, $state.actionFeatures[secondaryKey]);
-    });
-  });
-
+ 
+  function initBadgeCount() {
+    if (angular.isDefined($state.navFeatures)) {
+      angular.forEach(NavCounts.counts, updateCount);
+      angular.forEach(Navigation.items, function (value, key) {
```
AllenBW: Unless you're using [arrow shorthand](http://babeljs.io/learn-es2015/#ecmascript-2015-features-arrows-and-lexical-this)

prefer use of named functions, similar to the way `updateCount` is used on line 27

chalettu:
@miq-bot add_label wip

client/app/components/services/service-explorer/service-explorer.component.js
```diff
@@ -40,7 +40,7 @@ function ComponentController($state, ServicesState, Language, ListView, Chargeba
       viewType: 'listView',
       cardConfig: getCardConfig(),
       listConfig: getListConfig(),
-      listActions: getListActions(),
+      listActions: {},
```
AllenBW: I believe this to be a symptom the problem, not the underlying issue.  A fix shouldn't mess with this component.

client/app/core/session.service.js
```diff
@@ -87,42 +91,56 @@ export function SessionFactory($http, $sessionStorage, gettextCatalog, $window,
...
 
   function setRBAC(productFeatures) {
+    var deferred = $q.defer();
     RBAC.set(productFeatures);
-
     // FIXME move these to the RBAC service as well
-    $state.actionFeatures = setRBACForActions(productFeatures);
-    $state.navFeatures = setRBACForNavigation(productFeatures);
+
+    setRBACForActions(productFeatures).then(function (actionsRBAC) {
```
AllenBW: Furthering the feature bleed of RBAC functionality isn't ideal (that lil fixme is on to something)

chalettu:
@miq-bot remove_label wip

chalettu:
@AllenBW Take a look.  This PR is finally ready for merging

client/app/core/session.service.js
```diff
@@ -83,99 +99,10 @@ export function SessionFactory($http, $sessionStorage, gettextCatalog, $window,
...
-
-    return angular.isDefined(serviceFeature);
-  }
-
-  function isAnyActionAllowed(actions, productFeatures) {
-    for (var i = 0; i <= actions.length; i++) {
-      if (angular.isDefined(productFeatures[actions[i]])) {
-        return true;
-      }
+    if (model.token !== null) {
```
AllenBW: `return angular.isObject(model.token)? model.token : false;`
will that work here? (trade five lines for one)

client/app/core/rbac.service.js
```diff
@@ -21,4 +30,101 @@ export function RBACFactory() {
...
+  function entitledForServices(productFeatures) {
+    var serviceFeature = lodash.find(actionFeatures, function(o) {
+      return o.show === true;
+    });
+
+    return angular.isDefined(serviceFeature);
+  }
+
+  function isAnyActionAllowed(actions, productFeatures) {
+    for (var i = 0; i <= actions.length; i++) {
```
jjlangholtz: Doesn't matter all too much since `actions` is a small array but this is a job for `Array#some`.

```javascript
return actions.some(action => angular.isDefined(productFeatures[action]);
```

That way the function always returns a boolean and iteration will stop as soon as possible.

client/app/core/authorization.config.js
```diff
@@ -61,16 +65,18 @@ export function authInit($rootScope, $state, Session, $sessionStorage, logger, L
...
       $state.transitionTo('login');
 
       return;
     }
-
-    // trying saved token..
+    syncSession().then(rbacReloadOrLogin(toState, toParams));
+  }
+  function syncSession() {
+    var deferred = $q.defer();
```
jjlangholtz: For new code I think we should avoid using the `$q` service and use the native ES6 `Promise` instead.

The same 3 lines in this function using `$q` could be rewritten as `.then(Promise.resolve())` in the last part of the `then` chain.

Also, after looking this over I think you can avoid adding the extra `.then` step and instead return the Promise chain directly:

```javascript
return Session.loadUser()
  .then(Language.onReload)
  .then(ServerInfo.set)
  .then(ProductInfo.set)
  .catch(badUser);
```

In either case there is no longer any `.catch` for `rbacReloadOrLogin` where there was before.

client/app/core/rbac.service.js
AllenBW: Even less! though eslint will holla about `action` not being in parenthesis

client/app/core/authorization.config.js
```diff
@@ -80,13 +86,16 @@ export function authInit($rootScope, $state, Session, $sessionStorage, logger, L
       .then(Language.onReload)
       .then(ServerInfo.set)
       .then(ProductInfo.set)
-      .then(rbacReloadOrLogin(toState, toParams))
+      .then(function() {
```
chalettu: Since this functionality is resolving the promise, I can't move this functionality out to another function since that would be out of scope.

client/app/core/layouts.config.js
```diff
@@ -19,26 +19,30 @@ function getLayouts() {
...
   angular.forEach(NavCounts.counts, updateCount);
   angular.forEach(Navigation.items, function(value, key) {
-    lodash.merge(value, $state.navFeatures[key]);
+    navFeatures[key] = lodash.merge(value, navFeatures[key]);
     angular.forEach(value.secondary, function(secondaryValue, secondaryKey) {
-      lodash.merge(secondaryValue, $state.actionFeatures[secondaryKey]);
+      actionFeatures[secondaryKey] = lodash.merge(secondaryValue, actionFeatures[secondaryKey]);
     });
   });
+  RBAC.setNavFeatures(navFeatures);
```
chalettu: Are you saying you want me to change it to allow setting one feature at a time rather than all at once?

chalettu: @AllenBW

chalettu:
@AllenBW I addressed all your feedback  but your last comment about moving RBAC.setNavFeatures in layout.config.js.  I can't move those functions up because they would be inside foreach statements that need to complete.

AllenBW:
ah! I see they are already in the function, ok no worries then. @chalettu

client/app/core/authorization.config.js
```diff
@@ -45,11 +45,15 @@ export function authConfig($httpProvider) {
 }
 
 /** @ngInject */
-export function authInit($rootScope, $state, Session, $sessionStorage, logger, Language, ServerInfo, ProductInfo, $window) {
+export function authInit($rootScope, $state, Session, $sessionStorage, logger, Language, ServerInfo, ProductInfo, $window, $q, RBAC) {
```
AllenBW: can now kill `$q` <3

client/app/core/rbac.service.js
```diff
@@ -21,4 +32,91 @@ export function RBACFactory() {
...
+      dashboard: {show: entitledForDashboard(productFeatures)},
+      services: {show: entitledForServices(productFeatures)},
+      orders: {show: entitledForServices(productFeatures)},
+      requests: {show: entitledForRequests(productFeatures)},
+      catalogs: {show: entitledForServiceCatalogs(productFeatures)},
+      dialogs: {show: entitledForService(productFeatures)},
+      administration: {show: entitledForService(productFeatures)},
+    };
+
+    return features;
```
AllenBW: Still returning on a setter, unless this isn't a setter, in which case it should be renamed.

client/app/core/rbac.service.js
```diff
@@ -21,4 +32,91 @@ export function RBACFactory() {
...
+      serviceEdit: {show: angular.isDefined(productFeatures.service_edit)},
+      serviceTag: {show: angular.isDefined(productFeatures.service_tag)},
+      serviceDelete: {show: angular.isDefined(productFeatures.service_delete)},
+      serviceReconfigure: {show: angular.isDefined(productFeatures.service_reconfigure)},
+      serviceRetireNow: {show: angular.isDefined(productFeatures.service_retire_now)},
+      serviceRetire: {show: angular.isDefined(productFeatures.service_retire)},
+      serviceOwnership: {show: angular.isDefined(productFeatures.service_ownership)},
+    };
+
+    return features;
```
AllenBW: Still returning on a setter, unless this isn't a setter, in which case it should be renamed.

PR #499, opened 2017-02-07
Fixed dashboard links filtering to service explorer

chalettu:
PV https://www.pivotaltracker.com/story/show/138836753

### DIFF
--- a/client/app/states/dashboard/dashboard.state.js
+++ b/client/app/states/dashboard/dashboard.state.js
@@ -188,6 +188,31 @@ function StateController($state, RequestsState, ServicesState, definedServiceIds
     $state.go('services');
   };
 
+  vm.navigateToRetiredServicesList = function() {
+    ServicesState.setFilters([{'id': 'retired', 'title': __('Retired'), 'value': true}]);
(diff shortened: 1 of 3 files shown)


chalettu:
@miq-bot add_label bug
@miq-bot add_label euwe/no

client/app/states/dashboard/dashboard.state.js
```diff
@@ -188,6 +188,31 @@ function StateController($state, RequestsState, ServicesState, definedServiceIds
     $state.go('services');
   };
 
+  vm.navigateToRetiredServicesList = function() {
```
AllenBW: Rather than three functions that complete the same action, why not one function that acts in three different ways?

food for thought

chalettu: I thought about doing a function that acted in 3 different ways but opted against it.   I could possibly combine them all into one larger function but that is more of a style preference.  As for the requests dashboard links, they weren't a part of this PR.  This PR addresses a specific bug caused by clicking the services links on the dashboard.

PR #500, opened 2017-02-08
Disabled action when console button is clicked

chalettu:
This is a fix for the scenario that the console and cockpit are unavailable and were still allowing the buttons to be clicked and the actions to be executed.  PV https://www.pivotaltracker.com/story/show/139248149

### DIFF
--- a/client/app/components/services/service-details/service-details.component.js
+++ b/client/app/components/services/service-details/service-details.component.js
@@ -284,11 +284,15 @@ function ComponentController($state, $window, CollectionsApi, EventNotifications
   }
 
   function openConsole(item) {
-    Consoles.open(item.id);
+    if (item['supports_console?'] && item.power_state === 'on') {
+      Consoles.open(item.id);
+    }
   }
 
   function openCockpit(item) {
-    $window.open('http://' + item.ipaddresses[0] + ':9090');
+    if (item['supports_cockpit?'] && item.power_state === 'on') {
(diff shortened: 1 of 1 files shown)


client/app/components/services/service-details/service-details.component.js
```diff
@@ -284,11 +284,15 @@ function ComponentController($state, $window, CollectionsApi, EventNotifications
...
-    Consoles.open(item.id);
+    if (item['supports_console?'] && item.power_state === 'on') {
+      Consoles.open(item.id);
+    }
   }
 
   function openCockpit(item) {
-    $window.open('http://' + item.ipaddresses[0] + ':9090');
+    if (item['supports_cockpit?'] && item.power_state === 'on') {
+      $window.open('http://' + item.ipaddresses[0] + ':9090');
```
chriskacerguis: Should we add support for https somehow?

chriskacerguis: @chalettu ^

chalettu: I wouldn't know where to make this optional.  Our existing cockpit URL is like this so I am not sure I would be inclined to just change it.

chriskacerguis: fair enough.

chriskacerguis:
@chalettu can you update this branch?

chalettu:
@chriskacerguis branch updated.

PR #502, opened 2017-02-09
PT#138881709 Add catalog feedback

AllenBW:
screenshot from 2017-02-09 12-44-01

https://www.pivotaltracker.com/story/show/138881709

### DIFF
--- a/client/app/core/catalogs-state.service.js
+++ b/client/app/core/catalogs-state.service.js
@@ -97,7 +97,7 @@ export function CatalogsStateFactory(CollectionsApi, EventNotifications, sprintf
     }
 
     function createFailure(response) {
-      EventNotifications.error(sprintf(__("There was an error creating catalog %s."), catalogObj.name));
+      EventNotifications.error(__(response.data.error.message));
 
       if (skipResults !== true) {
         return response.data;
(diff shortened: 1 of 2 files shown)


client/app/core/catalogs-state.service.js
```diff
@@ -97,7 +97,7 @@ export function CatalogsStateFactory(CollectionsApi, EventNotifications, sprintf
     }
 
     function createFailure(response) {
-      EventNotifications.error(sprintf(__("There was an error creating catalog %s."), catalogObj.name));
+      EventNotifications.error(__(response.data.error.message));
```
chriskacerguis: This might be a translation nightmare.  Wondering if we should do a message like:

There was an error creating catalog [NAME] <- translate this, then 
[translate] Server Responded: [/translate] [SERVER_ERROR_MESSAGE]

thoughts?

AllenBW: oooooo down with that, yeah server kicks back a message we can use, don't think server cares much for translation... or does it? 

@jntullo you seen server changing response messages based on language?

AllenBW:
image
@chriskacerguis

client/app/core/catalogs-state.service.js
jntullo: @AllenBW if it's an error from the model then it'll be translated

chriskacerguis:
@AllenBW it is saying it needs update.  Other than that 👍

PR #503, opened 2017-02-09
Cascading auto refresh for dialog fields

eclarizio:
This is very similar to https://github.com/ManageIQ/manageiq-ui-classic/pull/264, except for the service UI.

https://www.pivotaltracker.com/story/show/139224685

/cc @gmcculloug I'm not sure who is best to review this on the service UI side of things, please advise 😄

### DIFF
--- a/client/app/core/dialog-field-refresh.service.js
+++ b/client/app/core/dialog-field-refresh.service.js
@@ -14,20 +14,15 @@ export function DialogFieldRefreshFactory(lodash, CollectionsApi, EventNotificat
 
   function listenForAutoRefreshMessages(allDialogFields, autoRefreshableDialogFields, url, resourceId) {
     var listenerFunction = function(event) {
-      var dialogFieldsToRefresh = autoRefreshableDialogFields.filter(function(fieldName) {
-        if (event.originalEvent.data.fieldName !== fieldName) {
-          return fieldName;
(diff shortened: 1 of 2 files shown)


chriskacerguis:
@eclarizio could you please add the appropriate labels?

eclarizio:
@miq-bot add_label dialogs, enhancement

I think those are the correct labels, but I'm not 100% clear on if we need to backport just yet...

@gmcculloug Do we want to backport this to Euwe and Darga or do we think most people running into the issue that the cascading was meant to fix were using OpsUI and this is strictly an enhancement for the Service UI?

gmcculloug:
This was being worked on for BZ https://bugzilla.redhat.com/show_bug.cgi?id=1411541 which is for Darga.  So this should be backported to both Ewue and Darga.

eclarizio:
@miq-bot add_label euwe/yes, darga/yes

chriskacerguis:
@eclarizio also, to answer your question that would be me 😄

simaishi:
@eclarizio looks like there was some refactoring in 'master' and I won't be able to backport this PR to Euwe/Darga as is. Can you create a separate PR?

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq-ui-service/issues/512

simaishi:
Backported to Darga via https://github.com/ManageIQ/manageiq-ui-service/pull/513

PR #504, opened 2017-02-09
Ensure radio buttons that should be visible are actually visible

eclarizio:
While I was testing https://github.com/ManageIQ/manageiq-ui-service/pull/503 I found that radio buttons just weren't appearing even though they were supposed to be. I've changed the logic to look at when they should be visible to be consistent with the other field types.

Before:
screen shot 2017-02-09 at 10 52 23 am

After:
screen shot 2017-02-09 at 10 51 46 am

### DIFF
--- a/client/app/components/dialog-content/dialog-content.html
+++ b/client/app/components/dialog-content/dialog-content.html
@@ -87,12 +87,12 @@
                     </select>
 
                     <span
-                      ng-if="vm.inputDisabled"
+                      ng-if="dialogField.read_only || vm.inputDisabled"
                       ng-switch-when="DialogFieldRadioButton"
                       class="btn-group">
                       <label>{{ vm.parsedOptions[dialogField.name] }}</label>
                     </span>
-                    <span ng-if="vm.inputDisabled === false"
+                    <span ng-if="dialogField.read_only === false || vm.inputDisabled === false"
                       ng-switch-when="DialogFieldRadioButton"
                       class="btn-group">
                       <label


chriskacerguis:
@eclarizio could you please add the appropriate labels?

chriskacerguis:
@serenamarie125 does this meet your UX standards?

client/app/components/dialog-content/dialog-content.html
```diff
@@ -87,12 +87,12 @@ <h2 class="text-capitalize no-wrap">{{ ::vm.dialog.label }}</h2>
...
 
                     <span
-                      ng-if="vm.inputDisabled"
+                      ng-if="dialogField.read_only || vm.inputDisabled"
                       ng-switch-when="DialogFieldRadioButton"
                       class="btn-group">
                       <label>{{ vm.parsedOptions[dialogField.name] }}</label>
                     </span>
-                    <span ng-if="vm.inputDisabled === false"
+                    <span ng-if="dialogField.read_only === false || vm.inputDisabled === false"
```
AllenBW: **totally** could get away with `<span ng-if="dialogField.read_only || vm.inputDisabled" ... ` but one can't ever go wrong with being explicit.

chriskacerguis:
@eclarizio can you update, GH is saying it is out of date

serenamarie125:
@eclarizio vertically stacked radio buttons are preferred, would that be to update?  I also think this visualization needs work, but I just looked at PatternFly, and i can't find anything there for them.  I'll open a separate issue on the visualization once PF has them ( actually i assume if PF includes a style, it would come for free? )

eclarizio:
@serenamarie125 I'm actually not sure why they're inline, according to the bootstrap docs I could find, if you want inline you should have to add a class but that class does not exist.

I tried locally changing the span to a div and removing the button classes but that was showing them inline 😕 ...So I suspect there's some other CSS stuff going on that I think should be fixed as a separate issue since this was really supposed to just make them visible again.

chriskacerguis:
@eclarizio can you rebase this PR, it says it's out of date.

eclarizio:
@chriskacerguis You got it 👍

PR #514, opened 2017-02-14
Fix test:watch failures due to leaky state

jjlangholtz:
Fixes issue where state was being carried over in test runs causing 6
test cases to fail every time after the first run in watch mode.


chriskacerguis:
@jjlangholtz can you update, this seem to be out of date?

PR #518, opened 2017-02-15
Move all unit tests next to the units under test

jjlangholtz:
Lots of test file movement to move away from a parallel hierarchy to the
co-located pattern that is recommend in future angular versions.

A few dead files were removed and a few other files were renamed to keep
naming conventions consistent.

Going forward, unit tests should be placed at the same level as the
tested file with an identical filename except with a `.spec.js`
extension.

https://www.pivotaltracker.com/story/show/137095739


chriskacerguis:
@jjlangholtz welp, I'll wait until everything passes before I review ;)

jjlangholtz:
Yeahhh forgot that we were ignoring eslint for tests...

chriskacerguis:
@jjlangholtz so, FWIW - When we move to Angular2 all the tests will be redone, so I'd say let's keep it ignored for the time being.  @AllenBW thoughts?

AllenBW:
@chriskacerguis yah that was my thought...

PR #519, opened 2017-02-16
Extract 'app.shared' module from 'app.components'

jjlangholtz:
Begin breaking down the grab-bag components module into a more modular
feature structure. Pull all filters, directives, and components that are
used in more than one feature area into the `SharedModule`. This module
will then be imported into each individual feature module.

This sets up follow-up PRs to create feature modules for each area of
the application (services, catalogs, dialogs, etc.). Units specific to
each area have been marked in both `app.core` (factories/services) and
`app.components`.

Additionally, adjust the webpack config to include the sass root when
using `@import`. Now it is very easy to import sass files directly by
the components that use them instead of maintaining a parallel file
structure for styles.

https://www.pivotaltracker.com/story/show/138278069

@miq-bot add_label euwe/no, refactoring, housekeeping


client/app/shared/ss-card/ss-card.component.js
```diff
@@ -0,0 +1,11 @@
+import './_ss-card.sass';
+import templateUrl from './ss-card.html';
+
+export const SSCardComponent = {
+  bindings: {
+    header: '<',
+    description: '<',
+    image: '<',
+  },
+  templateUrl,
```
AllenBW: ` controllerAs: 'vm',` ? (ill make the case below)

client/app/shared/ss-card/ss-card.html
```diff
@@ -0,0 +1,9 @@
+<h3>{{ ::$ctrl.header }}</h3>
```
AllenBW: We use vm everywhere, its predictable, understood.  Yes `$ctrl` should also be predictable, understood, but I got this funky feeling about mixing the use of  both, if we choose to go the `$ctrl` route, it should be one or the other in one huge swoop rather that bits and pieces (as this sends mixed messages as to the desired and accepted form of our code base; which brings up another thought entirely, updating the style guide to say as much)

client/assets/sass/styles.sass
```diff
@@ -4,7 +4,6 @@
 @import '_bem-support/index'
 
 // Variables: colors, sizes, ...
-@import 'colors'
 @import 'app_colors'
 @import 'app_urls'
```
AllenBW: weary eyes might be missing something, but why not include `../../app/shared` in here and remove the` import 'app_colors'` from each file in shared that requires colors?

AllenBW: ditto for `@import '_bem-support/index'`

client/app/core/core.module.js
```diff
@@ -1,45 +1,75 @@
 /* global _:false, ActionCable:false, $:false, sprintf: false, moment: false */
+/* eslint-disable sort-imports */
```
AllenBW: :metal: :dagger: :skull_and_crossbones:

client/app/shared/ss-card/ss-card.html
jjlangholtz: Agreed! I'll make the change.

client/assets/sass/styles.sass
jjlangholtz: So the issue here is that we are heavily reliant on requiring the sass files in a specific order when we do not need to anymore. With webpack we can now treat all of our assets as modules and explicitly require them where they are used. This makes it far easier to understand where things are coming from and also whether or not they are being used, thus leading to more maintainable code.

Also it is necessary to do this in order to import the sass directly to the components that use them such as in: 

https://github.com/ManageIQ/manageiq-ui-service/pull/519/files#diff-2a87b2c4b10a6970565a04a140ae0345R1

client/app/core/core.module.js
jjlangholtz: The import order shouldn't matter here, I added the disable clause for now because I already went through and separated out the services per feature module to make the follow-up PR easier.

client/assets/sass/styles.sass
AllenBW: Hmm, for modules that are repeatedly required is there any greater convenience we can provide?

chriskacerguis:
@jjlangholtz can you resolve the conflicts and I'll merge?

PR #522, opened 2017-02-20
[EUWE] Make cockpit image path relative

jjlangholtz:
Use relative path so that the image is sourced from the SUI directory
instead of being requested from the MIQ server.

https://www.pivotaltracker.com/story/show/140246813
https://bugzilla.redhat.com/show_bug.cgi?id=1418066

@miq-bot add_label bug


simaishi:
@jjlangholtz Does this have an equivalent 'master' PR?

jjlangholtz:
@simaishi This image is actually not used on master anymore, I believe it was removed as part of another change although I am having trouble finding the commit.

PR #527, opened 2017-02-22
Set X-Miq-Group header for every request

jjlangholtz:
Break down `Session.create` into two distinct setters because the auth
token and miq group values come from two separate HTTP requests.

The group value is returned as part of the user object when the
authorizations request is made.

https://www.pivotaltracker.com/story/show/140246987
https://bugzilla.redhat.com/show_bug.cgi?id=1422551

@miq-bot add_label euwe/yes, bug
/cc @chalettu 

Separate follow-up PR in progress for backport.


simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq-ui-service/pull/528

PR #530, opened 2017-02-22
Update TaggingService to use bulk tag APIs

jjlangholtz:
Rename taggingService to TaggingService to maintain consistency with
other services.

Refactor `TaggingService#assignTags` to use the new bulk tag assign and
unassign endpoints. With this change the number of HTTP requests for
bulk assignments will be bounded at up to 2 (assign and/or unassign).

https://www.pivotaltracker.com/story/show/134447009

### DIFF
--- a/client/app/core/tagging.service.js
+++ b/client/app/core/tagging.service.js
@@ -1,5 +1,5 @@
 /** @ngInject */
-export function taggingService(CollectionsApi, lodash, exception) {
+export function TaggingService(CollectionsApi, lodash, exception) {
   var service = {
     assignTags: assignTags,
     findSharedTags: findSharedTags,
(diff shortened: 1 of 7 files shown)


jjlangholtz:

client/app/core/tagging.service.js
```diff
@@ -1,5 +1,5 @@
 /** @ngInject */
-export function taggingService(CollectionsApi, lodash, exception) {
+export function TaggingService(CollectionsApi, lodash, exception) {
```
AllenBW: `exception` doesn't appear to be used anywhere

jjlangholtz: Ah looks like it is used in one place: https://github.com/ManageIQ/manageiq-ui-service/pull/530/files#diff-73359eb751fda8d22574afaa8f29469cR81

Starting to wonder if there is some kind of linter we can set-up to catch these kinds of things.

AllenBW: doh! my :we yes: fail me, http://eslint.org/docs/rules/no-unused-vars can be configured for no-unused arguments

PR #532, opened 2017-02-22
PT#140438905 - Reconfigure delete action to send resource array of actionalble items

AllenBW:
https://www.pivotaltracker.com/story/show/140438905
Woops guys, was accidentally deleting vms instead of all snapshots on a vm, but we're good now.

Nothing to see here, except when doing batch actions, there will be an event notification for each item acted on, (this is normal and as it should be, some operations will pass others fail for a myriad of reasons) . So if you delete 100000 snapshots at once, the :earth_asia: might :boom:

### DIFF
--- a/client/app/services/vms.service.js
+++ b/client/app/services/vms.service.js
@@ -60,13 +60,13 @@ export function VmsService(CollectionsApi) {
     return filters;
   }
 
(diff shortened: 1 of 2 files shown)


chriskacerguis:
@AllenBW  https://www.youtube.com/watch?v=1ytCEuuW2_A

PR #533, opened 2017-02-22
Configure eslint to detect unused vars

jjlangholtz:
Modify eslint configuration to error when a variable is unused. In some
situations it actually makes sense not to use select positional variable
arguments. In those scenarios you can prefix the variable with `_` to
hint eslint (and human readers) that the variable is intentionally not
being used.

The `angular/on-watch` rule was turned off because it conflicts with the
unused var rule by encouraging the use of variables in situations that
don't make sense.

Surprisingly, enabling this rule revealed several minor bugs related to
the use the wrong variables.

@miq-bot add_label euwe/no, enhancement, housekeeping

/cc @AllenBW 😄

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -23,6 +23,7 @@
   },
   "rules": {
     "accessor-pairs": "error",
+    "angular/on-watch": "off",
     "array-bracket-spacing": "off",
     "array-callback-return": "off",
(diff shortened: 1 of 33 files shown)


AllenBW:
OMG HE DID IT!! 

I was gonna sign off for the eve, but now can't. MUST READ! <3

PR #555, opened 2017-03-02
Ensured that sorting gets applied every time we query for requests.

chalettu:
Fix for PV https://www.pivotaltracker.com/story/show/140565343.  Ensured sort choices continue to work after polling occurs


chalettu:
@miq-bot add_label euwe/no
@miq-bot add_label bug
@miq-bot add_label requests

PR #556, opened 2017-03-02
Bootstrap both AngularJS and Angular frameworks

jjlangholtz:
Going forward AngularJS refers to the 1.x framework while Angular refers
to the 2+ framework.

Automatic bootstrapping using `ng-app` in `index.ejs` is not supported
for Angular and hybrid applications. Instead, manually bootstrap both
frameworks from `app/main.ts` and make it the new webpack entry point.

The new entry point imports polyfills required for Angular to operate
correctly and then imports the existing AngularJS bootstrapping process
from `app.js`.

This change also adds a number of new dependencies and peer dependencies
required to run Angular.

Aside from a slightly larger output bundle, the application should
function exactly the same as it did before. The only difference is that
we can now add new dependencies and units in Angular.

https://www.pivotaltracker.com/story/show/140847211

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,12 @@
     "vet": "eslint \"client/**/*.js\" && sass-lint --verbose --no-exit"
   },
   "devDependencies": {
+    "@angular/common": "4.0.0-rc.2",
+    "@angular/compiler": "4.0.0-rc.2",
+    "@angular/core": "4.0.0-rc.2",
+    "@angular/platform-browser": "4.0.0-rc.2",
+    "@angular/platform-browser-dynamic": "4.0.0-rc.2",
+    "@angular/upgrade": "4.0.0-rc.2",
     "actioncable": "5.0.1",
     "angular": "1.6.1",
     "angular-animate": "1.6.1",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -125,10 +125,19 @@
     "yargs": "6.6.0"
   },
   "dependencies": {
+    "@angular/common": "4.0.0-rc.2",
```
AllenBW: All as dependencies?  Can most of the @angular stuff can be `devDependencies`?

jjlangholtz: Hmm... I think it works the same in either case since we are always using `yarn install` in each environment. I installed them as `dependencies` just to follow the conventions that Angular follows in all of their `package.json`s.

From the [cli](https://github.com/angular/angular-cli/blob/master/package.json), for example, looks like they partition the actual dependencies that compose the application in `dependencies` and those that exist outside of a bundle in `devDependencies`.

I can move everything to `devDependencies` though for consistency.

AllenBW: ok soooo I've always thought `dependencies` are required to run, `devDependencies` are those things used for development.. since we run with da bundles it kinda makes most things `devDependencies`

http://stackoverflow.com/questions/18875674/whats-the-difference-between-dependencies-devdependencies-and-peerdependencies#22004559

AllenBW: The fact that we don't ship anything bundled makes this kinda wonky though... :thinking:

chriskacerguis:
@jjlangholtz any idea why travis keeps failing?

jjlangholtz:
@chriskacerguis Looking into it now, but it is an issue with Protractor understanding when the frameworks are bootstrapped.

PR #485, opened 2017-02-02
Fix missing Cockpit URL opening.

martinpovolny:
https://bugzilla.redhat.com/show_bug.cgi?id=1409956

This is just a minimal fix for the BZ. Moving forward we need to have a API action to acquire the Cockpit URL so that we can utitilize work done in https://github.com/ManageIQ/manageiq/pull/12506 (SSO).

PR #560, opened 2017-03-06
[EUWE] Fix opening of Cockpit consoles.

martinpovolny:
Euwe version of https://github.com/ManageIQ/manageiq-ui-service/pull/485

https://bugzilla.redhat.com/show_bug.cgi?id=1409956


martinpovolny:
I have no idea what this check is or why it is failing. ping @chriskacerguis

chriskacerguis:
@martinpovolny it checks for any security vulnerabilities.  However, in this case, it looks like it was an error on the check side (we can ignore it).

PR #563, opened 2017-03-06
Changed out filters from the dashboard work on explorer pages

chalettu:
PV https://www.pivotaltracker.com/story/show/141146471.  Changed how filters get passed to explorer pages to help fix an issue with the filters not clearing when services is clicked from left nav


chalettu:
@miq-bot add_label bug
@miq-bot add_label euwe/no
@miq-bot add_label services

PR #570, opened 2017-03-07
Updated our yarn command and turned on caching

chalettu:
This is a proposed enhancement that would just have us use the regular yarn install command so we can leverage caching.  This also turns on travis’s support for yarn caching

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,13 @@ node_js:
 - 6.9.1
 sudo: false
 cache:
+  yarn: true
   directories:
   - node_modules
 before_install:
 - npm install -g yarn@0.19.1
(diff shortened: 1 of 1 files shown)


chalettu:
@miq-bot add_label euwe/no

AllenBW:
I wanta add this thought to the record... 'Member when things are wonky with your build and blow away `./node_modules` and fetch anew? So we gain speed (of travis completion) at he cost of stability?

chalettu:
I here you about wonkiness and if it starts to become a regular problem we can easily switch it back.  Travis also has a button that maintainers can hit that clears out stored cache if we ever want to flush it.

PR #582, opened 2017-03-13
Enable RBAC on Requests tab

chalettu:
Fix for PV https://www.pivotaltracker.com/story/show/141617155.  Enabling RBAC on requests tab.

### DIFF
--- a/client/app/requests/requests-state.service.js
+++ b/client/app/requests/requests-state.service.js
@@ -8,8 +8,8 @@ export function RequestsStateFactory(ListConfiguration, RBAC) {
 
   function getPermissions() {
     const permissions = {
-      approval: RBAC.hasAny(['miq_request_approval','miq_request_admin','miq_request','miq_request_control']),
-      edit: RBAC.hasAny(['miq_request_edit','miq_request_admin','miq_request','miq_request_control']),
+      approval: RBAC.hasAny(['miq_request_approval']),
+      edit: RBAC.hasAny(['miq_request_edit']),
     };
 
     return permissions;
--- a/client/app/requests/requests-state.service.js
+++ b/client/app/requests/requests-state.service.js
@@ -8,8 +8,8 @@ export function RequestsStateFactory(ListConfiguration, RBAC) {
 
   function getPermissions() {
     const permissions = {
-      approval: RBAC.has('miq_request_approval'),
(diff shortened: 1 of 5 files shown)


chalettu:
@miq-bot add_label euwe/no
@miq-bot add_label requests

client/app/states/requests/details/details.state.js
```diff
@@ -1,21 +1,24 @@
 import templateUrl from './details.html';
 
 /** @ngInject */
-export function RequestsDetailsState(routerHelper) {
-  routerHelper.configureStates(getStates());
+export function RequestsDetailsState(routerHelper, RBAC) {
```
AllenBW: aww this isn't a component yet :( was gonna say we should throw this in the component not the state, but i guess that'll be done at a later time

client/app/states/requests/details/details.state.js
```diff
@@ -1,21 +1,24 @@
...
-function getStates() {
+function getStates(RBAC) {
   return {
     'requests.details': {
       url: '/:requestId',
       templateUrl,
       controller: RequestDetailsController,
       controllerAs: 'vm',
-      title: N_('Request Details'),
+      title: __('Request Details'),
```
AllenBW: good call. 
http://manageiq.org/docs/guides/i18n

client/app/requests/requests-state.service.js
```diff
@@ -1,8 +1,19 @@
...
-export function RequestsStateFactory(ListConfiguration) {
-  var service = {};
+export function RequestsStateFactory(ListConfiguration, RBAC) {
+  var service = {
+    getPermissions: getPermissions,
+  };
 
   ListConfiguration.setupListFunctions(service, { id: 'requested', title: __('Request Date'), sortType: 'numeric' });
 
+  function getPermissions() {
```
AllenBW: Does this also need `miq_request`?

AllenBW: serious question... thinking that maybe not... because an admin. will have these checked... but.... not positive...

chalettu: I looked at it and it looks like those permissions exist for an admin, so I added them in.  It may technically be duplicitous but we can just be explicit in our permissions.

PR #583, opened 2017-03-13
Fix services filter count and pagination controls

jjlangholtz:
Pagination controls for services were not usable because the
`filterCount` was not being updated. The failure to update was due to an
attempt to call a previously moved function.

Update to call `getFilters` from the new location.

@miq-bot add_label euwe/no, bug, services


AllenBW:
@jjlangholtz feeling kinda facepalm as this one was on me, thanks!! 🙇

PR #589, opened 2017-03-15
Updated how navigation translates menu item titles

chalettu:
PV . https://www.pivotaltracker.com/story/show/141818307.  Ensured that Menu item titles can be translated.


chalettu:
@miq-bot add_label euwe/no
@miq-bot add_label bug

PR #596, opened 2017-03-20
PT#141947277 - denote retired resources

AllenBW:
https://www.pivotaltracker.com/story/show/141947277

@chalettu @chriskacerguis @jjlangholtz hey ya'll could use feedback on the approach here, mainly the use of eventnotification and the removal of the resource actions (we could just disable them, but my gut says that isn't going far enough... if one can't act on the resource, one shouldn't see actions on a resource)

## Action 🛒 
image

☝️ so a few details about this, warning banner does **not** update unread notification count, also it persists for a really long time and follows the user anywhere on the page, totally dismissable

image

In this shot we see two disabled resources, in addition to the row being disabled, all actions are removed from the resource and a "retired" icon is added in place of the power icon.  Maybe add a column  stating "retired resource" in addition to the other cues in place? 

image
☝️ a retired and active resource !

### DIFF
--- a/client/app/core/event-notifications.service.js
+++ b/client/app/core/event-notifications.service.js
@@ -277,7 +277,7 @@ export function EventNotificationsFactory($timeout, lodash, CollectionsApi, Sess
 
   function showToast(notification) {
     notification.show = true;
-    notification.persistent = notification.type === 'danger' || notification.type === 'error';
+    notification.persistent = notification.data.persistent || notification.type === 'danger' || notification.type === 'error';
     state.toastNotifications.push(notification);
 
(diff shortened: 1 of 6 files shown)


chriskacerguis:
So, I like this a lot.  I agree with you, to remove the actions if it is retired.  I think that the one thing that "worries" me is if the retired banner goes away (I like the banner), but wondering if we should put something else in there (something permanent) and indicates on the details page that it is retired as well?

AllenBW:
We can make the banner permanent and also add a row to indicate instance status

AllenBW:
@chriskacerguis Updated 🌮 💃

PR #603, opened 2017-03-22
PT#140975359 secure headers

AllenBW:
https://www.pivotaltracker.com/story/show/140975359

no visual changes, requires no ss

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,6 +136,7 @@
   },
   "dependencies": {
     "express": "4.15.2",
+    "helmet": "3.5.0",
     "http-proxy": "1.16.2",
     "morgan": "1.8.1",
     "serve-favicon": "2.4.1"
--- a/server/app.js
+++ b/server/app.js
@@ -1,23 +1,25 @@
 /* eslint-disable no-undef, no-console, no-process-env, angular/log, no-path-concat */
 
-'use strict';
 const path = require('path');
-var http = require('http');
-var express = require('express');
-var bodyParser = require('body-parser');
-var favicon = require('serve-favicon');
-var logger = require('morgan');
-var httpProxy = require('http-proxy');
-var four0four = require('./utils/404')();
-var proxyService = require('./utils/proxy')();
-var serviceApp = require('./utils/serviceApp');
-var serviceApi = require('./utils/serviceApi');
-var wsProxy = require('./utils/wsProxy');
-var buildOutputPath = process.env.BUILD_OUTPUT || './';
-var app = express();
+const http = require('http');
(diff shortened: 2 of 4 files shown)


martinpovolny:
@AllenBW, @chriskacerguis: gentleman, given the Pivotal task references my name, I'd appreciate a ping at least.

But to the point: I understand that on an appliance the app is running from staticaly served precompiled assets. Is that right?

If so I don't see how this is going to help. Can you explain, please?

I emphasize that I have not looked into the issue and, please, correct me if I am wrong, but my first thought is that this needs to be done on the Apache side.

martinpovolny:
I am trying to understand what you actually did.

Seems the headers are being added for the javascript server serving the SUI assets in dev-mode. Right?

Seems to me that only this line: 

https://github.com/ManageIQ/manageiq-ui-service/pull/603/commits/71447caf4be36dee0665a42a735ce1c170284664#diff-9e4a215b4f5f613a802fa43d8ab35270R22 is really relevant for the topic of this PR. 

The rest seems to be refactoring/cleanup is that correct?

Based on recent talks about 100% test coverage I wonder if there's any test coverage for this. I'd love to learn how this is done with the stack of this project.

chriskacerguis:
@martinpovolny - This was a follow up task from the email chain where we were discussing security issues with secure_headers.  From the email:

> Action item: I am afraid that we do not have an equivalent for the secure_headers for the assets of SUI. We should add that!

Regarding a few other things, yes you are correct any security headers need to be added by Apache (as everything is served out of that in production).  However, due to the concerns for security, we added the secure headers to be added when running in "dev" mode.

Regarding the title, I agree, I think that this is an older ticket and "newer" tickets have the correct title.

I'll let @AllenBW talk more specifically about the code itself.

AllenBW:
Hi @martinpovolny :wave: We only use app.js when deploying to heroku, so this won't help at any other time, but then again sui doesn't "control itself" any other time :laughing:  You are correct that line and adding the dependency to the app are the primary changes, rest is refactoring.  Not sure what ya mean by test coverage for this, if helmet operates as expected?

martinpovolny:
@AllenBW : on test coverage: I don't know, I hoped to learn that here :-D

Would be insterested in reading a test for that and learn something new!

Please, ping me in a PR where you are going to deal with the appliance part of the action item. I'll really appreciate that!

chriskacerguis:
@martinpovolny will do!  RE: Testing, are you looking for something specific?  We can point you do some PR's with testing if that is what you are looking for.  Happy to help in any way we can, just let me know.

PR #620, opened 2017-03-31
Fixed explorers count result delay

chalettu:
PV https://www.pivotaltracker.com/story/show/140348513

### DIFF
--- a/client/app/requests/order-explorer/order-explorer.component.js
+++ b/client/app/requests/order-explorer/order-explorer.component.js
@@ -243,16 +243,18 @@ function ComponentController($filter, lodash, ListView, Language, OrdersState, E
     if (!refresh) {
       vm.loading = true;
     }
-    getFilterCount();
     var existingOrders = (angular.isDefined(vm.ordersList) && refresh ? angular.copy(vm.ordersList) : []);
 
     vm.offset = offset;
-    OrdersState.getOrders(
-      limit,
-      offset,
-      OrdersState.getFilters(),
-      OrdersState.getSort().currentField,
-      OrdersState.getSort().isAscending).then(querySuccess, queryFailure);
+    getFilterCount().then(() => {
+      OrdersState.getOrders(
+        limit,
+        offset,
+        OrdersState.getFilters(),
+        OrdersState.getSort().currentField,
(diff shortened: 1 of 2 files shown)


chalettu:
@miq-bot add_label fine/yes
@miq-bot add_label bug
@miq-bot add_label services
@miq-bot add_label orders

client/app/services/service-explorer/service-explorer.component.js
```diff
@@ -340,7 +341,7 @@ function ComponentController($state, ServicesState, Language, ListView, Chargeba
       vm.loading = false;
     }
     vm.offset = offset;
-
+    getFilterCount();
```
AllenBW: samsies here

client/app/requests/order-explorer/order-explorer.component.js
```diff
@@ -243,6 +243,7 @@ function ComponentController($filter, lodash, ListView, Language, OrdersState, E
     if (!refresh) {
       vm.loading = true;
     }
+    getFilterCount();
```
AllenBW: thinking ya might want to throw this in the handleSuccess, otherwise we'll be pounding the user with error messages (in the case it fails)

AllenBW:
hmmm, ok given the nature of this, maybe reverse this, do the get minimal first then in that success go the getServices...  so we'll have our result count before we have our results...

these some api magic that would give us both on one call (unpaginated results of the query, but it hasn't been high enough priority to implement? or something like that)

client/app/requests/order-explorer/order-explorer.component.js
chalettu: I could probably chain these to where the filter count goes first and handles the other query in its success handler.   I thought about doing it but didnt want to shift too much code around but you have a point about multiple error messages.

simaishi:
Fine backport details:

```
$ git log -1
commit 2d8496a9d7b34af6d9627f2f2e3c1f5b441a89b8
Date:   Fri Mar 31 17:29:31 2017 -0500

    Merge pull request #620 from chalettu/explorer-result-counts
    
    Fixed explorers count result delay
    (cherry picked from commit fd02705361d8cf627b0e0cf156c963b00debf065)
```

PR #637, opened 2017-04-05
Relax node & npm engine version constraits to anything greater

himdel:
Right now, we depend exactly on node 6.9.1 and npm 3.10.3, which doesn't make much sense.

Adding a `>=` to the version constraint.

An alternative would be to use the tilde `~ 6.9.1`, but.. this makes sense for dependencies, but maybe not so much for engines. (We're not expecting node 8 to break stuff.)

Should fix the docker build `error manageiq-api-mock@1.0.0: The engine "node" is incompatible with this module. Expected version "6.9.1".`

Cc @bdunne, @simaishi


himdel:
(Same change as https://github.com/ManageIQ/manageiq-api-mock/pull/3)

chriskacerguis:
@himdel I would think that we would want to required an exact version to avoid any potential issues with node getting upgraded without us knowing it?  Thoughts?

himdel:
Well, I disagree, and not only because there's no such RPM for that specific version :).

If you want to be extra sure, I can change that to the tilde, so that we depend on `6.9.*` instead of anything greater. But then we'll need to change that to `~6.9.1 | ~7.0.0 | ~8.0.0` once people start using those..

Mostly, I don't see the point, if you look at https://nodejs.org/en/blog/release/v7.0.0/, doesn't seem like there's any breaking changes in version 7, and .. well, unless you do somethign weird, any node version will do.

chalettu:
I think dot revisions within the same major point versions of node is acceptable but not just a blanket >=.  Forward versions of node could possibly break things or create incompatibility with dependencies we have.

himdel:
Allright, changed to tildes, but please keep this updated for future builds :).

chriskacerguis:
@himdel makes sense to me.

chriskacerguis:
@chalettu I get what you are saying but @himdel makes a good point about the RPM.  It's just going to be whatever the latest LTS Node is.  So, it really doesn't make sense to use.

himdel:
So.. @chriskacerguis I guess it's up to you.. tilde or `>=`? :)

Either should fix the build for now :)

chalettu:
Thinking more about it, we would probably catch any issue before it arises downstream so I think we would be safe for the most part.  Its probably fine to relax to relax it.  We can always lock it back down should an issue ever arise.

chriskacerguis:
@himdel I'd go with >=

Let's give @AllenBW and @jjlangholtz a chance to respond.

himdel:
Ok, changed back to `>=` for now, we can always restore f87d47b2 again :).

himdel:
@miq-bot add_label fine/yes

(gitter [April 5, 2017 3:13 PM](https://gitter.im/ManageIQ/manageiq/ui?at=58e509a38fcce56b20d1e33b))

simaishi:
Fine backport details:

```
$ git log -1
commit 0fdf0a487b0c091601fcdf081d359a3849f80b4e
Date:   Wed Apr 5 11:05:06 2017 -0500

    Merge pull request #637 from himdel/node-engine
    
    Relax node & npm engine version constraits to anything greater
    (cherry picked from commit d4ef587465b1ef9f409a44c5a6acb7de56fe16ea)
```

PR #619, opened 2017-03-31
Fix for service dialog being submitted before entry point code on dynamic fields has completed execution

eclarizio:
This fix will simply disable the add to cart button when dynamic fields are processing so that it's not possible to submit the dialog while the automate methods are still running.

https://bugzilla.redhat.com/show_bug.cgi?id=1431688

@jjlangholtz Can you review, please?

/cc @gmcculloug 

@miq-bot add_label bug

PR #668, opened 2017-04-11
[EUWE] Fix for service dialog being submitted before entry point code on dynamic fields has completed execution

eclarizio:
Backport of #619, fixing conflicts from the cherry-pick.

Conflicts were minor, tests were using the incorrect path and the main file that changed in #619 was a different file in Euwe so the changes just needed to be moved over.

/cc @simaishi


simaishi:
@eclarizio can you take a look at the travis error?

PR #697, opened 2017-04-20
Removes accidential title  as a result of using reserved word

AllenBW:
Hover over any field would display a caption that does not align with ux, (tooltips is our jam n stuff), this changes up the offending component param name.

again ignore the color, gnome is on drugs...

### before, you'll notice the text bubble by the mouse
screen shot 2017-04-20 at 1 59 51 pm

### after, no text bubble by the mouse
screenshot from 2017-04 -20 13-48-12

@serenamarie125 if this isn't a bug we can just close this pr

### DIFF
--- a/client/app/services/detail-reveal/detail-reveal.component.js
+++ b/client/app/services/detail-reveal/detail-reveal.component.js
@@ -5,7 +5,7 @@ export const DetailRevealComponent = {
   controller: ComponentController,
   controllerAs: 'vm',
   bindings: {
-    title: '@',
+    detailTitle: '@',
     detail: '@',
     icon: '@',
     translateTitle: '<',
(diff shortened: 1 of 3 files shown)


chriskacerguis:
@AllenBW do we need UX review for this, since we are addressing a bug?  #justaskin

AllenBW:
@chriskacerguis yeah want to confirm intended behavior is not to have multiple types of tooltips

chriskacerguis:
@ManageIQ/core-ux for your review

cshinn:
As long as it doesn't affect accessibility in some way to remove the title text, I don't see a reason to have a tooltip that displays no additional information. Especially if it looks different than our default one.

chriskacerguis:
@cshinn is that a 👍 or 👎 (just checking) :)

serenamarie125:
@AllenBW I'm not sure i understand.  Is the issue content or visual?  

Is your request to remove this from every field?

I agree with @cshinn in that if the tooltip just matches the field name, it is no value.  But do we know that here?

In the case where there is just an icon, and the tooltip is representing a state, I think it's valuable to show a tooltip.

That being said, we should be using PatternFly tooltips (http://www.patternfly.org/pattern-library/widgets/#tooltip)

AllenBW:
@serenamarie125 issue is visual. left side of the table has the title right side has the details, the tooltip, that does not conform with patternfly convention, appears when you hover over each row, and displays duplicates the title

serenamarie125:
If it's just a visual issue, use the PatternFly tooltip, don't remove it.
If it's a content issue, let's use the correct content in the tooltip :)

AllenBW:
? its an unintended side affect of using the incorrect attribute, the fields aren't meant to have any kind of value on mouseover

PR #712, opened 2017-04-24
Update translation .pot

AllenBW:
https://www.pivotaltracker.com/story/show/144103713

Part one of two, second part will be to upload compiled translations.... stay tuned!

PR #713, opened 2017-04-25
Extracted translations post removal of requests tab

AllenBW:
what https://github.com/ManageIQ/manageiq-ui-service/pull/712 should have been

@mzazrivec so the new gettext extraction, does that need to be backported? I'm thinking so, also it should overwrite the previous rather than merge correct (for @simaishi )


mzazrivec:
@AllenBW Backported where? To other branches? No. You don't backport catalog PRs.
For catalog updates, you need to run the extraction and create PRs for every separate
branch.

AllenBW:
@mzazrivec thanks for that! since fine is presently tracking master (more or less) I'll wait till the removal of requests is backported to fine before I run the extraction there

PR #665, opened 2017-04-11
PT#143511261-Where possible, update responses to batch operations

AllenBW:
https://www.pivotaltracker.com/story/show/143511261
depends on https://github.com/ManageIQ/manageiq-ui-service/pull/664

Affects
- delete catalogs
- delete orders
- approve/deny requests/orders
- set-ownership services
- remove services

A number of api actions still return an object instead of a results array, notifications for these actions have remained intact

a few ss of the action

ownership
service-delete
order-removing
catalog-delete

PR #732, opened 2017-05-02
PT#140440299 - Refactor notificaitons

AllenBW:
https://www.pivotaltracker.com/story/show/140440299

* Refactor EventNotification to support info, warn, error, success types as per https://github.com/ManageIQ/manageiq/blob/master/app/models/notification_type.rb#L9
* EventNotifications.bach now accepts results of batch operation, for all actions executed over 6 items, rather than spamming the user with toasts, the correct notification tray dropdown is indicated with the action results
* Refactoring for cleanliness/clarity event notifications service and navigation controller

as with https://github.com/ManageIQ/manageiq-ui-service/pull/665, only those operations that support batch feedback will be benefit from these efforts

### rather than being spammed with 9 toasts, a single toast indicating the results of the batch operation is shown
screenshot from 2017-05-03 13-38-15
### aftermath of batch operation
screenshot from 2017-05-03 13-38-24
### we still show individual toasts up to 6 
screenshot from 2017-05-03 13-45-41
### Different warning types show up in different notification dropdowns
screenshot from 2017-05-03 13-39-12


AllenBW:
@chriskacerguis soooooo we probably want fine/no for this right?

chriskacerguis:
Yeah, fine/no I wouldn't get consider this a blocker.

PR #751, opened 2017-05-10
[WIP] PT#145115749 - Suppresses the creation of duplicate notifications (by message and type)

AllenBW:
This pr affects when toasts appear, only if there is unique notification (by type and message), duplicate ones are suppressed.  That being said, if a polling error notification is made then cleared, it will appear again. 

This is gonna be great for e2e testing, no longer will we have to worry about toast elements overlapping buttons 🙌 

## The before case, death by toasts
screen shot 2017-05-09 at 3 37 38 pm

## The after case, a world where toast only help
screen shot 2017-05-10 at 9 34 24 am

### DIFF
--- a/client/app/core/event-notifications.service.js
+++ b/client/app/core/event-notifications.service.js
@@ -157,20 +157,23 @@ export function EventNotificationsFactory($timeout, lodash, CollectionsApi, Sess
 // Private
   function add(notificationType, type, message, notificationData, id) {
     const group = lodash.find(state.groups, {notificationType: notificationType});
-    const newNotification = {
-      id: id,
-      notificationType: notificationType,
-      unread: angular.isDefined(notificationData) ? notificationData.unread : true,
-      type: type,
-      message: message,
-      data: notificationData || {},
-      href: id ? '/api/notifications/' + id : undefined,
-      timeStamp: (new Date()).getTime(),
-    };
+    // Ignoring duplicate notifications of the same type
+    if (!lodash.find(group.notifications, {message: message})) {
+      const newNotification = {
+        id: id,
+        notificationType: notificationType,
+        unread: angular.isDefined(notificationData) ? notificationData.unread : true,
+        type: type,
+        message: message,
+        data: notificationData || {},
+        href: id ? '/api/notifications/' + id : undefined,
+        timeStamp: (new Date()).getTime(),
+      };
 
-    group.notifications.unshift(newNotification);
-    updateUnreadCount(group);
-    showToast(newNotification);
+      group.notifications.unshift(newNotification);
+      updateUnreadCount(group);
+      showToast(newNotification);
+    }
   }
 
 


chriskacerguis:
Let's review this on Friday with @serenamarie125 at our meeting before we merge.

serenamarie125:
As i mentioned in gitter on the 9th, we should come up with a solution which we can promote to all products.  I definitely see the need for a solution, but want to make sure that we have designers involved, and that if we propose something for SUI, that it is also implemented in the Ops UI @dclarizio .  Inconsistencies between CloudForms UIs is killing us.  If those inconsistencies are being introduced intentionally, we should be sure that all parties are aware, support it and we have a mitigation plan in place if needed

chriskacerguis:
Sounds good. We'll mark this as a WIP and comeback to it when we have a plan, also should include anything to ensure the CUI and SUI notification drawers align.

AllenBW:
@chriskacerguis @serenamarie125s  

Want to retouch on this pr.  Urgency is with regards to polling, SUI does it, opsui kinda doesn't, when polling fails, it looks real bad having that long list of errors...
screen-shot-2017-05-24-at-9 52 58-am

serenamarie125:
@AllenBW - rather than providing a solution to suppress duplicate notifications, I think we should consider fixing the route of the problem.  Why are so many duplicate notifications being issued in a short timeframe?

chriskacerguis:
@serenamarie125 so there are 2 use cases that I am aware of where this appears.

1. Since we now do polling in the background, if the MIQ backend becomes unavailable, each time the system polls, it will generate an alert
2. In the case of a bulk operation, you would get lots of alerts as things completed.

@AllenBW feel free to chime in, if I missed something.

AllenBW:
Yep that sums it up. @serenamarie125 polling is the difference and the issue here, if you bang the refresh button in ops you can reproduce this issue.  I am of the opinion that the fact that you can produce two identical notification is a core issue.

serenamarie125:
@chriskacerguis @AllenBW So, rather than suppresses the creation of all duplicate notifications, can you do so specifically for these individual use cases where the MIQ backend becomes unavailable?

AllenBW:
@serenamarie125 So the backend becoming unavailable manifests a myriad of notifications depending on where the user is in the application.  

Is there a case where in we would want to have multiple identical notifications (genuinely wondering, I couldn't think of any use case)?  If so, maybe putting these types of notifications in the drawer rather than spamming the view portal with toasts is a more polite approach.

serenamarie125:
@AllenBW Geez I feel like my GH comments got lost.  

Yes, there are certain situations where toasts are not needed.
The case I'm thinking of identical notifications being shown as being important is invalid login.  In this case, this could be a security issue, and would be very important.

Again, IMO this comes back to identifying the particular situation, and handling it appropriately.  I agree that if the MIQ backend is unavailable, we do not need to continue to notify the end user repetitively with toasts.  But I would think that for every user action which finds that the backend is unavailable should present one toast, so they realize why their action isn't being processed.

There's a longer term issue which we are all aware of, in that we need a notification management system around all these "conditions".  Admins should be able to select default notification types for all conditions which are detected, and users should be able to change their own default notifications as well.  But, that's a whole nother can of worms 😃

AllenBW:
Bahahaha I hear ya :P JUST TODAY I realized there was a mention I neglected from 4 days agooooooo 👎 

So a case by case basis is difficult to handle, but showing the latest (not duplicate) toast and logging all toasts (even duplicates) might is the best catch all.  

For login, sui doesn't use toasts, we do the whole error thing on the page so we're good there...

Agreed so very much agreed that other can of worms is something well worth attention.  That being said, looking to do whatever we can do at this time.  Which is the big question, what can/are we approved to do at this time? Archive this as a food for thought bag for that bigger can meal? Do the thing above?  🤔

serenamarie125:
@jeff-phillips-18 can you chime in here?  I know that you have been involved in implementing the Notification Drawer in a few products

jeff-phillips-18:
@serenamarie125 My first thought here is to leave a single toast and add a counter for the number of times it has been repeated. Second thought is for the code that is generating multiple alerts for the same issue to track that the issue exists and not re-create the notifications until the condition has been resolved. 

The worry I have with suppression is that somewhere the determination is being made that the notification is in fact a duplicate and that the user is no longer concerned about that alert. If the alert exists but is hidden, is the alert re-shown? How do I know when the condition no longer exists? How do I know the condition still exists and is causing me problems that I may not know are the result of the condition?

I have seen this in past experiences use 'repeated N times' effectively, updating the time stamp each time it re-occurs.

AllenBW:
Gonna do a tl;dr of my proposal for posterity.  Key verbal distinction is notifications vs toasts (things in the tray vs on the screen).

1. Older duplicate **toasts** are not shown,(duplication is determined by notification string, or anything we want).  
an example of this:
```
- notification "foo! foo! foo!" timestamp: 0800
- notification "foo! foo! foo!" timestamp: 0801
```
**Present state reaction:** to this is to show both as notifications **and** toasts, fill up the screen and then some
**Proposed state reaction:** to show the 0800 toast then **replace** it with the 0801 toast when it is encountered, both notifications will remain in the tray

sorry for introducing the word suppression, much regret :( 😭 makes it seem like imma fan of with holding info from the users, not true! 💃

chriskacerguis:
@serenamarie125 @Loicavenel thoughts on @AllenBW proposal?

Loicavenel:
Counting make a lot of sense for me. if we are sure the nofication is from the same "Source" then we should add a counting to it..

chriskacerguis:
@AllenBW So chatting with @Loicavenel and @serenamarie125 we have come to the agreement that this is the tip of the iceberg for a bigger issue (better event notice system).  So, we will need to take the suppression case-by-case.

For right now, let's suppress dups (and add the counter) for the following cases:
- Unable to connect to the backend
- Any JS errors (like what is pictured in your screenshot)

PR #770, opened 2017-05-16
PT#145434051-Adds documentation for creating functional areas in sui

AllenBW:
https://www.pivotaltracker.com/story/show/145434051

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -13,8 +13,6 @@ If any notional data is required for the test ensure an appropriately titled fol
 If you would like to run unit tests for the project, you can run   
 ```yarn test ```  
 
-
-
 If you would like to run the integration tests locally for the project please be sure to first have a running copy of the frontend and a running backend server.  You can either use a ManageIQ backend or test with our Mock API running locally.  These are the commands to launch the frontend and our mock api
   
 ``` yarn start ```  (Starts a frontend web server - Port 3001)
@@ -29,3 +27,6 @@ After those are running please ensure you have selenium server installed and run
 Now that your selenium server is running and listening open another terminal window and from the projects root directory, run   
  
 ```yarn test:e2e ``` (Kicks off tests)
+
+## How To...
+* [Add a Functional Zone](./howto.md#Addafunctionalzone)
--- a/docs/howto.md
+++ b/docs/howto.md
@@ -1,10 +1,21 @@
 # How to
 
 ## Add A Functional Zone
-[My Services](../client/app/services), [My Orders](../client/app/requests) and [Service Catalogs](../client/app/catalogs)
-are each unique functional zones of the Service User Interface (sui).  Each has a root resource explorer and a details
-page corresponding to each resource listed.  This guide will identify each aspect of a functional zone and explain the
-code patterns that make it work.
+The sum total of the following parts, a left navigation entry point, root/supporting states and root/supporting components
+are the core components that make up a functional zone.  As of writing, the Service UI, (SUI) contains four functional zones:
+* [Dashboard](../client/app/dashboard)
+* [My Services](../client/app/services)
+* [My Orders](../client/app/requests)
+* [Service Catalogs](../client/app/catalogs)
+
+Each functional zone has a root resource explorer and a details page corresponding to the listed resource.
+This guide will identify each aspect of functional zones and explain the code patterns that make it work.  As a developer
+once asked:
+
+> What is the UX term for the "bottom right half of the screen"
+
+It is more than a menu button, more than a navigation item, it is those things *plus* everything that makes what the user
+interacts with work.
 
 ### Left Navigation (lnav) Entry Point
 Each functional zone is accessed by clicking on the lnav entry point.
--- a/docs/howto.md
+++ b/docs/howto.md
@@ -1,0 +1,135 @@
+# How to
+
+## Add A Functional Zone
+[My Services](../client/app/services), [My Orders](../client/app/requests) and [Service Catalogs](../client/app/catalogs)
+are each unique functional zones of the Service User Interface (sui).  Each has a root resource explorer and a details
+page corresponding to each resource listed.  This guide will identify each aspect of a functional zone and explain the
+code patterns that make it work.
+
+### Left Navigation (lnav) Entry Point
+Each functional zone is accessed by clicking on the lnav entry point.
+
+#### Definition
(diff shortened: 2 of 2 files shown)


docs/howto.md
```diff
@@ -0,0 +1,59 @@
...
+code patterns that make it work.
+
+### Left Navigation (lnav) Entry Point
+Each functional zone is accessed by clicking on the lnav entry point.
+
+#### Definition
+The definition of the services functional zone lnav entry point occurs [here](../client/app/core/navigation.config.js#L12):
+```
+  const services = createItem({
+    title: N_('My Services'),
```
chriskacerguis: might want to make a note as to what the N_ means (and link to the docs)

AllenBW:
@adahms @serenamarie125 @chriskacerguis  @chalettu  Was hoping ya'll could give this a quick review at some point, documenting the process of adding a functional zone, ie, for the sui, a tab in the lnav and a buncha stuff that does stuff.  This is intended to be high level overview of operations, touching on all the code that needs to be touched on to make the thing happen, this is intentional.

chriskacerguis:
Ugh...Travis failed...gonna be one of those days I guess.

docs/contributing.md
```diff
@@ -13,8 +13,6 @@ If any notional data is required for the test ensure an appropriately titled fol
 If you would like to run unit tests for the project, you can run   
 ```yarn test ```  
 
-
-
 If you would like to run the integration tests locally for the project please be sure to first have a running copy of the frontend and a running backend server.  You can either use a ManageIQ backend or test with our Mock API running locally.  These are the commands to launch the frontend and our mock api
```
chriskacerguis: We should put a link to the Mock API repo

docs/howto.md
```diff
@@ -0,0 +1,232 @@
...
+  .factory('PowerOperations', PowerOperationsFactory)
+  .factory('ServicesState', ServicesStateFactory)
+  .factory('VmsService', VmsService)
+  .name;
+```
+When a new functional zone is created a new zone module will need to be included in the root [app.module.js](../client/app.module.js).
+The services.module.js is defined in this file as follows:
+```
+import { ServicesModule } from './services/services.module.js';
+```
```
chriskacerguis: Maybe something about how to do tests?

chriskacerguis:
@chalettu can you please review and add your thoughts?

@himdel @martinpovolny would you all like to review to add your thoughts?

/cc @chessbyte @dclarizio

serenamarie125:
@AllenBW I'm not sure what this is :)   Can you share what the requirements around this are?

docs/howto.md
AllenBW: Well that might be in a separate section, not immediately related to this topic...

AllenBW:
@serenamarie125 :point_up: [May 12, 2017 11:56 AM](https://gitter.im/ManageIQ/manageiq-ui-service?at=5915db1a8a05641b1176a2b1)

this conversation and a wee bit earlier, documenting the addition of a functional zone was requested

docs/howto.md
```diff
@@ -0,0 +1,232 @@
+# How to
+
+## Add A Functional Zone
```
serenamarie125: I would suggest changing the terminology from "Add a Functional Zone" to "Add an item to Primary Navigation" or "Add a Primary Navigation Item"

himdel:
Yeah, no, I'm going to ask for some real descriptions, other than "we chatted about that a while ago, around there".

Please use PR descriptions, it *is* a useful feature ;)

chriskacerguis:
@himdel this is documentation for how to add a section to the SUI

himdel:
So.. a Functional Zone = Menu section?

Now it even makes sense, thanks! :) So.. I'd suggest calling it menu sections :).

docs/howto.md
AllenBW: So I was thinking about that, but this extends just the scope of a simple Primary Navigation entry (part of a larger super set of functionality)

AllenBW:
So the menu section is a part of what we're discussing here, but not the whole picture, trying to encompass that other stuff in the name.

himdel:
@AllenBW Uh... OK, I guess that makes sense, but it really just brings me back to the original request .. can you please add a description of what a Functional Zone actually means and does?

AllenBW:
@himdel so first sentence of the file in question reads the following: 

> +[My Services](../client/app/services), [My Orders](../client/app/requests) and [Service Catalogs](../client/app/catalogs)
+are each unique functional zones of the Service User Interface (sui).  Each has a root resource explorer and a details
+page corresponding to each resource listed.  This guide will identify each aspect of a functional zone and explain the
+code patterns that make it work.

Totally down to add additional clarity!  Just unsure of what would be most comprehensible 🤔

himdel:
Sooo... a functional zone is just a menu section with its own explorer + details?

So.. just a specific kind of menu section, that's it?

AllenBW:
Hmm yesish?  But not so specific as only a a menu section with its own explore + details.  When I read "menu section" I think only of the navigational element, not the functionality activating said element conjures.  For the SUI most of the time we conjure explorers + detail views, but we aren't limited to that, in this document, the explorer + details paradigm is used only as an example.

himdel:
Aaah... so we're both thinking the same thing :) 

But.. given this discussion, I'm *sure* you can make the description more clear ;). 

I don't know, even reading something like "A functional zone is simply one whole menu section, including the code that implements it. Such a zone usually contains an explorer view and various detail screens but is not limited to such." would have helped me understand what this is about tremendously :).

AllenBW:
> I wonder if you should also consider documenting adding an item to secondary navigation as well ?

> Additionally, do you think that it's worth cross-linking to the PatternFly vertical navigation pattern? 

@serenamarie125 SUI doesn't use secondary nav anymore, might be distracting to add considering... Pattern is already in place, adding to the existing framework is the goal...
Will swap the word 'zone' for the word 'area'

AllenBW:
@serenamarie125 @chalettu , your pr approval OR DENIAL is desired if possible...

AllenBW:
@serenamarie125 Ah!! Ok, was thinking about this last night...  from a design perspective yep, wholly agree the focus should be on the entry point, adding a navigation or menu item is where the story would begin... from a code perspective the entry point is about six lines in some file compared to much more code in other files that are fundamentally essential for functionality (and in our case the code has, gotta follow some specific patterns to operate as expected)... this doc leans heavy in the direction of the developer audience.. less focused on what the user experiences....

PR #774, opened 2017-05-17
Login and page refresh not working

chalettu:
PV https://www.pivotaltracker.com/story/show/145635931


chalettu:
@miq-bot add_label bug
@miq-bot add_label fine/no

PR #784, opened 2017-05-19
Added tests for shopping cart service

chalettu:
PV https://www.pivotaltracker.com/story/show/145296897


chalettu:
@miq-bot add_label fine/no
@miq-bot add_label test

chalettu:
No UX changes. Just tests

chalettu:
@AllenBW if you wonder why I rolled back the version of the chai-as-promised package its because the newer version is written in es6 and causes our setup to tank when trying to run tests.

AllenBW:
@chalettu we gotta document that, not just in the pr message, but in a separate commit, something that'll show up when someone does git blame/ annotate

AllenBW:
@chalettu but ya know this also raises a good question... maybe we should have a running list of why some deps  are pegged at a version...

PR #792, opened 2017-05-25
Make utilization graphs on VM details page

chalettu:
PV https://www.pivotaltracker.com/story/show/140404667

[Finishes #140404667]


chalettu:
@miq-bot add_label fine/no
@miq-bot add_label my services / details
@miq-bot add_label ux/review

chalettu:
vm-details-graphs
Screenshot of implementation.

chalettu:
@serenamarie125 here is the PR that takes the same feature we had built into the Service Details page and implements it on the VM details page.   We didn't change any UX from the other screen but I attached what it looks like on the VM details page.

client/app/services/usage-graphs/usage-graphs.spec.js
```diff
@@ -56,4 +56,42 @@ describe('Usage Graph component', () => {
         expect(graphs[1].innerHTML).to.eq("6");
         expect(graphs[2].innerHTML).to.eq("38");
   })
+
```
AllenBW: not sure why eslint didn't pick up this space 🤔 

we might not check .spec.js files (probs not gonna guess)

client/app/services/vm-details/vm-details.component.js
```diff
@@ -65,21 +73,30 @@ function ComponentController($stateParams, VmsService, ServicesState, sprintf, l
...
-
+  function viewSelected(view) {
+    vm.viewType = view;
+  }
   function pollVM() {
     resolveData(true);
   }
   function resolveData(refresh) {
-    VmsService.getVm($stateParams.vmId, refresh).then(handleSuccess, handleFailure);
+    return VmsService.getVm($stateParams.vmId, refresh).then(handleSuccess, handleFailure);
```
AllenBW: why return this?

client/app/services/vms.service.js
```diff
@@ -91,9 +91,15 @@ export function VmsService(CollectionsApi, RBAC) {
         'tools_status',
         'uid_ems',
         'users',
+        'used_storage',
```
AllenBW: eeeeekkkkkkkkkk ya know what we're gonna need here, update to miq-mock-api

client/app/services/usage-graphs/usage-graphs.service.js
```diff
@@ -0,0 +1,79 @@
...
+      },
+      data: {
+        'used': usedValue,
+        'total': totalValue,
+      },
+      label: __('used'),
+    };
+  }
+
+  function getStorageChartConfig(used, total) {
```
chriskacerguis: Do we need a check here for a 0 total value so as to avoid a negative number? (same in the other methods too)

client/app/services/usage-graphs/usage-graphs.service.js
```diff
@@ -0,0 +1,79 @@
+/** @ngInject */
+export function UsageGraphsFactory() {
+  var service = {
+    cpuChartConfig: getCPUChartConfig,
+    memoryChartConfig: getMemoryChartConfig,
+    storageChartConfig: getStorageChartConfig,
+    convertBytestoGb: convertBytestoGb,
+  };
+
+  function getCPUChartConfig(used, total) {
```
chriskacerguis: Seems like we could maybe make 1 method, instead of 3...passing in the units and the chartId?  Thoughts?

client/app/services/usage-graphs/usage-graphs.service.js
```diff
@@ -0,0 +1,79 @@
...
+        chartId: 'storageChart',
+      },
+      data: {
+        'used': usedValue,
+        'total': totalValue,
+      },
+      label: __('used'),
+    };
+  }
+  function convertBytestoGb(bytes) {
```
AllenBW: There exists a filter for this... not sure if ya could rig it to work here...

client/app/services/usage-graphs/usage-graphs.service.js
```diff
@@ -0,0 +1,79 @@
...
+    memoryChartConfig: getMemoryChartConfig,
+    storageChartConfig: getStorageChartConfig,
+    convertBytestoGb: convertBytestoGb,
+  };
+
+  function getCPUChartConfig(used, total) {
+    let usedValue = 0;
+    let totalValue = 0;
+
+    if (angular.isDefined(used)) {
```
AllenBW: this exact logic is repeated three times, recommend abstracting it out into a private (to this service) function

client/app/services/usage-graphs/usage-graphs.service.js
```diff
@@ -0,0 +1,79 @@
...
+      usedValue = used;
+      totalValue = total;
+    }
+
+    return {
```
AllenBW: ditto the above... give it a few params, have it return an object, you'll be sitting pretty a mighty powerful itty bitty service

> his exact logic is repeated three times, recommend abstracting it out into a private (to this service) function

chriskacerguis:
@serenamarie125 for your review :)

client/app/services/vm-details/vm-details.component.js
chalettu: Helps make it a lot easier for unit testing.

client/app/services/usage-graphs/usage-graphs.service.js
chalettu: Looked into it and it wouldn't work being used here.

client/app/services/usage-graphs/usage-graphs.service.js
chalettu: Reworked this service to have a single function that takes care of building up the config.

chriskacerguis:
@AllenBW and @serenamarie125 could you all please review :)

client/app/services/vms.service.js
AllenBW: @chalettu we ever get an update to mock-api for this change?

chalettu:
V 1.3.2 of NPM package is the updated mock api

serenamarie125:
Looks great!  Couple questions @chalettu 

1. Is this a rollup of the utilization for all resources in the service?
2. Is CPU technically feasible?  If not, i'd pull it rather than show 'no data available'

chalettu:
@serenamarie125 .  This PR is on the Resource level.  
1.  This is not on the service level it is at the individual resource level
2.  CPU info is available but I am pretty sure it is only available for VM's and not so much on cloud instances.  We can remove it if you think we should.

serenamarie125:
@chalettu Thanks for the info.  @Loicavenel please chime in here as well! 

Regarding #1 - I think i totally misunderstood- so this is the Dashboard view of a VM which is a resource of a service ?!  If so this is AWESOME and just what we want 👯 

Regarding #2 - Base on the resource type, if CPU Memory & Storage are all available, display them all.  If the CPU is not available do not show that specific graph.  @Loicavenel do you agree?

client/app/services/vm-details/vm-details.html
```diff
@@ -12,8 +12,20 @@
         </li>
       </ol>
   </span>
+  <div class="view-selector-bar">
+    <div class="form-group toolbar-pf-view-selector">
```
AllenBW: will have to change for pf4

client/app/services/vm-details/vm-details.html
```diff
@@ -12,8 +12,20 @@
...
+      </button>
+      <button class="btn btn-link" ng-click="vm.viewSelected('detailsView')"
+              ng-class="{active: vm.viewType === 'detailsView'}">
+        <i class="fa fa-list-alt"></i>
+      </button>
+    </div>
+  </div>
 </div>
-<div ng-if="vm.listActions.length" pf-toolbar class="section-toolbar section-toolbar-right-actions"
+<div ng-if="vm.listActions.length && vm.viewType === 'detailsView'" pf-toolbar class="section-toolbar section-toolbar-right-actions"
```
AllenBW: gonna have to change for pf4

chalettu:
Hi @serenamarie125 .  I just got to check on this PR and wanted to see if you approve of it before we can complete it.   It currently works the way you had requested .  It will only show data graphs when data for that metric is present.

chriskacerguis:
@chalettu can you please address @AllenBW comments?

serenamarie125:
@chalettu in that GIF it shows "no data available" for CPU.  I'm suggesting in that case that it won't even display the CPU placeholder.  Can you just confirm that's the way it works, thanks!

chalettu:
@serenamarie125 I can make it work as you suggested but would you like to keep the spot where it would have been blank or do you want all charts to shift to the left.

client/app/services/vm-details/vm-details.html
chalettu: I looked at this and these are just CSS classes not patternfly components.  I don't believe we need to do anything in this case.

chriskacerguis:
@chalettu I would say "re-center" what charts you have.  @serenamarie125 thoughts?

serenamarie125:
imo, left aligned.  If you are on a big monitor i think it's better to have things top left aligned.  Asking @cshinn's opinion here as well ?

chalettu:
@serenamarie125 What do you want to happen if none of the graphs data is available?

chriskacerguis:
@chalettu display nothing

cshinn:
I agree that left-aligned is the way to go if one of them is missing. Is there any other content on that page if there are no graphs? 

Just to be clear, are we talking about a situation where we tried to request data and got back an error, or a situation where we will never be able to get data and we know that?

chalettu:
Yes, this is the situation where we requested the data and it was unavailable for that particular VM resource.

chalettu:
left_aligned
Here is what it looks like if one of the charts are empty.

chalettu:
@serenamarie125 we have implemented your requested feedback.  Do we have approval to go ahead and close out this PR?

serenamarie125:
Looks great!  In the event where there are no charts available, the dashboard view should even be available

chriskacerguis:
@serenamarie125 if you are good, could you please add your approval to the PR?

chriskacerguis:
@serenamarie125 That was done as a part of this PR :)

serenamarie125:
👍🏼😊

PR #793, opened 2017-05-25
Fix for service catalog service dialog refresh function behaving differently

eclarizio:
This is the SUI complement to [this PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/1349). More information can be found at that link.

@miq-bot add_label euwe/yes, fine/yes, bug
@miq-bot assign @chriskacerguis 

@AllenBW Can you review for me, or tag someone else to review? Thanks!

### DIFF
--- a/client/app/core/core.module.js
+++ b/client/app/core/core.module.js
@@ -23,6 +23,7 @@ import { BaseModalController } from './modal/base-modal-controller.js';
 import { BaseModalFactory } from './modal/base-modal.factory.js';
 import { ChargebackFactory } from './chargeback.service.js';
 import { CollectionsApiFactory } from './collections-api.factory.js';
+import { DialogFieldListenerFactory } from './dialog-field-listener.service.js';
 import { DialogFieldRefreshFactory } from './dialog-field-refresh.service.js';
 import { EventNotificationsFactory } from './event-notifications.service.js';
 import { ExceptionModule } from './exception/exception.module.js';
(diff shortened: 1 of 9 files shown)


AllenBW:
I'm lazy, gonna post the linked explanation  in here because I'm lazy
>The problem this PR solves stems from if users had something like a text area with information at the top that they expected to be able to be auto-refreshed every time they made a change to a lower dialog field. Cascading auto-refresh was intended to only trigger fields from top to bottom and only after the field that did the triggering.
This PR will treat every trigger as if it's triggering from the "0th" element, that is, if the very first dialog field is listening for an auto-refresh, it will be refreshed when any other element triggers an auto-refresh. The element that did the triggering will not trigger itself.
https://bugzilla.redhat.com/show_bug.cgi?id=1439761

AllenBW:
@chalettu would ya mind lending some 👁 on this? its been a while since I have euwed, wanta make sure nothing's fallin' through the cracks...

chriskacerguis:
@eclarizio we are going to have to back this out, it dropped the test coverage from 60% to 23%

simaishi:
@eclarizio 5 out of 9 files are conflicting - please create an Euwe PR for this.

chriskacerguis:
@simaishi we reverted this, so I'm thinking we should hold off on back porting it.

simaishi:
I thought reverting didn't happen? https://github.com/ManageIQ/manageiq-ui-service/issues/795 shows closed.

Anyway, please make sure to remove the `yes` labels if you want to hold off backporting for whatever reason.

chriskacerguis:
It didn't there...but, it happened here: https://github.com/ManageIQ/manageiq-ui-service/pull/799 (long story).  Anyway, will remove the labels.  Apologies for any confusion this caused.

simaishi:
Didn't notice the 2nd revert PR... thanks!

PR #794, opened 2017-05-26
BZ#145227511-Refactors order-explorer refresh row for clarity/conciseness

AllenBW:
https://www.pivotaltracker.com/story/show/145227511

And if we're really lucky CodeClimate will love what we've done with the place!

So ahhh the cc issues are a duplicate code issue... we can either increase or refactor code, this pr opts to refactor while maintaining same behavior as before.


AllenBW:
Looks like cc thinks 2 issues have been fixed :trollface:

PR #795, opened 2017-05-30
Revert "Merge pull request #793 from eclarizio/BZ1439761"

chriskacerguis:
This reverts commit f248880fce83dc2fa23571156271ecd740970d4f, reversing
changes made to 23c1ebbe6e9b1de5bca052702a2d5e5ca022132b.

Reason being is that the code coverage dropped from 60% to 23%


chessbyte:
@chriskacerguis I don't understand this PR.  You merged #793 where (per your comment) this was introduced.  Why not just work on fixing the drop in code coverage?

chriskacerguis:
@chessbyte I think the code is good. There is no possible way that about 200 lines of code can drop the coverage by about 40%. I'm not entirely sure what's going wrong but I think it's warrant investigating and I wanted to back this out to see what would happen.

chriskacerguis:
@fryguy @eclarizio @chalettu what do you all think?

@eclarizio sorry I was digging into code / Travis logs and didn't make that clearer

chriskacerguis:
Thinking about things since the code is good. I'm just going to close this PR and we can troubleshoot

eclarizio:
@chriskacerguis That's really odd that it dropped coverage, considering the fact that I *added* test cases in #793 to account for the refactoring I did. I don't know enough about the coverage code/reports to be able to understand why it would have dropped by so much, sorry 😢

chriskacerguis:
@eclarizio hey...it's all good.  Actually, we figured out the issue...wasn't you at all (well kinda).  Looks like CC reports on last run build (so 23% issue was on Euwe not master).  We fixed that here: https://github.com/ManageIQ/manageiq-ui-service/pull/796 SOOOOO...great job finding that.  Much awesome!

PR #797, opened 2017-05-31
Changed when we run code climate [Finishes #146306965]

chalettu:
PV https://www.pivotaltracker.com/story/show/146306965


chalettu:
@miq-bot add_label fine/no

PR #798, opened 2017-05-31
PT#14636890 - Update deps

AllenBW:
https://www.pivotaltracker.com/story/show/146368969

Nothing changes with this, cept a lil bitta pruning deps we don't use


chalettu:
I am going to request we halt this PR until we fix unit testing.  I need to review some of your proposed changes but do not remove any of the stuff around promises for unit testing.  I added those and tests will fail if you remove it.  The tests aren't failing right now because of that other issue with our tests not running.

chriskacerguis:
@AllenBW can you rebase this with the latest code that has the undo merge in it to make sure all is good?

AllenBW:
@chriskacerguis yep doing presently!

AllenBW:
So the karma-chai-as-promised dep can be updated, but we gotta reconfigure the way webpack runs tests, something for a future update, not today, that aside, @chalettu we good to go here?

PR #812, opened 2017-06-07
PT#146772249 - Error loading services without vms

AllenBW:
https://www.pivotaltracker.com/story/show/146772249

I checked, https://github.com/ManageIQ/manageiq-ui-service/pull/792 doesn't address this block of code

## Before
service

## After
service-fixed


chriskacerguis:
No UX changes.

PR #815, opened 2017-06-09
Adds icon-status component

AllenBW:
https://www.pivotaltracker.com/story/show/146029083 

**TL;DR** 
- adds a new reusable component called `icon-status`, give it a status and array of possible values for different statuses, and you got ICONS 👶 
- employs aforementioned component in ansible details component

Takes 5 parameters ; and the [icon class name pulled from here](https://www.patternfly.org/styles/icons/)
- status - string of the current status
- success - array of strings for which to display the success icon; `pficon pficon-ok`
- error - array of strings for which to display the error icon; `pficon pficon-error-circle-o`
- queued - array of strings for which to display the circle clock icon; `fa fa-clock-o`
- inprogress - array of strings for which to display the in progress spinner; `spinner spinner-xs spinner-inline`

If none of the following inputs match the status, the unknown icon will be displayed; `pficon pficon-help`

## Examples of icons used
screen shot 2017-06-12 at 12 34 38 pm

### The unknown catchall
screen shot 2017-06-12 at 12 41 44 pm

### String values are still shown to user
screen shot 2017-06-12 at 12 49 41 pm

### DIFF
--- a/client/app/services/service-details/service-details-ansible.html
+++ b/client/app/services/service-details/service-details-ansible.html
@@ -12,8 +12,8 @@
                   <div class="form-group">
                     <label class="control-label col-sm-3" translate>Status</label>
                     <div class="col-sm-8">
-                      <input class="form-control text-capitalize" readonly
-                             value="{{item.stack.status || ('Unknown' | translate)}}"/>
+                      <icon-status status="item.stack.status" success="['successful', 'succeeded']"
+                                   error="['failure', 'failed']"></icon-status>
                     </div>
                   </div>
                   <div class="form-group">
@@ -94,7 +94,7 @@
               </span>
                   </div>
                   <pf-list-view ng-if="item.credentials.length > 0" config="vm.credListConfig"
-                       items="item.credentials" custom-scope="vm">
+                                items="item.credentials" custom-scope="vm">
                     <div class="row">
                       <div class="col-lg-3 col-md-3 col-sm-3 col-xs-3">
                         <div class="list-view-stacked-item">
@@ -144,7 +144,7 @@
                 </span>
                   </div>
                   <pf-list-view ng-if="item.jobs.length > 0" config="vm.playsListConfig"
-                       items="item.jobs" custom-scope="vm">
+                                items="item.jobs" custom-scope="vm">
                     <div class="row">
                       <div class="col-lg-3 col-md-3 col-sm-3 col-xs-3">
                         <div class="list-view-stacked-item">
@@ -157,9 +157,8 @@
                       <div class="col-lg-2 col-md-2 col-sm-2 col-xs-2">
                         <div class="list-view-stacked-item">
                           <strong translate>Status</strong>
-                          <div class="text-capitalize">
-                            {{ item.resource_status }}
-                          </div>
+                          <icon-status status="item.stack.status" success="['successful', 'succeeded']"
+                                       error="['failure', 'failed']"></icon-status>
                         </div>
                       </div>
                       <div class="col-lg-2 col-md-2 col-sm-2 col-xs-2">
--- a/client/app/services/service-details/service-details-ansible.html
+++ b/client/app/services/service-details/service-details-ansible.html
@@ -12,7 +12,7 @@
                   <div class="form-group">
                     <label class="control-label col-sm-3" translate>Status</label>
                     <div class="col-sm-8">
-                      <icon-status status="item.stack.status" success="['successful', 'succeeded']"
+                      <icon-status status="item.stack.status" success="['successful', 'succeeded', 'create_complete']"
                                    error="['failure', 'failed']"></icon-status>
                     </div>
                   </div>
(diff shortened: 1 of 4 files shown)


Loicavenel:
@AllenBW any possible screenshots here? I would like to see it

AllenBW:
@Loicavenel yeah sure when we're outta **[wip]** there will be all da goodies :+1:

AllenBW:
@Loicavenel outta wip! GO GO GO GO GO 💃

chriskacerguis:
A few thoughts...

1. The clock icon appears to be a different size than the PF icons.
2. The alignment of the icons vs. the "Status" label seem "off"

@serenamarie125 thoughts on those?

(see attached screenshot)


27044500-beee8df8-4f6b-11e7-9af9-d3f0f944584e

client/app/services/service-details/service-details-ansible.html
```diff
@@ -12,8 +12,8 @@ <h3 translate>Results</h3>
                   <div class="form-group">
                     <label class="control-label col-sm-3" translate>Status</label>
                     <div class="col-sm-8">
-                      <input class="form-control text-capitalize" readonly
-                             value="{{item.stack.status || ('Unknown' | translate)}}"/>
+                      <icon-status status="item.stack.status" success="['successful', 'succeeded', 'create_complete']"
```
chriskacerguis: Do we want to define the stings here?  That could become a VERY long list.

AllenBW: Presently not a long list, hence the deliberate inclusion for this first run. If it gets much longer tho you're spot on, const in the js is gonna be da way to play

AllenBW:
@chriskacerguis RIGHT!! icon alignment yo, used all things pitched by pf tho so.... idk :-/ i guess we could add custom class to get em all aligned, but that seems like the bandaid for the wrong problem...

chriskacerguis:
@ManageIQ/core-ux for ya'lls review :)

cshinn:
@AllenBW I think there may be a mixup regarding the icons used for `queued` and `inprogress`; the descriptions don't match the provided icon classes. 

Is it possible for an item to have two different status icons at the same? If not, then I don't think the size issue is severe enough to warrant specific code to fix it. I do agree, however, that the size discrepancy in the icons is unfortunate. Your thoughts @serenamarie125?

AllenBW:
good catch, the mixup was pr description deep, eyyyyooo @cshinn swapped them around 🙇‍♀️ 

its technically possible for an item to have more than one icon at a time, this is by design, but one would have to deliberately desire this behavior (and code towards it) most of the time and in this pr, one icon per status.  does not affect any other items in page, just looks a lil off when ya line em all up 😆 

@Loicavenel icons can be used anywhere, application of this pr is an ansible service.

serenamarie125:
happened again.  I swear i must not hit "COMMENT" after commenting.  
THANK YOU @AllenBW for doing this - this is great and will ensure more consistency.  Would be assume if this could be shared with Ops UI too :D 

Couple of comments 
1. Regarding size, this happens often with FA icons.  Often we will end up duplicating them in PF to be sure size/spacing is appropriate.  Let's see if we stick with the clock, because i don't think it's right.
2. success - perfect, this is PF approved ✅ 
3. error - perfect, this is PF approved ✅ 
4. queued - array of strings for which to display the circle clock icon; fa fa-clock-o❌ let me get back to you be COB
5. inprogress - array of strings for which to display the in progress spinner; spinner spinner-xs spinner-inline ❓ let me get back to you by COB
6. unknown icon will be displayed; pficon pficon-help ❌ let me get back to you by COB

HTH!

chriskacerguis:
@serenamarie125 any update for us?

/cc @Loicavenel

serenamarie125:
@chriskacerguis the current update is that PatternFly does not have appropriate icons at this point.  IMO, we have 2 options
1 - use temp icons and open up another issue when icons are available
2 - wait for PF icons to be created ( could be 2+ weeks )

AllenBW:
@serenamarie125 i vote temp! :P

chriskacerguis:
@AllenBW temp will work...I think it would be a better UX with PF icons...@serenamarie125 would you be able to give a more definitive timeline?

serenamarie125:
Based on the fact that PF does not currently have standards for queued, in progress & unknown, I'd like to suggest one of 2 things:
1 - wait for 2-3 weeks for PF to create these icons
2 - merge as is, but update the icons as soon as PF creates them

If #2, i think we should pick a different icon for queued, because i don't think that one makes sense.  Let me know your thoughts @chriskacerguis @Loicavenel

serenamarie125:
@AllenBW can you do me a favor and explain what the Queued status reflects in this instance?  Is it the same as pending ?  my concern with the clock icon is that users may confuse it with "scheduled"

serenamarie125:
@chriskacerguis is 2-3 weeks enough of a timeline?  It will be added to our sprint planning, and our 2 week sprint will begin on 6/27 so I anticipate it being completed that sprint.

chriskacerguis:
@serenamarie125 that works great, I think that it would be a better experience to have matching icons.

AllenBW:
@serenamarie125 updated! 🎖 ⚓️

client/app/shared/icon-status.component.js
```diff
@@ -0,0 +1,50 @@
...
+  controller: ComponentController,
+  bindings: {
+    status: '<',
+    success: '<?',
+    error: '<?',
+    queued: '<?',
+    inprogress: '<?',
+  },
+  template: `
+   <i class="pficon pficon-ok" ng-if="vm.isSuccess"
```
AllenBW: Intentionally verbose, favoring readability rather than conciseness, as was demonstrated above, likely these values are going to change

One below, don't have to worry about translations breaking functionality value and check value(s) are passed by the user, most often specified by the backend for both

PR #820, opened 2017-06-15
PT#146810903 Adds and wires pf-table deps

AllenBW:
https://www.pivotaltracker.com/story/show/147228327

Also fixes a bug where in selected results are not updated when pagination affects list views

No visual changes, just picking the correct deps for this functionality


chriskacerguis:
Been things about this...and just thinking out loud here.  Do we want to inject something that we know is not complete and that it may be awhile before we get it in?

AllenBW:
> inject something 

well we're not really injecting anything here... no more or less offensive than leaving  unused deps hanging out there (first commit fixes this).. except at least we'll be using this at some point in the future

PR #823, opened 2017-06-19
Adds "View Graphs" link to service details

AllenBW:
https://www.pivotaltracker.com/story/show/146157721

Transitions user to vms-details state with a defalut viewType specified
At the moment, graphs component does not appear to be working correctly, can't handle being the default view, following up with pr to fix

screen shot 2017-06-19 at 10 30 37 am

### gif of the bug i'll be working on next
graph-issue

### DIFF
--- a/client/app/services/service-details/service-details.html
+++ b/client/app/services/service-details/service-details.html
@@ -337,6 +337,10 @@
                           <a ng-click="$ctrl.customScope.gotoComputeResource(item)" translate>View
                             Details</a>
                         </li>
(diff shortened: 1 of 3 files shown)


Loicavenel:
@serenamarie125 @AllenBW Should we really call this "View Graph"?? and not "Utilisation" like in Ops UI... Graph is really generic

AllenBW:
@Loicavenel can call it whateva yo ❤️ desires, just 🗣 the 🔠

PR #824, opened 2017-06-19
[WIP] Allow clientside validation of dialog fields [Finishes #146312755]

chalettu:
PV https://www.pivotaltracker.com/story/show/146312755

### DIFF
--- a/client/app/core/dialog-utilities.service.js
+++ b/client/app/core/dialog-utilities.service.js
@@ -1,0 +1,39 @@
+export function DialogUtilitiesFactory() {
+  const service = {
+    validateField: validateField,
+    validateAllFields: validateAllFields,
+  };
+
+  return service;
+
+  function validateField(field) {
+    const fieldValue = field.default_value;
+    if (field.required) {
+      if (fieldValue === '') {
+        field.fieldValidation = false;
+        field.errorMessage = __('field is required');
+
+        return false;
+      }
+      if (field.validator_type === 'regex') {
+        const regex = new RegExp(`${field.validator_rule}`);
+        const regexValidates = regex.test(fieldValue);
+        field.fieldValidation = regexValidates;
+        field.errorMessage = __("field doesn't match required format");
+
+        return regexValidates;
+      }
+    }
+
+    return true;
+  }
+
+  function validateAllFields(fields) {
+    let fieldsValidate = false;
+    fieldsValidate = fields.every(function(field) {
+      return validateField(field);
+    });
+
+    return fieldsValidate;
+  }
+}
--- a/client/app/core/dialog-utilities.service.spec.js
+++ b/client/app/core/dialog-utilities.service.spec.js
@@ -1,0 +1,40 @@
+describe('Dialog Utilities - ', () => {
+    let requiredField = {};
+    let regexField = {};
+  beforeEach(() => {
+      requiredField = {
+          default_value: 'blah',
+          required: true
+      };
+      regexField = {
+          default_value: '1',
+          required: true,
+          validator_type: 'regex',
+          validator_rule: '[0-9]'
+      };
+    module('app.core');
+    bard.inject('DialogUtilities');
+  });
+
+  it('should validate a single field successfully', () => {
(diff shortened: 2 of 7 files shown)


chalettu:
@miq-bot add_label fine/no
@miq-bot add_label catalogs

chalettu:
@miq-bot add_label ux/review

chalettu:
Screen shots of what the clientside validation errors look like.  
regex_field

chalettu:
This screenshot shows validation on submit of a form.  
addtocard

Loicavenel:
This is very nice. the only missing information is why is not valid but this will come from the other requirements where we can put an information that will provide some information when clicking on it or mouse over it.

chalettu:
@Loicavenel , I created two messages for errors , "field doesn't match required format" and "field is required".   This goes back to our discussion last week with the greater issue of how do we deal with translating error messages.  I chose the "field doesn't match required format" message as a default for regex errors since we don't currently support any type of custom error messages at the api level and I couldn't think of something to really clearly indicate, "User got the regex wrong".

chriskacerguis:
@ManageIQ/core-ux for your review :)

/cc @Loicavenel

client/app/shared/dialog-content/dialog-content.component.js
```diff
@@ -52,8 +56,11 @@ function DialogContentController(API_BASE, lodash) {
     if (Object.keys(vm.parsedOptions).length > 0) {
       item.dialog_fields.forEach(iterateDialogFields);
     }
+    item.dialog_fields.forEach(setFieldDefaults);
```
AllenBW: 59 to 62 could be simplified with .map and =>

```
item.dialog_fields.map((item)=>{return item.fieldValidation =null});
```

client/app/states/catalogs/details/details.state.js
```diff
@@ -108,7 +108,10 @@ function Controller(dialogs, serviceTemplate, EventNotifications, DialogFieldRef
     if (!ShoppingCart.allowed()) {
       return;
     }
-
+    const allFieldsValidate = DialogUtilities.validateAllFields(allDialogFields);
+    if (!allFieldsValidate) {
```
AllenBW: consider combining the if from line 108 with this one...

```
if (!ShoppingCart.allowed() && !allFieldsValidate){
return;
}
```

client/app/shared/dialog-content/dialog-content.component.js
AllenBW: consider using .map and => for 59-62

```
item.dialog_fields.map((item)=>{item.fieldValidation = null;});
```

client/assets/sass/_forms.sass
```diff
@@ -1,3 +1,4 @@
+@import 'colors'
```
AllenBW: this should be 'app_colors' (which is the only file that imports colors directly)

client/app/core/dialog-utilities.service.js
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+  return service;
+
+  function validateField(field) {
+    const fieldValue = field.default_value;
+    if (field.required) {
+      if (fieldValue === '') {
+        field.fieldValidation = false;
+        field.errorMessage = __('field is required');
```
AllenBW: Leading with lowercase doesn't conform to patternfly validation example presented [here](http://www.patternfly.org/pattern-library/forms-and-controls/errors-and-validation/)

client/app/core/dialog-utilities.service.spec.js
```diff
@@ -0,0 +1,40 @@
+describe('Dialog Utilities - ', () => {
```
AllenBW: Test name might be incomplete?  Unsure as to the meaning of the - at the end...

chriskacerguis:
@chalettu we need to add a better "description" so people know what is wrong

chriskacerguis:
Closing this, as Chris has stated that this will be rolled into the fix with the dialog merge to the ui-components.

PR #834, opened 2017-07-06
[FINE] Ensures skip-token-header is correctly formed

AllenBW:
https://bugzilla.redhat.com/show_bug.cgi?id=1468000
https://www.pivotaltracker.com/story/show/148415143

This ensures sui times out, but at the moment, session timeout happens ~4 ~2 minutes after ops... thats the wip part of this pr, working on figuring out why...

updates lock file because apparently lock file needs to be updated

### DIFF
--- a/client/app/core/session.service.js
+++ b/client/app/core/session.service.js
@@ -26,7 +26,7 @@ export function SessionFactory($http, $q, $sessionStorage, $window, $state, $coo
   function setAuthToken(token) {
     model.token = token;
     $http.defaults.headers.common['X-Auth-Token'] = model.token;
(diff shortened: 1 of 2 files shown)


AllenBW:
@chriskacerguis @simaishi un-wipping this as not much headway is being made into discovering factors impacting irregular session timeouts in sui (this pr brings us inline with the what one will observe watching network tab on master)

AllenBW:
OOOO but wait... i might be on to something.... back into wip

AllenBW:
So I was thinking this might be in play, https://stackoverflow.com/questions/44060131/zone-js-violation-warnings-on-console-in-angular-project-only-on-chrome#44339214

but behavior between master and fine is identical so maybe not :-/

PR #804, opened 2017-06-01
PT#139486261 - Update angular patternfly

AllenBW:
https://www.pivotaltracker.com/story/show/139486261

### Major work performed
- Switch all attribute directives to component forms, i.e `<div pf-list-view>` to `<pf-list-view>`
- Find and replace usage of `pf-datepicker` which was also removed in angular-patternfly 4
- Upgrade application code to use lodash 4
- rework pf-select
- replace customScope with $ctrl.customScope
- correct form implementations, they're off in places
- Audit usage of each converted component, in a short time of testing found a few issues with pf-list-view customScope not working as expected

### GIF of exploring each of the changes (I tried to retire something that isn't retierable, excuse this thing that looks like a bug, not related to the pf4 jump)
ang-pf4

PR #836, opened 2017-07-07
[FINE] Restores catalog scroll bar, no change to explorers

AllenBW:
https://www.pivotaltracker.com/story/show/148488889
https://bugzilla.redhat.com/show_bug.cgi?id=1466049

Removes extraneous margin, pf list view now displays in viewport (on top of other elements)

## PROOF
catalognostroll

### DIFF
--- a/client/assets/sass/_explorer.sass
+++ b/client/assets/sass/_explorer.sass
@@ -8,7 +8,6 @@
 
   .list-view-pf
     height: calc(100vh - 237px)
-    overflow-y: auto
 
   .explorer-children-list
     .list-view-pf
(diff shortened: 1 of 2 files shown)


chriskacerguis:
@AllenBW you have this as [FINE] with fine/yes.  Should it be master with "fine/yes"?  

[FINE] (on the fine branch) implies fine/yes so I just want to make sure

AllenBW:
oooooo this only appears to plague fine, is set to go as is

is fine/yes only from master to fine? any flags for fine specific prs?

simaishi:
@AllenBW No extra label is needed when a PR is going directly to the branch. Do you know which PR fixed this in master?

AllenBW:
@simaishi oh gosh, no i don't offhand, it'll take a while to find it, can probably get ya words closer to that on monday

AllenBW:
@simaishi thinking this work could be the fix on master https://github.com/ManageIQ/manageiq-ui-service/pull/804

PR #821, opened 2017-06-15
Pt/#146312989 typeahead dialog select

AllenBW:
https://www.pivotaltracker.com/story/show/146312989

@serenamarie125 @Loicavenel @chriskacerguis 

### Heres the action shot! (shots?)
dialog-select

🌮 💃

PR #840, opened 2017-07-11
BZ#145848583 - Introduce service tag filtering

AllenBW:
**Update, this pr is done, but there are two api bugs that I would like to be sorted prior to merging this in (detailed below)**

~~Why are we wipping? present issues (and prs solving when they are up)~~ All of the following blockers are resolved:
1. ~~polling results in a call back sorta effect whenever filter or sorting is applied, **need** to rework this immediately~~ 
**We stop polling whenever a new request for resources has been made, ie a filter or sorting event (or the component is destroyed)
After successfully grabbing these new resources, we begin polling on the successful call
This avoids the previous issue of stale polling**
2.  ~~api anding of multiple filters does not quite work as expected, issue has been identified and is being looked at (not an sui issue)~~
**as was pointed out in the bz, this is actually expected behavior, we are looking for services with only the tags added not including the tags**
https://bugzilla.redhat.com/show_bug.cgi?id=1470251
3.  ~~api throws error when changing sort direction of filtered tags~~
https://bugzilla.redhat.com/show_bug.cgi?id=1470260
https://github.com/ManageIQ/manageiq/pull/15559

### No more polling artifact <3 <3 <3
filterbytags

### What filtering by tag looks like, also what this polling artifact bug looks like :(
filterbytags

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -158,7 +158,7 @@
     "no-bitwise": "error",
     "no-caller": "error",
     "no-catch-shadow": "error",
-    "no-confusing-arrow": "error",
+    "no-confusing-arrow": ["error", {"allowParens": true}],
     "no-continue": "off",
     "no-div-regex": "error",
     "no-duplicate-imports": "error",
--- a/client/app/services/service-explorer/service-explorer.component.js
+++ b/client/app/services/service-explorer/service-explorer.component.js
@@ -11,7 +11,7 @@ export const ServiceExplorerComponent = {
 /** @ngInject */
 function ComponentController($state, ServicesState, Language, ListView, Chargeback, TaggingService, TagEditorModal,
                              EventNotifications, ModalService, PowerOperations, lodash, Polling, POLLING_INTERVAL) {
-  var vm = this;
+  const vm = this;
 
   vm.$onDestroy = function() {
     Polling.stop('serviceListPolling');
@@ -19,14 +19,13 @@ function ComponentController($state, ServicesState, Language, ListView, Chargeba
 
   vm.$onInit = () => {
     vm.permissions = ServicesState.getPermissions();
-    if ($state.params.filter) {
-      ServicesState.services.setFilters($state.params.filter);
-      ServicesState.services.filterApplied = true;
-    } else {
-      ServicesState.services.setFilters([]);
-      ServicesState.services.filterApplied = false;
-    }
+    $state.params.filter ? ServicesState.services.setFilters($state.params.filter) : ServicesState.services.setFilters([]);
     ServicesState.services.setSort({id: "created_at", title: "Created", sortType: "numeric"}, false);
+
+    TaggingService.queryAvailableTags().then((response) => {
+      vm.filterTags = response;
+    });
+
     angular.extend(vm, {
       loading: false,
       title: __('Services'),
@@ -53,14 +52,20 @@ function ComponentController($state, ServicesState, Language, ListView, Chargeba
       cardConfig: getCardConfig(),
       listConfig: getListConfig(),
       listActions: getListActions(),
-      headerConfig: getHeaderConfig(),
+      toolbarConfig: {
+        sortConfig: serviceSortConfig(),
+        filterConfig: serviceFilterConfig(),
+        actionsConfig: {
+          actionsInclude: true,
+        },
+      },
       menuActions: getMenuActions(),
       serviceChildrenListConfig: createServiceChildrenListConfig(),
       pollingInterval: POLLING_INTERVAL,
     });
     vm.offset = 0;
 
-    Language.fixState(ServicesState.services, vm.headerConfig);
+    Language.fixState(ServicesState.services, vm.toolbarConfig);
 
     resolveServices(vm.limit, 0);
     Polling.start('serviceListPolling', pollUpdateServicesList, vm.pollingInterval);
@@ -86,7 +91,7 @@ function ComponentController($state, ServicesState, Language, ListView, Chargeba
 
   function handleSelectionChange() {
     vm.selectedItemsList = vm.servicesList.filter((service) => service.selected);
-    vm.headerConfig.filterConfig.selectedCount = vm.selectedItemsList.length;
+    vm.toolbarConfig.filterConfig.selectedCount = vm.selectedItemsList.length;
   }
 
   function isAnsibleService(service) {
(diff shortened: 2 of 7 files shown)


AllenBW:
@chriskacerguis This pr is done, but there are two api bugs that I would like to be sorted prior to merging this in (detailed below)

chriskacerguis:
@allenbw Can you please make BZs for those issues and put the links in the PR comments?

AllenBW:
@chriskacerguis yeah! totally did, travis passed, but its not being reported, any idea whats up?

chriskacerguis:
Looks like it passed: https://travis-ci.org/ManageIQ/manageiq-ui-service/builds/252562261

Do you want me to merge or wait for those bugs to pass?

AllenBW:
@chriskacerguis thats a gooooooooooood question.... i know one is on dev presently, the other should be soonish... without them filtering by tags still works... its just breakable :P

As there is no more work to do here and bz are made... i think merging is ok to do

chriskacerguis:
@ManageIQ/core-ux for your review.

/cc @Loicavenel

AllenBW:
I anticipate one feedback might be "ew, those tags are long and silly" buttttttt this the pf-ang lib doesn't support an tag value alias, there might be another way to manage this, i have an idea, working it, but the need for the ability to display a tag alias (rather than the search value) has also been expressed here https://bugzilla.redhat.com/show_bug.cgi?id=1470868

AllenBW:
Yeah totally possible, for filter thing to support alias! Thinking of doing that work in separate pr, aligned to the ref... Thoughts ?

chriskacerguis:
@AllenBW since we are still waiting for UX review, I would say let's just add it to this one.

Loicavenel:
I will leave Serena reviewing this one. But, I see 2 issues:
- this list can really long..
- "Managed/owner/prod_ops"  is not really friendly. Each Category has a description, and each value has a description, do we retrieve it?  It will be a much better experience to have the description including in when we display Tags... If an API has to be updates, let's do it.
I think we should also offer a menu in 2 phases where user see first the categories and then when he is selecting a category, UI list all tags in the category...

serenamarie125:
@AllenBW definitely think that this is a useful capability.  Colleen and I worked on this design, and would like this to be implemented:

filterbytag

Upon selecting to filter by tag [1], then there will be two components added:
Filter by Category... [2] and Filter by Value... [3]

Both of those components should use the typeahead dialog select, as you did here:  https://github.com/ManageIQ/manageiq-ui-service/pull/821

When the value [3] is designated, a filter chip [4] should be added in the format:
Tag: <category>: <value>

Does this seem achievable?  Let me know your thoughts @AllenBW @Loicavenel !

Loicavenel:
@serenamarie125 looks great... I just want to look if we can get the Description and not the value.
Attached some screenshots showing what we see when we select Tags in VM
screen shot 2017-07-14 at 22 26 18
screen shot 2017-07-14 at 22 26 06

AllenBW:
@serenamarie125 @Loicavenel not impossible (would totally love to as its far more usable), but angular-patternfly presently does not have provisions to support a multi-dropdown filter (what we have to work with presently: http://www.patternfly.org/angular-patternfly/#/api/patternfly.filters.component:pfFilter)

Thoughts on how to proceed?

AllenBW:
Also, I lied above, alias isn't possible with ang-pf, a mild version is possible with some string hackery, but this involves deconstructing valid filter values from api, displaying them, then reconstructing them, undesired. Ideally, would prefer to have an field [filterAlias or something here](https://github.com/patternfly/angular-patternfly/blob/master/src/filters/filter-fields-component.js) heck even an ignored param that could be passed around filter implementation would be handy

serenamarie125:
@jeff-phillips-18 just tagged you in the angular-patternfly repo.  You can see the desired design above ... I'm thinking you may have done something similar to this in the past.  If not, I think this would be great to have supported in PF as a generic option, would you agree?  @AllenBW of course contributing to PF is always a great option :)

jeff-phillips-18:
@serenamarie125 yeah, I don't recall doing this in the past.

AllenBW:
Thanks for the supa fast reply @jeff-phillips-18 ❤️ 🙇‍♀️ (I'm not confused, its the closet reaction github provides to a sad face)

@serenamarie125 given that this isn't present functionality and will require a nontrivial amounta effort to get in, you down with adding the present functionality as is with the pt to circle back around and "fix" when the desired look/feel exists?

chriskacerguis:
@serenamarie125 would @AllenBW suggestion (https://github.com/ManageIQ/manageiq-ui-service/pull/840#issuecomment-316056592) be ok with you?

/cc @Loicavenel

serenamarie125:
@AllenBW I think there are usability issues with the current PR, and don't think that it should be added.  Although PF doesn't currently support it, I would think that you could either enhance the component and contribute it back, or just tailor in the SUI and contribute back at a later date?

Loicavenel:
+1

AllenBW:
@serenamarie125 if we're to continue to use patternfly, this component would have to be added to patternfly, not something we can tailor in the sui, I'll keep this pr out here and mark the pt as blocked

serenamarie125:
@AllenBW i don't think this should block SUI implementing a usable solution.  PatternFly is an OpenSource community, please feel free to contribute what you need to Angular PatternFly.

AllenBW:
@serenamarie125 it was my understanding that this pr couldn't go through until it met the mock above?  

that can't happen until the additions are made to the repo, that we have to hold off on this pr until those changes are present means this pr is blocked 👍 🌮 😿

Loicavenel:
@AllenBW @chriskacerguis If this new "UI fonction" is not available today, why can you make it? 
Why you guys cannot add it to PartternFly community? then everybody can benefits from it..

AllenBW:
Woaaaaaah @Loicavenel I never said we *couldn't* add it to patternfly, apologies if I wasn't clear, the functionality is not in ang-pf at this very moment, which means this pr is blocked 😺

Loicavenel:
@AllenBW ok, so we can move on this PR with doing it in Partnerfly or doing it our own way close to Partnerfly and someone will move it to partnerfly later

AllenBW:
@Loicavenel sounds good to me!

serenamarie125:
@chriskacerguis how did this get merged?  Both @Loicavenel and I requested a change

chriskacerguis:
Whoops. My bad. You did. I'll back it out

serenamarie125:
Thanks @chriskacerguis

PR #841, opened 2017-07-12
removed node 7, added node 8

chriskacerguis:
removed node 7, added node 8 from travis testing


AllenBW:
Not touching 6 because appliance still supports it?  @chriskacerguis

chriskacerguis:
@AllenBW correct.  I think it is best to always test LTS (which is what the appliance will run) and the latest Node version.

Issue #850, opened 2017-07-22
Include icons provided by font-fabulous


skateman:
@epwinchell created a custom fonticon set for our needs. As we want to support [fonticons on custom buttons](https://github.com/ManageIQ/manageiq-ui-service/pull/851), we need this iconset to be included in the SUI.

You can find the repo here: https://github.com/ManageIQ/font-fabulous
There is (untested) support for npm/yarn, but we haven't pushed any versions yet.
If you need a version or some extra modifications in the `package.json`, let me know.

Thanks

AllenBW:
Closed by#851 and #854

PR #851, opened 2017-07-22
Display colored fonticons before custom buttons

skateman:
Pivotal story in the OPS UI: https://www.pivotaltracker.com/story/show/147779325
Depends on: https://github.com/ManageIQ/manageiq-schema/pull/39
Some icons might be not displayed becauseof this: https://github.com/ManageIQ/manageiq-ui-service/issues/850

screenshot from 2017-07-22 11-43-45

### DIFF
--- a/client/app/services/custom-button/custom-button.html
+++ b/client/app/services/custom-button/custom-button.html
@@ -4,6 +4,7 @@
           ng-click="vm.invokeCustomAction(button)"
           uib-tooltip="{{button.description}}"
           tooltip-placement="top">
+    <i ng-class="button.options.button_icon" ng-style="{color: button.options.button_color }"></i>
     {{ button.name }}
   </button>
 </span>
@@ -14,6 +15,7 @@
             uib-dropdown-toggle
             uib-tooltip="{{buttonGroup.description}}"
             tooltip-placement="top">
+      <i ng-class="buttonGroup.set_data.button_icon" ng-style="{color: buttonGroup.set_data.button_color }"></i>
       {{ buttonGroup.name.split('|')[0] }}
       <span class="caret"></span>
     </button>
@@ -26,6 +28,7 @@
            ng-click="vm.invokeCustomAction(button)"
            uib-tooltip="{{button.description}}"
            tooltip-placement="left">
+          <i ng-class="button.options.button_icon" ng-style="{color: button.options.button_color }"></i>
           {{ button.name }}
         </a>
       </li>


serenamarie125:
@chriskacerguis i think should should have a ux/review tag on it along with a UX review?

chriskacerguis:
@serenamarie125 /agree :)

chriskacerguis:
@serenamarie125 I've added the label, but it looks like this is still WIP.

skateman:
@chriskacerguis I market it as WIP because of #850

chriskacerguis:
@skateman ahhh...thanks for clarifying.

serenamarie125:
@chriskacerguis understood, but having visibility to what's being worked on is helpful, thank you!

chriskacerguis:
@serenamarie125 makes sense to me :)

client/app/services/custom-button/custom-button.html
```diff
@@ -4,6 +4,7 @@
           ng-click="vm.invokeCustomAction(button)"
           uib-tooltip="{{button.description}}"
           tooltip-placement="top">
+    <i ng-class="button.options.button_icon" ng-style="{color: button.options.button_color }"></i>&nbsp;
```
AllenBW: Skeptical about use of `&nbsp;` why is this necessary?

skateman: @AllenBW I can add some padding-right instead

AllenBW: yes please!! @skateman 🙇

chalettu: So has the API been updated to send the icons in the custom button API response?

skateman:
Rebased, ready to merge!

chriskacerguis:
@serenamarie125 ready for review :)

serenamarie125:
What am i supposed to be reviewing here :)  I haven't seen any icon picker, is that it?  I see one image with a green star, but 👀 unsure

skateman:
@serenamarie125 that's the thing :wink: we're just displaying the icons in the SUI and the [picker](https://github.com/ManageIQ/ui-components/pull/78) will be used in the OPS UI only.

chriskacerguis:
@serenamarie125 echoing the same thing that @skateman said, since this is just the SUI...it is just showing the icons...not picking them.  I think that the image is showing is what the button with an icon / color will look like in the SUI.

PR #854, opened 2017-07-27
PT#149582399 - Adds font fabulous dep

AllenBW:
https://www.pivotaltracker.com/story/show/149582399

See that tiny little broom? proooooof it works! (ignore the ⌛️ its just playing around, not committed)
screen shot 2017-07-27 at 9 58 53 am

but hold your 🐴's this is dependent on [changes that need to be made here](https://github.com/ManageIQ/font-fabulous/issues/43).  

once these are in, we'll see the above :broom: which is in turn work to support https://github.com/ManageIQ/manageiq-ui-service/issues/850, such a 🌀 of 💜 

🙇‍♀️ @skateman 

PS: once those changes are made ☝️ I'll have to relock the the ff dep (as in lieu of versions yarn grabs on to commit hashes)

### DIFF
--- a/client/assets/sass/styles.sass
+++ b/client/assets/sass/styles.sass
@@ -3,6 +3,10 @@ $img-base-path: '/' !default
 
 @import 'patternfly'
 
+// Include font-fabulous
+$ff-font-path: '~font-fabulous/assets/fonts/font-fabulous'
+@import 'font-fabulous'
+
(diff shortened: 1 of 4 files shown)


AllenBW:
Annnnndddddddd we're ready to unwip, commit hash matches dep lock https://github.com/ManageIQ/font-fabulous/commit/447376ce2bf290c77d1946ca316ece36b882c264 👀

PR #875, opened 2017-08-18
Set leftBoundary as zero if there is nothing to display

skateman:
**Before:**
screenshot from 2017-08-18 13-51-38

**After:**
screenshot from 2017-08-18 13-51-17

This fixes https://github.com/ManageIQ/manageiq/issues/15847


chriskacerguis:
Simple bug fix, no UX required.

PR #920, opened 2017-09-08
[Finishes #146028835] Allowed users to duplicate services

chalettu:
@miq-bot add_label enhancement
@miq-bot add_label fine/no
@miq-bot add_label catalogs / dialogs
@miq-bot add_label  ux/review

See https://github.com/ManageIQ/manageiq-ui-service/issues/937

PR #979, opened 2017-09-19
CollectionsApi refactored for extensibility (infinite options), conciseness

### LINKED ISSUE #978: Collections-api service query building is not sustainable
Description of problem:
Function with a bunch of if's and object key references isn't a sustainable way to handle building queries as new methods become available. 

This needs to be refactored into something a lil more maintainable, hence this ticket

AllenBW:
Behavior remains identical to before
No need to specify an option to be added to factory (ie filter, attributes), this wasn't sustainable, Needed for `metric_rollups` work, but too good to pigeon hole in that pr 😏  
Fixes tests that probably shouldn't have been passing

🌮 💃 

required by #920 , closes #978

### DIFF
--- a/client/app/core/collections-api.factory.js
+++ b/client/app/core/collections-api.factory.js
@@ -1,136 +1,56 @@
-/* eslint camelcase: "off" */
-
 /** @ngInject */
-export function CollectionsApiFactory ($http, API_BASE) {
-  const service = {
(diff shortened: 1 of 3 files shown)


AllenBW:
Aware that code coverage dropped, not sure what else can be done in this the `collectionsapi.spec` to fix this :-/

chriskacerguis:
This is good stuff

PR #986, opened 2017-09-19
Updates ui-router-dep

AllenBW:
closes #926 

in support of reducing the number of things bithound complains about

PR #987, opened 2017-09-19
Adds deps for bithound to mute, we can't upgrade these at the moment

AllenBW:
In support of reducing the number of things bithound complains about, these four deps can not be upgraded at this time


AllenBW:
Those two failing deps are fine, fixed in #986

chriskacerguis:
Just "testing" bit hound, merging regardless of fail (all "normal" tests are passing).

PR #1022, opened 2017-09-27
BZ 1490103 - Fixed issue for VM custom button actions

chalettu:
@miq-bot add_label bug
@miq-bot add_label fine/no

BZ https://bugzilla.redhat.com/show_bug.cgi?id=1490103

### DIFF
--- a/client/app/states/services/custom_button_details/custom_button_details.state.js
+++ b/client/app/states/services/custom_button_details/custom_button_details.state.js
@@ -12,26 +12,7 @@ function getStates () {
       templateUrl,
       controller: StateController,
       controllerAs: 'vm',
-      title: __('Service Custom Button Details'),
-      params: {
-        button: {
-          value: null
-        },
-        serviceId: {
-          value: null
-        }
-      },
-      resolve: {
-        dialog: resolveDialog,
-        service: resolveService
-      }
-    },
-    'services.vm_custom_button_details': {
-      url: '/:serviceId/vm/:vmId/custom_button_details',
-      templateUrl,
-      controller: StateController,
-      controllerAs: 'vm',
-      title: __('VM Custom Button Details'),
+      title: __('Custom Button Details'),
       params: {
         button: {
           value: null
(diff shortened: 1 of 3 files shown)


chriskacerguis:
@chalettu can you please address that NSP fail?

chalettu:
I checked the library that is the culprit. It is a mime package.  This vulnerability was released today and our dependent packages that depend on that package have not been updated yet.   I recommend we ignore it for now.

chriskacerguis:
Thanks for the info.

@AllenBW I'm good with the merge (with this being so new).  Your vote?

client/app/states/services/custom_button_details/custom_button_details.state.spec.js
```diff
@@ -122,4 +122,80 @@ describe('services.custom_button_details', function() {
       });
     });
   });
+  
+  describe('Custom button actions for a VM', () => {
+    var collectionsApiSpy;
```
AllenBW: prefer, const or let to var. consider stacking all the ones without assignments on the same line

client/app/states/services/custom_button_details/custom_button_details.state.js
```diff
@@ -51,6 +73,7 @@ function StateController ($state, $stateParams, dialog, service, CollectionsApi,
   vm.dialogs = dialog.content
   vm.service = service
   vm.serviceId = $stateParams.serviceId
+  vm.vmId = ($stateParams.vmId ? $stateParams.vmId : null)
```
AllenBW: `$stateParams.vmId || null`  ternary might be a bit of overkill

client/app/states/services/custom_button_details/custom_button_details.state.js
```diff
@@ -25,6 +25,28 @@ function getStates () {
         dialog: resolveDialog,
         service: resolveService
       }
+    },
```
AllenBW: these two states appear very similar, instead of duplicating the code with a mild name change, why not make the vmId an optional param on the existing state?

chalettu: Thought about it.  It is really just a judgement call.  Initially I wanted to make it this way so we could refresh the page successfully but it looks like we are passing button data and that isn't reproducible so we can't refresh either way.  I am kinda fine either way.

chalettu:
vm-actions
@AllenBW , here is a ss of the fix.

PR #1030, opened 2017-09-28
Replace 'Self Service' for 'Service UI' text

AllenBW:
Sooo it was pointed out 👉 : [September 28, 2017 5:52 AM](https://gitter.im/ManageIQ/manageiq-ui-service?at=59ccc64c177fb9fe7e11badc) we're still throwing out 'Self Service' in some places, this does not align with da past edict which state, 
> Thou shal now be the Service UI, sans all self

This pr fulfills that edict. 

Flagging ux to check that the words are alright.

### DIFF
--- a/client/app/skin/skin.module.js
+++ b/client/app/skin/skin.module.js
@@ -1,9 +1,9 @@
 const text = {
   app: {
-    name: 'ManageIQ Self Service'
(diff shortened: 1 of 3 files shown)


chriskacerguis:
Fixing name typo.  No UX review needed.

PR #1046, opened 2017-10-02
Added RBAC to App Launcher

chalettu:
@miq-bot add_label fine/no

### DIFF
--- a/client/app/core/navigation/navigation-controller.js
+++ b/client/app/core/navigation/navigation-controller.js
@@ -64,7 +64,7 @@ export function NavigationController (Text, Navigation, Session, API_BASE, Shopp
         about: RBAC.has('about'),
         documentation: RBAC.has('documentation')
       },
-      appLauncher: RBAC.has('sui_app_launcher'),
+      suiAppLauncher: RBAC.has('sui_app_launcher'),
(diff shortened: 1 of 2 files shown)


client/app/core/navigation/navigation-controller.js
```diff
@@ -64,6 +64,7 @@ export function NavigationController (Text, Navigation, Session, API_BASE, Shopp
         about: RBAC.has('about'),
         documentation: RBAC.has('documentation')
       },
+      appLauncher: RBAC.has('sui_app_launcher'),
```
AllenBW: :ducks: don't hate me, any way we can change vm.appLauncher to vm.suiAppLauncher or vm.sui_app_launcher

I have this feelin' down in my 🍖 that one day we'll thank ourselves for being redic explicit... as in... there won't be any way that conditional rendering can be confused with anything other than rbac.

ALSO looks to be that all the rback we'll be using going forward will lead off with sui so thinking carrying that onwards will also help future us...

chriskacerguis:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1497732

chriskacerguis:
No UX changes here; just adding RBAC.

PR #1082, opened 2017-10-13
BZ#:1501234-Updates sessionstorage when language is switched

AllenBW:
https://bugzilla.redhat.com/show_bug.cgi?id=1501234

lang


AllenBW:
Will work on coverage

PR #1112, opened 2017-10-19
BZ#1503287-Ensure we only fetch those services with display set to true

AllenBW:
https://bugzilla.redhat.com/show_bug.cgi?id=1503287

So the bz points out we don't filter services based on display, best place to do it is in the request, otherwise the pagination will get wacky (imagine, fetching first 20 services and if 18 of them were display none!!)


AllenBW:
oh yeah, i'll fix that

PR #1117, opened 2017-10-19
Removed unnecessary filtercount functions from explorer views

### LINKED ISSUE #1114: Update explorer pagination to use `subquery_count` for pagination
Description of problem:
Well this https://github.com/ManageIQ/manageiq-api/pull/113 kinda borked our pagination, gonna fix it by changing the value we use to set the upperlimit for each explorer

AllenBW:
field subquery_count is used as this does not change with the max_results_per_page flag

closes #1114

### DIFF
--- a/client/app/catalogs/catalog-explorer.component.js
+++ b/client/app/catalogs/catalog-explorer.component.js
@@ -90,23 +90,13 @@ function ComponentController ($state, CatalogsState, ListView, EventNotification
     return CatalogsState.getServiceTemplates(limit, offset).then(success, failure)
 
     function success (response) {
-      vm.serviceTemplateList = response.resources
-
       CatalogsState.getCatalogs(limit, offset).then((response) => {
         vm.catalogsList = response.resources
         vm.loading = false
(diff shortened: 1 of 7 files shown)


AllenBW:
So we have a question @chriskacerguis SO this works, fixes the issue butttttt lets say you have a super fun admin that sets the limit to 1, but pagination in the sui gives the illusion that you could set the limit to 20, 100, 1000… would be great to say “oh, no, sorry user, not options, limit locked in at 1”  

The `max_results_per_page` may one day be exposed to the api, which at that point we can use it to inform our pagination limit, but until that timeeeeeeeeeee I vote we merge this in, cuz broken things need to be fixed

cc: @imtayadeway 👋 ❤️ , @abellotti 🙇 🍕

imtayadeway:
@AllenBW did I help you delete code? :heart: :heart: :heart:

AllenBW:
YOU DID!!!! <3 <3 <3 @imtayadeway

AllenBW:
Created https://github.com/ManageIQ/manageiq-api/issues/136 to track the followon work for to this pr

Issue #1123, opened 2017-10-20
Remove Navigation from RBAC service


chalettu:
Remove all references to Navigation and navigation permissions from RBAC service

chalettu:
@miq-bot assign chalettu
@miq-bot add_label technical debt

PR #1124, opened 2017-10-20
Remove Navigation from RBAC service

chalettu:
GH issue https://github.com/ManageIQ/manageiq-ui-service/issues/1123
@miq-bot add_label fine/no

### DIFF
--- a/client/app/core/rbac.service.js
+++ b/client/app/core/rbac.service.js
@@ -1,7 +1,6 @@
 import { productFeatures } from './product-features.constants.json'
 /** @ngInject */
 export function RBACFactory (lodash) {
-  var navFeatures = {}
   let features = {}
   let currentRole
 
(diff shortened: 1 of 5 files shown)


client/app/core/rbac.service.js
```diff
@@ -45,19 +35,13 @@ export function RBACFactory (lodash) {
...
     currentRole = newRole
   }
 
-  function getNavFeatures () {
-    return navFeatures
-  }
-
-  function navigationEnabled () {
-    return lodash.some(navFeatures, (item) => item.show)
+  function suiAuthorized () {
```
AllenBW: Ya know, while we're at it, why not just shove this function into `authorization.config.js`?  It doesn't so much belong in rbac service anyways

AllenBW: Oh i guess the login state is using it as well 🤔

chalettu: Yeah, thats why i renamed it.  I also figured that defining what allows someone to get into SUI might change so it might make sense to have a small function that determines it.

AllenBW: but for now this doesn't hurt

PR #1133, opened 2017-10-23
Add support for Redux

chalettu:
@miq-bot add_label fine/no


chriskacerguis:
@chalettu this is a good first start!  Much wow.

client/app/actions/token.js
```diff
@@ -0,0 +1,9 @@
+import { TOKEN } from '../constants/token'
+function addToken (token) {
+  return {
+    type: TOKEN.ADD_TOKEN,
+    payload: token
+  }
+}
+
+export default { addToken }
```
ohadlevy: nitpick: I would simply do:
```js
export function addToken (token) { ...
```

or even in its shorter form:
```js
export const addToken = token => ({
  type: TOKEN.ADD,
  payload: token
});
```

client/app/constants/token.js
```diff
@@ -0,0 +1,3 @@
+export const TOKEN = {
+  ADD_TOKEN: 'ADD_TOKEN'
```
ohadlevy: nitpick: I think that TOKEN.ADD would suffice, not sure if we need to be that verbose.

client/app/core/config.js
```diff
@@ -7,14 +9,23 @@ var DEVEL_DOMAINS = [
...
   $logProvider.debugEnabled(isDevel)
   $compileProvider.debugInfoEnabled(isDevel)
 
   // TODO: Remove following line as per: https://docs.angularjs.org/guide/migration#migrate1.5to1.6-ng-services-$compile
   $compileProvider.preAssignBindingsEnabled(true)
 
   $qProvider.errorOnUnhandledRejections(false)
+
+  const storeEnhancers = [
+    autoRehydrate()
```
ohadlevy: what is this used for? is this a premature optimization? ;)

client/app/core/redux.service.js
```diff
@@ -0,0 +1,31 @@
...
+export function ReduxFactory ($rootScope, $ngRedux) {
+  const service = {
+    loadRedux: loadRedux,
+    clear: clear
+  }
+
+  return service
+  function clear () {
+    return new Promise((resolve, reject) => {
+      persistStore($ngRedux, { storage: asyncSessionStorage }, () => {
```
ohadlevy: why not provide an action to clear the state? I would avoid changing the state not via actions.

client/app/core/redux.service.js
```diff
@@ -0,0 +1,31 @@
...
+
+export function ReduxFactory ($rootScope, $ngRedux) {
+  const service = {
+    loadRedux: loadRedux,
+    clear: clear
+  }
+
+  return service
+  function clear () {
+    return new Promise((resolve, reject) => {
```
ohadlevy: imho for everything async please consider using redux-thunk or redux-saga instead.

client/app/core/session.service.js
```diff
@@ -129,9 +139,9 @@ export function SessionFactory ($http, $sessionStorage, $cookies, RBAC, Polling)
   }
 
   // Helpers
-
   function active () {
-    // may not be current, but if we have one, we'll rely on API 401ing if it's not
-    return angular.isString(model.token) ? model.token : false
+    const token = vm.token.token || false
```
ohadlevy: why would it be not defined? your state should have a default state avoid this edge case.

client/app/reducers/token.js
```diff
@@ -0,0 +1,11 @@
+import { TOKEN } from '../constants/token'
+const initialState = {}
```
ohadlevy: this is were you can define your token as false, or maybe uninitialized: true or something along those lines.

ohadlevy: I'm also missing the usage of immutable or seamless-immutable

client/app/core/authorization.run.js
```diff
@@ -0,0 +1,84 @@
...
+  if ($sessionStorage.token) {
+    syncSession()
+  }
+
+  function changeStart (event, toState, toParams, _fromState, _fromParams) {
+    Redux.loadRedux().then(() => {
+      // first lets see if they are logged in and session is valid
+      event.preventDefault()
+      let nextState = ''
+      if (!Session.active()) {
```
ohadlevy: I'm not following this logic at all - can you expand on why its needed? is changeStart is an angular router state or redux state?

client/app/core/authorization.run.js
```diff
@@ -0,0 +1,84 @@
...
+        $state.transitionTo(nextState)
+        return
+      } else {
+        return
+      }
+    })
+  }
+  function authorizedForState (toState) {
+    let authorized = true
+    if (angular.isDefined(toState.data)) {
```
ohadlevy: why not use the redux state? (I think I'm missing the bigger play here)

client/app/core/config.js
```diff
@@ -1,4 +1,6 @@
 /* eslint angular/window-service: "off" */
+import { RootReducer } from '../reducers'
+import { autoRehydrate } from 'redux-persist'
```
ohadlevy: v5 of redux-persist seems to be moving away from autoRehydrate (https://github.com/rt2zz/redux-persist#migration-from-v4-to-v5)

client/app/core/session.service.js
```diff
@@ -1,6 +1,8 @@
+import TokenActions from '../actions/token'
 /* eslint-disable dot-notation */
 /** @ngInject */
-export function SessionFactory ($http, $sessionStorage, $cookies, RBAC, Polling) {
+
+export function SessionFactory (lodash, $http, $q, $sessionStorage, $cookies, $ngRedux, RBAC, Redux, Polling) {
```
ohadlevy: why do you need both $ngRedux and Redux?

ohadlevy:
@chriskacerguis IMHO this was merged pre-maturely - please don't include new code (especially low level infrastructure code) without test coverage. thanks.

AllenBW:
@chriskacerguis We need to revert this. While tests might be passing it totally borks the login process. 

screen shot 2017-10-24 at 8 18 35 am

ohadlevy:
@AllenBW - I can confirm - I saw the same thing earlier today as well. ( I assumed it was pbcak - happy to know its not just me :))

client/app/core/authorization.run.js
chalettu: Change start is an angular UI router event hook.  This whole section of code is refactoring of existing code.  I know the authorization.run.js file is new as of this PR so it seems like new code but it's not.  It is correcting an issue where we had combined an angular config file with an angular run block.  Some of the changes I did implement inside of it were for readability sake since this code hadn't been touched in a long time and needed some refactoring

client/app/core/authorization.run.js
chalettu: I would have to sit down and look at redux state but I think redux state might be out of context for this change.  State in this case refers to an Angular UI router state.  This project uses angular UI router.

PR #1142, opened 2017-10-24
Added ability for login screen to clear password on failed attempt

chalettu:
@miq-bot add_label fine/no

### DIFF
--- a/client/app/states/login/login.state.js
+++ b/client/app/states/login/login.state.js
@@ -77,7 +77,8 @@ function StateController (exception, $state, Text, RBAC, API_LOGIN, API_PASSWORD
     .catch((response) => {
       if (response.status === 401) {
         vm.credentials.password = ''
-        Notifications.message('danger', '', __('Incorrect username or password.'), false)
+        const message = response.data.error.message
+        Notifications.message('danger', '', __('Login failed, possibly invalid credentials. ') + `(${message})`, false)
       }
       Session.destroy()
     })
--- a/client/app/states/login/login.state.js
+++ b/client/app/states/login/login.state.js
@@ -74,8 +74,11 @@ function StateController (exception, $state, Text, RBAC, API_LOGIN, API_PASSWORD
         Session.destroy()
       }
     })
-    .catch(() => {
-      exception.catch('Login failed, possibly invalid credentials.')
+    .catch((response) => {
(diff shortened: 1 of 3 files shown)


chalettu:
Fix for BZ https://bugzilla.redhat.com/show_bug.cgi?id=1089786

client/app/states/login/login.state.js
```diff
@@ -74,8 +74,11 @@ function StateController (exception, $state, Text, RBAC, API_LOGIN, API_PASSWORD
         Session.destroy()
       }
     })
-    .catch(() => {
-      exception.catch('Login failed, possibly invalid credentials.')
+    .catch((response) => {
+      if (response.status === 401) {
+        vm.credentials.password = ''
+        Notifications.message('danger', '', __('Incorrect username or password.'), false)
```
AllenBW: If the response has a server message we should probably display that here

client/app/states/login/login.state.js
```diff
@@ -74,8 +74,11 @@ function StateController (exception, $state, Text, RBAC, API_LOGIN, API_PASSWORD
         Session.destroy()
       }
     })
-    .catch(() => {
-      exception.catch('Login failed, possibly invalid credentials.')
+    .catch((response) => {
+      if (response.status === 401) {
+        vm.credentials.password = ''
```
AllenBW: should `vm.credentials.login` also be cleared out?

AllenBW: I realize ref was for only password to be blank, but if it was an unsuccessful login attempt :-/

chalettu: The RFE just requested password be cleared out.

client/app/states/login/login.state.js
chalettu: We probably could but a couple of thoughts.  The message that comes back just says "Authentication failed".  At a minimum the message we display is a little more specific and it is also translated vs the raw message from the server.   Keep in mind, I didn't add this notification message, merely moved it from the factory up to the state that it pertains to.

AllenBW: And normally I wouldn't press on this matter, but my 👃 is in a bz right now where not showing the full error message issssssss no bueno https://bugzilla.redhat.com/show_bug.cgi?id=1498984

Issue #1144, opened 2017-10-24
Add Redux for State Management


chriskacerguis:
We implement Redux for state management within the service UI in order to facilitate the UIs working more closely together.

- [ ] Migrate appliance info setting to utilize Redux

> Change ApplianceInfo.set(response) to not have to run between page changes. Only set that when someone logs in and use redux to set and retrieve the appliance info.

- [ ] Migrate functionality to set language to utilize Redux

> Update how we persist selected language to use Redux. This needs to change in such a way that removes this notion of having to set language on page reload. Language factory should support retrieving selected language from redux when requested. This means deprecating Language.onReload(response)

- [ ] Migrate functionality for LoadUser to use Redux

> Move all functionality related to a user into its own service and remove it from Session Factory
> Set this use redux to store user data between page reloads

chriskacerguis:
Closing this, tracking in: https://bugzilla.redhat.com/show_bug.cgi?id=1502778

PR #1146, opened 2017-10-24
Updated how appliance info is set and retrieved

chalettu:
@miq-bot add_label fine/no

This is a subtask of https://github.com/ManageIQ/manageiq-ui-service/issues/1144

### DIFF
--- a/client/app/core/appliance-info.service.js
+++ b/client/app/core/appliance-info.service.js
@@ -1,6 +1,6 @@
 import gitHash from '../../version/version.json'
 /** @ngInject */
-export function ApplianceInfo () {
+export function ApplianceInfo ($sessionStorage) {
   let applianceInfo = {}
 
   return {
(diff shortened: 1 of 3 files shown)


AllenBW:
Looks good to me! 

Ok question time though, with this change, appliance info will only be updated when login happens, should we be worried about this info growing stale?  (on current master, if ya reload the app, you fetch and set new appliance info, which is kinda expected, refresh the page, all teh things are refreshed)  I propose we keeeeep this pr, but continue to set it on page refresh (if we already have the fresh info, no reason no to set it)

chalettu:
Based on how this was done and how we pull from session every time someone refreshes the page, the info doesn't change between refreshes either way.  This info also is only ever changed when a appliance is built.  None of the data changes between refreshes.

AllenBW:
Okie dokie lg2m!

PR #1155, opened 2017-10-25
Change explorer default limit to 10 items

### LINKED ISSUE #1147: Set 10 as the default for all list views
Set 10 as the default for all list views

AllenBW:
closes #1147 

## after 1900 x 1200 resolution
screen shot 2017-10-25 at 1 39 08 pm
screen shot 2017-10-25 at 1 39 01 pm
screen shot 2017-10-25 at 1 38 54 pm

### DIFF
--- a/client/app/catalogs/catalog-explorer.component.js
+++ b/client/app/catalogs/catalog-explorer.component.js
@@ -17,7 +17,7 @@ function ComponentController ($state, CatalogsState, ListView, EventNotification
     angular.extend(vm, {
       title: __('Catalogs'),
       loading: false,
-      limit: 20,
+      limit: 10,
       filterCount: 0,
       catalogsList: [],
       serviceTemplateList: [],
--- a/client/app/catalogs/catalog-explorer.component.spec.js
+++ b/client/app/catalogs/catalog-explorer.component.spec.js
@@ -21,7 +21,7 @@ describe('Catalog explorer component - ', () => {
     const catalogSpy = sinon.stub(CatalogsState, 'getServiceTemplates').returns(Promise.resolve({'status': 'success'}))
     ctrl.$onInit()
     done()
-    expect(catalogSpy).to.have.been.calledWith(20, 0)
+    expect(catalogSpy).to.have.been.calledWith(10, 0)
   })
   it('should allow pagination to be updated', () => {
     ctrl.$onInit()
--- a/client/app/orders/order-explorer/order-explorer.component.js
+++ b/client/app/orders/order-explorer/order-explorer.component.js
@@ -22,7 +22,7 @@ function ComponentController ($filter, $state, lodash, ListView, Language, Order
       currentUser: Session.currentUser(),
       loading: false,
       orders: [],
-      limit: 20,
+      limit: 10,
       filterCount: 0,
       ordersList: [],
       selectedItemsList: [],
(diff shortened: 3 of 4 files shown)


chriskacerguis:
@AllenBW could you please post the load times for the page with 20 vs. 10 items?

AllenBW:
@chriskacerguis which page, each page?

AllenBW:
### service explorer `api/services` load times 10: 941 ms

screen shot 2017-10-26 at 8 24 39 am

### service explorer `api/services` load times 20: 5.35s 

screen shot 2017-10-26 at 8 22 35 am

@chriskacerguis

Loicavenel:
it does not seem linear, what about 50? 60?

ohadlevy:
> it does not seem linear, what about 50? 60?
sounds like a N+1 API calls? how many API calls do we do per service ?

AllenBW:
We do a single api call to get the services, ya wanta see time for a 50 to 60 service call?

ohadlevy:
@AllenBW in that case, I would guess its a N+1 problem on the api call itself?

AllenBW:
@ohadlevy yeaaaaaaah seems to be? @imtayadeway is checkin' it out 😍 🏓

serenamarie125:
@ohadlevy @chriskacerguis @Loicavenel let's be aware of how we are affecting the user experience.  I understand there are performance issues, but changing to a default of 10 items per page is only going to help the performance issue the first time they look at a few.

Do we in SUI have the ability to change the page size across the app?  I don't think so, but we should.

By making a change like this, users will need to change the page size every time they come to a view.  So sure the time for the view to initially load may be quicker, but they likely are then going to change the page size, and then have to wait anyways.  From what we have heard from customers to date, typically 50 is the size they want.  

If we have to make this change due to performance, I would request that we at a minimum consider a user setting to set the default page size, so that users can change the default.

chriskacerguis:
@serenamarie125 

> Do we in SUI have the ability to change the page size across the app?

No.  I would think that would be in a "My Settings" page, which I don't think we have (but, I agree with you, that would be helpful to be able to do)

> From what we have heard from customers to date, typically 50 is the size they want.

Why is that?  To quote @ohadlevy "are they searching with Ctrl + F"?  To me it seems that 50 items on the page is a lot of data for the human eye to process.  Just wondering if reducing would be better overall.  Disclaimer: I am totally not a UX expert...just asking questions :)

Loicavenel:
In Ops UI, you can go to My Settings and change how many Items you see in List view across all the pages.

 I am not fully against the 10 by default but if the user wants to browse all the services then select 100, the performance issue will be still there.
We can always think about a better filtering to avoid this long list scrolling (Can we define all possible filtering???) but many users will always do it this way...

chriskacerguis:
@Loicavenel Not sure we have an API to get that setting (I still think having a local settings in the SUI would be a good thing).  That said, ultimately, this is an API issue...you are correct, all we are doing is trying to help any way we can.  Regarding the filtering, that's a UX thing, and so I'll differ to @serenamarie125 

All that said, I'm wondering if we should just hold off on this PR, until @imtayadeway finishes he work on the API.

Thoughts @Loicavenel @ohadlevy @serenamarie125 ?

chriskacerguis:
Thinking about this, I think we should wait and let @imtayadeway do his mojo.

PR #1162, opened 2017-10-26
clear user name field on failed login

ohadlevy:
based on discussion on a similar PR on the ops UI (https://github.com/ManageIQ/manageiq-ui-classic/pull/2514#issuecomment-339520635)
I'm suggesting we align across the two.

I would also appericiate help in how to test the fact
the field is actually empty.

### DIFF
--- a/client/app/states/login/login.state.js
+++ b/client/app/states/login/login.state.js
@@ -76,6 +76,7 @@ function StateController (exception, $state, Text, RBAC, API_LOGIN, API_PASSWORD
     })
     .catch((response) => {
       if (response.status === 401) {
+        vm.credentials.login = ''
         vm.credentials.password = ''
         const message = response.data.error.message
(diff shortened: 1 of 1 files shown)


AllenBW:
I expressed this same sentiment in this comment here, https://github.com/ManageIQ/manageiq-ui-service/pull/1142#discussion_r146620941

Defer to @chriskacerguis on this

chriskacerguis:
I think that alignment is good.  That said, the RFE was for the password only.  Perhaps in the future, we should be aligned in the BZ/RFE before code starts?

ohadlevy:
thanks @chriskacerguis - as a follow up - what is the correct way to write a test for this? (as a user i would expect the form user field to be empty)?

chriskacerguis:
@ohadlevy I think for a solid test for this, we need to turn Sauce Labs back on.  We turned it off b/c the free version we were using took too long to run.

PR #1165, opened 2017-10-26
Replace supports_launch_cockpit with supports_cockpit

AllenBW:
closes #1164

PR #1166, opened 2017-10-26
[FINE] For Fine, replace supports_launch_cockpit with supports_cockpit

AllenBW:
#1165 for fine


AllenBW:
Adding @jntullo for another sanity check 🙇‍♀️

PR #1203, opened 2017-11-02
[WIP]Removed the static $ from the currency display

chriskacerguis:
Removed the static $ from the currency display

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1508802

### DIFF
--- a/client/app/states/dashboard/dashboard.html
+++ b/client/app/states/dashboard/dashboard.html
@@ -83,10 +83,10 @@
                   <span class="pficon magic-dollar"></span>
                 </span>
                 <span class="card-pf-aggregate-status-notification" ng-if="vm.chargeback.used_cost_sum">
-                  ${{ vm.chargeback.used_cost_sum | number:3 }}
+                  {{ vm.chargeback.used_cost_sum | number:3 }}
                 </span>
                 <span class="card-pf-aggregate-status-notification" ng-if="!vm.chargeback.used_cost_sum" translate>
-                  $0
+                  0
                 </span>
               </p>
             </div>


chriskacerguis:
@AllenBW is that something "translatable" (is that a word)?  Honestly, I think that the currency symbol should come from the API.  I can see many scenarios where the language / currency would be different.  For example, a US based company uses USD, but then you have an employee in some other county that speaks another language...but, the currency should still be in USD.

Additionally, there is the matter of currency conversion...so simply translating might not work.

Thoughts?

AllenBW:
@chriskacerguis totally is translatable, the markup i posted above will ensure its picked up by the gettext extract method and added to the master file of things that need to be translated.  Agree with ya though, we need to know if the currency is converted in the api.  If it is not, and is always in usd, keeping the dollar sign or indicating usd might be a good idea

ohadlevy:
my 2cents is that until we do offer a multi currency billing, we should probably keep defaulting to $. (maybe you can change in under settings?) /cc @Loicavenel ?

chriskacerguis:
Well, by "defaulting" I mean it was literally a "$" in the code.

chriskacerguis:
@ohadlevy according to the BZ, that is something we do...it looks like the chargeback reports have that data.

Loicavenel:
When you create a chargeback rate, you can set the currency you want to use (limited to 4 today).. Then the report will use it.

AllenBW:
SOOOOO looking into chargeback reports... presently we don't get any info about currency back...
#### The call
`GET http://localhost:3001/api/services?expand=resources&filter%5B%5D=service_id%3Dnil&attributes=chargeback_report`

#### The result
```
"chargeback_report": {
        "results": [
          {
            "start_date": "2017-08-01T00:00:00.000Z",
            "display_range": "Aug 2017",
            "vm_id": "10000000001510",
            "vm_name": "Test",
            "storage_allocated_metric": 10737418240.0,
            "storage_allocated_cost": "353.0",
            "storage_used_metric": 1064616539.3767705,
            "storage_used_cost": "699.9999999999996460974216461181640625"
          }
        ]
      }
```

`

Loicavenel:
API update !!

chriskacerguis:
@Loicavenel @serenamarie125 @ohadlevy so, I was speak to @niroy about this, and he brings up a few interesting use cases, and concerns.  Talking with the emoji-nator ( @AllenBW ) and digging through the code it presents with more "interesting" things.

1.  As @AllenBW said, we don't get anything back about the currency.  We should look into adding that.
2. In the Catalog we are representing the cost using a $ (essentially t-shirt sizing) the cost, but those are calculated by us with USD in mind, so those "t-shirt sizes" don't translate to other currencies.

Thinking that the API needs to (1) return the currency and (2) return the "appropriate t-shirt size".  I would suggest they return a 1 - 4 int value and then we can "transform" that to the appropriate currency symbol.

Thoughts?

Loicavenel:
@chriskacerguis what are the rules today about "T-shirt Size" in SUI?

AllenBW:
wipping this till it either 👍 or 👎

AllenBW:
@chriskacerguis can we close this one?

PR #1239, opened 2017-11-14
Fixed issue with buttons not supporting being disabled

chalettu:
Fixed issue with custom buttons supporting disabled state
@miq-bot add_label bug
@miq-bot add_label  gaprindashvili/yes

### DIFF
--- a/client/app/services/custom-button/_custom-button.sass
+++ b/client/app/services/custom-button/_custom-button.sass
@@ -22,13 +22,13 @@
 
 .custom-button-action
   &.disabled
-    color: $color-gray-3
+    color: $app-color-light-gray-7
 
     &:hover
       background-color: transparent
       border-color: transparent
-      color: $color-gray-3
-            
+      color: $app-color-light-gray-7
+
 .toolbar-pf-actions
   .dropdown-menu
     .disabled
--- a/client/app/services/custom-button/_custom-button.sass
+++ b/client/app/services/custom-button/_custom-button.sass
@@ -19,3 +19,14 @@
       border-color: $app-color-light-blue-4
       color: $app-color-medium-gray-2
       text-decoration: none
+.custom-button-action
+  &.disabled
+    color: $color-gray-3
+    &:hover
+      color: $color-gray-3
+      background-color: transparent 
+      border-color: transparent 
+.toolbar-pf-actions
+  .dropdown-menu
+    .disabled
+      cursor: not-allowed
--- a/client/app/services/custom-button/_custom-button.sass
+++ b/client/app/services/custom-button/_custom-button.sass
@@ -19,13 +19,16 @@
       border-color: $app-color-light-blue-4
       color: $app-color-medium-gray-2
       text-decoration: none
+
 .custom-button-action
   &.disabled
     color: $color-gray-3
+
     &:hover
+      background-color: transparent
+      border-color: transparent
       color: $color-gray-3
-      background-color: transparent 
-      border-color: transparent 
+            
 .toolbar-pf-actions
   .dropdown-menu
     .disabled
--- a/client/app/services/custom-button/custom-button.component.js
+++ b/client/app/services/custom-button/custom-button.component.js
@@ -26,6 +26,9 @@ function CustomButtonController ($state, EventNotifications, CollectionsApi, RBA
   }
 
   function invokeCustomAction (button) {
+    if (!button.enabled) {
+      return
+    }
     if (button.resource_action && button.resource_action.dialog_id) {
       const options = {
         button: button,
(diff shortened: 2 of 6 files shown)


client/app/services/custom-button/_custom-button.sass
```diff
@@ -19,3 +19,14 @@
       border-color: $app-color-light-blue-4
       color: $app-color-medium-gray-2
       text-decoration: none
+.custom-button-action
+  &.disabled
+    color: $color-gray-3
```
AllenBW: but `$color-gray-3` doesn't exist in `_app_colors.sass` ?  (below its added to `_colors.sass`

client/app/services/custom-button/custom-button.component.js
```diff
@@ -26,6 +26,9 @@ function CustomButtonController ($state, EventNotifications, CollectionsApi, RBA
   }
 
   function invokeCustomAction (button) {
+    if (!button.enabled) {
```
AllenBW: rather than an empty return consider wrapping the function body in an `if(button.enabled)`

client/app/services/custom-button/_custom-button.sass
```diff
@@ -19,3 +19,14 @@
       border-color: $app-color-light-blue-4
       color: $app-color-medium-gray-2
       text-decoration: none
+.custom-button-action
+  &.disabled
```
AllenBW: writing a new disabled style feels a lil queasy, why not just extend an existing disabled class?

client/app/services/custom-button/custom-button.html
```diff
@@ -23,12 +23,13 @@
     </button>
     <ul uib-dropdown-menu role="menu" aria-labelledby="{{buttonGroup.id}}">
       <li ng-repeat="button in buttonGroup.buttons" role="presentation">
-        <a class="custom-button-action"
+        <a class="custom-button-action" 
            role="menuitem"
            tabindex="-1"
+           ng-class="{'disabled': !button.enabled}"
```
AllenBW: this is all ya should need to conditionally apply disabled style, unsure about rolling your own class

client/app/services/custom-button/_custom-button.sass
chalettu: The sass work is to create a more specific selector that only applies to custom action buttons.  The specificity was needed to override some default classes that were applying styles

client/app/services/custom-button/_custom-button.sass
chalettu: Do you think it should be in app_colors instead of colors?

client/app/services/custom-button/custom-button.html
chalettu: We didn't have any classes that matched the disabled styling that is traditionally done in our other drop downs.  Without rewriting the button component to match exactly the UI-bootstrap drop downs, this was the easiest path to a solution.

AllenBW: 🤔 but we do have disabled buttons, does applying the disabled status alone yield the disabled appearance?

client/app/services/custom-button/_custom-button.sass
AllenBW: So `_colors.sass` isn't imported in this file, `$color-gray-3`  doesn't exist in `_app_colors.sass` ya shouldn't be seeing any color change in the case of an absent variable, so I'm think we're either pulling color from another disabled class (which removing the sass work would reveal) or we have a bigger issue with sass variable bleed

client/app/services/custom-button/custom-button.html
chalettu: This is actually for drop downs not "buttons" per se.  We dont have any classes that handled disabling menu options in a drop down.  The html is actually A tags so some styling had to be done to simulate a disabled state since A tags don't natively have a disabled state.  I made the styling match how disabled seemed to look and operate on uib-dropdown elements.

AllenBW: 👍 can we get a ss of the change in action?

chalettu:
test_dialog button
Sample of what a disabled custom button looks like.

client/app/services/custom-button/custom-button.html
AllenBW: Ok last ask then i'll approve... still super suspicious about that color being used that isn't defined in the color file being imported, would you mind building the app, checking from /ui/service? Promise after that i'll go away

chriskacerguis:
@chalettu what's the BZ for this?

chalettu:
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1509959

simaishi:

```
$ git log -1
commit 8a83b3165f71657a38d36c74ae8349314ba0ffe1
Date:   Tue Nov 14 13:12:31 2017 -0600

    Merge pull request #1239 from chalettu/disable-custom-buttons
    
    Fixed issue with buttons not supporting being disabled
    (cherry picked from commit 70eb2df7285b67c8cad66986a9a64cd6be7657b1)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1513498
```

PR #1257, opened 2017-11-21
[Gaprindashvili] Add Gaprindashvili translations

mzazrivec:


client/gettext/json/available_languages.json
```diff
@@ -1 +1 @@
-{"es-ES":"Español","fr":"Français","ja":"日本語","ko":"한국어","nl":"Nederlands","pt-BR":"Português do Brasil  ","pt-PT":"Português","zh-Hans-CN":"简体中文","zh-TW":"國語"}
```
AllenBW: @mzazrivec  I'm a little concerned about the missing language names for pt-PT zh-TW, ko and nl... if we don't have values specified here, the user won't be able to select any of those languages

mzazrivec: What I used to do was if the catalog did not have `locale_name` translated in Zanata, I entered it there myself.

mzazrivec:
> Need to figure out why avaliable_languages.json is losing language names

Either the catalog in question doesn't have `locale_name` translated or there's a problem
in the `available-languages` task.

client/gettext/json/available_languages.json
AllenBW: Yeah that's what I did for the ones that were absent above

AllenBW:
Looks like the catalogs don't have locale_name, maybe we do this pr without the update to avaliable_languages.json?

mzazrivec:
> Looks like the catalogs don't have locale_name, maybe we do this pr without the update to avaliable_languages.json?

Right, we cannot have `available-languages.json` without those pretty locale names.

mzazrivec:
@AllenBW I restored the original `available-languages.json`.

AllenBW:
@mzazrivec thanks a bunch for this, you're a :godmode: among 🐙 🙇‍♀️

PR #1280, opened 2017-11-27
Update console behavior as per api changes

AllenBW:
https://github.com/ManageIQ/manageiq/pull/16519

PR #1245, opened 2017-11-15
BUG#1242 - Cleanup login state, notifications fill full viewport

AllenBW:
closes #1242
BEM was negating a great deal of patternfly styles we were missing here, pulled it!

### what we look like now
screen shot 2017-11-16 at 7 57 28 am

PR #1286, opened 2017-11-29
[Changelog] Update for Sprint 74

JPrause:
[skip ci]


AllenBW:
@JPrause hmmm ok so things that were slated for this sprint, but not merged in time, should be pushed to next sprint?  #1245 and #1280 

Also

screen shot 2017-11-29 at 1 38 49 pm

JPrause:
@AllenBW I think it easier if I e-mail you regarding the PRs I didn't include and the reason behind it.
Short answer, since the changelog is for the community, there are certain tagged PRs that we exclude from a changelog.

PR #1289, opened 2017-11-30
[GAPRINDASHVILI]BZ#1514080-Update pf dep, bugfixes resolve dropdown calculation error

AllenBW:
https://bugzilla.redhat.com/show_bug.cgi?id=1514080

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "angular-gettext": "2.3.10",
     "angular-gettext-cli": "1.2.0",
     "angular-messages": "1.6.6",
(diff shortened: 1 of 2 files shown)


simaishi:
@AllenBW Can I just backport https://github.com/ManageIQ/manageiq-ui-service/pull/1287 ?

AllenBW:
@simaishi oh yah know, i was thinking not, because there might be a lockfile conflict... but maybe so.... could give it a try, if it works then i'll close this if not, then yay !

PR #1293, opened 2017-11-30
Notifications should gated on sui_notification product feature

AllenBW:
closes https://bugzilla.redhat.com/show_bug.cgi?id=1518279

~~### this needs to be merged first https://github.com/ManageIQ/manageiq/pull/16817~~

Ok I could see how this bz might cause a much heartbreak though... a bunch of existing roles that have sui_product features, **do not** have the product feature `sui_core` or `sui_notifications`

for instance, `EvmRole-vm_user` has 
  - sui_vm_details_view
  - sui_vm_console
  - sui_vm_web_console
  - sui_vm_tags
  - sui_vm_retire
  - sui_vm_start
  - sui_vm_stop
  - sui_vm_suspend
  - sui_orders_show
  - sui_orders_operations

and is missing core or notifications... so this user would NOT see notifications

SOOO @Loicavenel @chriskacerguis maybe we should revisit  `miq_user_roles.yml` and add more product features (specifically everything under `sui_core`) to each role that has any sui product features?

### DIFF
--- a/client/app/core/event-notifications.service.js
+++ b/client/app/core/event-notifications.service.js
@@ -1,5 +1,5 @@
 /** @ngInject */
-export function EventNotificationsFactory ($timeout, lodash, CollectionsApi) {
+export function EventNotificationsFactory ($timeout, lodash, CollectionsApi, RBAC) {
   const state = {}
   const toastDelay = 8 * 1000
   const service = {
@@ -19,6 +19,7 @@ export function EventNotificationsFactory ($timeout, lodash, CollectionsApi) {
     dismissToast: dismissToast,
     setToastDisplay: setToastDisplay
   }
+
   notificationsInit()
 
   return service
@@ -152,26 +153,30 @@ export function EventNotificationsFactory ($timeout, lodash, CollectionsApi) {
     removeToast(notification)
     service.markRead(notification)
   }
+
   function setToastDisplay (enabled) {
     state.toastsEnabled = enabled
   }
+
   // Private
   function add (notificationType, type, message, notificationData, id) {
-    const group = lodash.find(state.groups, {notificationType: notificationType})
-    const newNotification = {
-      id: id,
-      notificationType: notificationType,
-      unread: angular.isDefined(notificationData) ? notificationData.unread : true,
-      type: type,
-      message: message,
-      data: notificationData || {},
-      href: id ? '/api/notifications/' + id : undefined,
-      timeStamp: (new Date()).getTime()
-    }
+    if (RBAC.has(RBAC.FEATURES.CORE.NOTIFICATIONS)) {
+      const group = lodash.find(state.groups, {notificationType: notificationType})
+      const newNotification = {
+        id: id,
+        notificationType: notificationType,
+        unread: angular.isDefined(notificationData) ? notificationData.unread : true,
+        type: type,
+        message: message,
+        data: notificationData || {},
+        href: id ? '/api/notifications/' + id : undefined,
+        timeStamp: (new Date()).getTime()
+      }
 
-    group.notifications.unshift(newNotification)
-    updateUnreadCount(group)
-    showToast(newNotification)
+      group.notifications.unshift(newNotification)
+      updateUnreadCount(group)
+      showToast(newNotification)
+    }
   }
 
   function miqFormatNotification (text, bindings) {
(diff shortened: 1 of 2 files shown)


AllenBW:
cc @Loicavenel @chriskacerguis mentioned ya both above for input, not sure if editing the message still pings yah

client/app/core/event-notifications.service.js
```diff
@@ -154,26 +155,30 @@ export function EventNotificationsFactory ($timeout, lodash, CollectionsApi, Ses
...
-      id: id,
-      notificationType: notificationType,
-      unread: angular.isDefined(notificationData) ? notificationData.unread : true,
-      type: type,
-      message: message,
-      data: notificationData || {},
-      href: id ? '/api/notifications/' + id : undefined,
-      timeStamp: (new Date()).getTime()
-    }
+    if (!RBAC.has(RBAC.FEATURES.CORE.NOTIFICATIONS)) {
```
himdel: This should probably not be negative...

`RBAC.has('sui_notifications')` gives me `true`... but no notifications ever appear :).

himdel:
@AllenBW not sure `EventNotificationsFactory.add` is the right place ... because this also affects all the SUI-generated notifications.

I will no longer get a flash message when submitting the shopping cart for example.

Should the RBAC feature *really* affect all the notifications? Or is this just about notifications coming from the server?

AllenBW:
@Loicavenel need your confirmation here before we move on this...

With this pr, unless a role has the sui_notifications product feature, they will see no notifications.  Is this the desired behavior?

himdel:
@Loicavenel ^ this would mean the user gets no confirmation for any of their actions.

No success message when editing something, no error message when something fails, etc.

AllenBW:
@Loicavenel sooooo need to know, what we're doing here... there is a bz attached to this pr...

I should mention, this will only be notifications generated by api responses, this does not impact those from the server, conveyed through websockets...

himdel:
> I should mention, this will only be notifications generated by api responses, this does not impact those from the server, conveyed through websockets...

@AllenBW .. you sure? ;)

https://github.com/ManageIQ/manageiq-ui-service/pull/1293/files#diff-2664ce6e85761a22ded3cf80dd5b676eR200

AllenBW:
fine @himdel FINE, this will impact em all, see I'm wiser in the afternoon than in the morning :-)

AllenBW:
@himdel updated as per your comments (yuge face palm over that one, i forgot the product features of the role i was logged in as), can yah approve, but not merge this when you get chance?

himdel:
Did we ever get an ACK from @Loicavenel that this is indeed what is wanted?

I still think the product feature should only affect notifications coming from the server.

So far, the only clue I have is https://bugzilla.redhat.com/show_bug.cgi?id=1518279#c3 .. 
But the problem is: even people with read-only access should get to see errors, shouldn't they?

---

Either way, assuming this is the right fix, you have my :+1:, it works, the user will see no notifications at all.

(Any point in approving, since travis is red?)

Loicavenel:
@himdel there is one use case. If the user is simply a "reader" type, he can see all MyServices but cannot run any actions then Notify is not required. So, the rules is we should have notify only for user who can perform actions

himdel:
@Loicavenel Ok, thanks, makes sense I think :). (I'm guessing even a user who cannot perform actions may run into an error they won't see, but hey, at least they won't complain about it :).)

Loicavenel:
@himdel if the UI itself has an issue, will he not see a "Pop up" with the issue? 
Or everything is via notification now?

himdel:
@Loicavenel definitely any errors like failing to access a resource, or any javascript errors.

@AllenBW I'm not sure if there is a different kind of popup for at least some class of errors? (Like, the server going away?)

AllenBW:
@Loicavenel @himdel With this work, users will see no notifications, js or otherwise..

This pr should only be merged only after the product feature `sui_notifications` has been added to each role though, otherwise no user will see notifications :-P (working on that pr presently)

AllenBW:
@gtanzillo Here's the pr where we enable sui notification product feature... with this, if the sui_notification feature is absent, no notification query will be made

AllenBW:
@himdel looks like https://github.com/ManageIQ/manageiq/pull/16817 made it in!!! We're good to go on this pr 👍

himdel:
LGTM, merged :)

(with the already mentioned reservations that users without this feature will not be able to tell if anything went wrong or why)

simaishi:

```
$ git log -1
commit 72d90a2fc8d03279f8505c3e3f1cfe3e5fcc2360
Date:   Tue Feb 27 12:35:59 2018 +0000

    Merge pull request #1293 from AllenBW/bug/master/#1518279-notification-creation-gated-by-rbac
    
    Notifications should gated on sui_notification product feature
    (cherry picked from commit 280e297f032346b622af69937d38c151348d9ba3)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1552801
```

PR #1295, opened 2017-11-30
A dep no longer needed, causes loading of angular too many times error

AllenBW:
this dep was introduced to resolve this issue, https://github.com/angular/angular.js/issues/8934

It has since been closed, we no longer need this dep, pruning it to remove the following error
 
### Here was the issue, console error "loading angular too many times"
screen shot 2017-11-30 at 2 43 45 pm

### This fix resolves the issue, dep is no longer required, svgs show up just fine in both prod and dev (rh os img is an svg in the screenshots)
screen shot 2017-11-30 at 2 41 59 pm
screen shot 2017-11-30 at 2 51 16 pm
screen shot 2017-11-30 at 2 50 25 pm


himdel:
The angular issue was closed with the caveat that you need to be using `$locationProvider.html5Mode({enabled: true, requireBase: false});` (or not using the `<base>` tag).

Since we do neither, we're still affected by that issue.

But.. looking at the original, looks like this only effects svg properties with URLs inside them - both `clip-path` and `mask` can take an URL.

Our SVGs don't seem to be using `mask` at all, but they are using `clip-path` with URL:

```
client/assets/images/providers/vendor-openstack_infra.svg
15:			<g clip-path="url(#SVGID_2_)" enable-background="new    ">
```

But.. the url is a fragment in the same CSS, which probably means we're not affected by the issue after all :).

---

Would love to be sure though, @AllenBW where do I find a place where vendor SVGs are actually used?

I see a reference in `resource-details` but haven't quite been able to figure out the right screen.

himdel:
.. never mind, `http://localhost:3001/services/10000000000025/resource=10000000001130` works for me :).

It's: Services > pick a service with more than 0 vms, of type openstack > click on the detail screen > click on the VM.

himdel:
LGTM, merging :)

(but if we ever add a svg with a clip-path that's an actual url, this may break)

himdel:
@AllenBW Apparently, you need to add a review for yourself before I can merge :/.

simaishi:

```
$ git log -1
commit ecd81c8d44c20a5c92819da05815db8c3cbcd315
Date:   Tue Dec 5 17:12:55 2017 +0000

    Merge pull request #1295 from AllenBW/bug/master/#0000-warning-loading-angular-too-many
    
    A dep no longer needed, causes loading of angular too many times error
    (cherry picked from commit 50fd12110a45f19f91141947fa2020e5d0edad12)
```

PR #1327, opened 2017-12-07
Remove greenkeeper, cuz gets really noisy whenever angular updates

AllenBW:
Dropping this in favor of a weekly dep update cadence

@Fryguy so this work is half of the effort, we still have to remove the app from the repo (something I can't do)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,10 @@ cache:
   - node_modules
 before_install:
 - npm install -g yarn@1.0.1
-- npm install -g greenkeeper-lockfile@1
-before_script: greenkeeper-lockfile-update
 install:
 - yarn install
 after_script:
  - cat reports/coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
- - greenkeeper-lockfile-upload
 notifications:
   webhooks:
     urls:
--- a/README.md
+++ b/README.md
@@ -5,7 +5,6 @@
 [![Code Climate](https://codeclimate.com/github/ManageIQ/manageiq-ui-service/badges/gpa.svg)](https://codeclimate.com/github/ManageIQ/manageiq-ui-service) 
 [![Chat](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/ManageIQ/manageiq-ui-service?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 [![Translate](https://img.shields.io/badge/translate-zanata-blue.svg)](https://translate.zanata.org/zanata/project/view/manageiq-ui-service)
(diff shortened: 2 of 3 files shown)


AllenBW:
@ohadlevy ignoring the angular dep it is definitely an option.

Though, I am worried the deps will become neglected/forgotten in the noise of multiple update cadences.

The stronger case against greenkeep, *any* dependency that has a flurry of versions is going to create this same raukaus, we use many dependencies (greater chances of this happening) and we can't ignore them all 🤔

himdel:
Coming in as an almost-outsider, IMO the problems with greenkeeper are that...

* it can't group related upgrades together - it should never do 7 PRs for 1 angular update
* it can't group unrelated upgrades together / work in digest mode - 3 PRs a day is a bit too much, once a week would be about perfect
* it pings everyone on the CODEOWNERS list, just because
* none of the PRs are ever tested. IMO we need less PRs which were never tested by the author, not more :).


(I still much prefer the approach in ui-classic, where we always default to `~1.2.3`, and lock specific versions only in releases, but.. I guess there's Reasons to do it differently here :).)

--- 

Because greenkeeper specifically decided not to allow digest PRs (https://github.com/greenkeeperio/greenkeeper/issues/253), I think we may want to look for something else - and since there's no integration tests anyway, this could be as simple as somebody running a weekly cronjob to do the update, testing it works, and pushing the PR :).

(IMO they are trying to solve the wrong problem - yes, when doing digest PRs, you don't know why the tests fail if they do, but so what, if they do, you can start looking. Most often they won't fail, and then the choice between testing 1 PR a week or 3 a day sounds pretty clear :).)

AllenBW:
@himdel wooah wooaah wooaah, I test EVERY pr before approving it (update the lockfile, spin up the sui, make sure things work). Also these prs have to pass travis, so if the dep changes the behavior that our testing checks for, we know.

but tired of testing each one, woud be much better to happen as a batch, do deps weekly would be nicer...

ohadlevy:
if its not useful for you, than i have no problem stop using it, i find it useful in other projects, but i can understand how painful it is. i do hope that our automated tests can simulate the type of testing you are doing (e.g. to figure out regressions).

himdel:
Don't worry Allen, I meant the author of the PR, not the merger :) - so, greenkeeper not you :)

So.. exactly my point, you either need to test 15 or 1 :).

AllenBW:
@ohadlevy depends on the dep that changes.. travis doesn't presently check that the `yarn build` task operates correctly.. or its a dep that has a visual component, presently, tests don't have provisions to ensure consistency of views...

ohadlevy:
right.. the project i use it has jest snapshot testing for react - so its
trivial to see if something breaks...

PR #1344, opened 2017-12-14
BZ#1525883-Add missing translation tags to order details view

AllenBW:
closes https://bugzilla.redhat.com/show_bug.cgi?id=1525883

So one might think that the "Order #" text is translatable, but it is in fact provided by the server, ends up ya can add names to orders, but if there is no, the default one shows up: https://github.com/ManageIQ/manageiq/blob/d3544b2c4d4a0c6796a236619349fa46bf57f2e7/app/models/service_order.rb#L35

### DIFF
--- a/client/app/states/orders/details/details.html
+++ b/client/app/states/orders/details/details.html
@@ -22,7 +22,7 @@
               </div>
               <div class="ss-details-header__title">
                 <h2>{{ ::vm.order.name }}</h2>
-                <h4>ordered {{ ::(vm.order.placed_at || vm.order.updated_at) | date }}</h4>
+                <h4> {{'ordered on' | translate}} {{ ::(vm.order.placed_at || vm.order.updated_at) | date }}</h4>
               </div>
             </div>
           </div>


client/app/states/orders/details/details.html
```diff
@@ -22,7 +22,7 @@
               </div>
               <div class="ss-details-header__title">
                 <h2>{{ ::vm.order.name }}</h2>
-                <h4>ordered {{ ::(vm.order.placed_at || vm.order.updated_at) | date }}</h4>
+                <h4> {{'ordered' | translate}} {{ ::(vm.order.placed_at || vm.order.updated_at) | date }}</h4>
```
himdel: This is slightly problematic in that it assumes "ordered (on) 2017-12-15" is the same "ordered" as in the sentence "It's already ordered." Which it may not be. Also, some languages may prefer to say "2017-12-15 ordered".

According to https://angular-gettext.rocketeer.be/, looks like just moving the `translate` attribute on that `h4` should do the right thing.

(Essentially we want the equivalent of `sprintf(_("ordered %{on}"), { on: textNotSubjectToTranslation })` instead of `_("ordered") + textNotSubjectToTranslation`, so that the translator gets that extra context info.)

AllenBW: @himdel thanks for the perspective!! Agree, the context of the word "ordered" matters. `<h4 translate>` will effectively put the translator in charge of modifying the rest of that html though, a situation I do not believe is ideal. Maybe this should be reworded all together?

himdel: Aah, yeah, if that's a thing we don't want...

What about just adding a preposition? `{{"ordered on" | translate}}` gives more context. (But still can't be reordered.)

After that.. well we can always do `sprintf(__("ordered %s"), $tilter('date')(vm.order.placed_at || vm.order.updated_at))` in the controller..

himdel: Cc @mzazrivec, if you have a preference here.. :)

AllenBW:
@himdel yah ok with this one? (reworded as requested)

simaishi:

```
$ git log -1
commit aa896a12b20db9da0d4207bc9448514e33623cc1
Date:   Fri Jan 19 17:58:56 2018 +0100

    Merge pull request #1344 from AllenBW/bug/master/#1525883-untranslated-order-text
    
    BZ#1525883-Add missing translation tags to order details view
    (cherry picked from commit ff66554f9180b13c74387385f00ee9738f537981)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1553475
```

PR #1346, opened 2017-12-15
[FINE] BZ#1518017- Dialog fields load values on init

chalettu:
@miq-bot add_label bug
@miq-bot add_label fine
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1518017

### DIFF
--- a/client/app/states/catalogs/details/details.state.js
+++ b/client/app/states/catalogs/details/details.state.js
@@ -1,4 +1,5 @@
 /* eslint-disable arrow-body-style */
+/* eslint camelcase: "off" */
 import templateUrl from './details.html';
 
 /** @ngInject */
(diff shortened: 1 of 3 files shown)


chalettu:
Screenshot demo of functionality.  First field when it is updated triggers refresh.  Second field is a dynamic one that can be updated.  After clicking on refresh the dialog loads data.  Going back to first field and editing then will trigger field 2 updating.  
refresh_demo

chalettu:
@miq-bot add_label fine/yes

chalettu:
@eclarizio , could you take a look at this one and review it for me.  Thanks in advance

client/app/states/catalogs/details/details.state.js
```diff
@@ -81,9 +82,11 @@ function Controller($stateParams, CollectionsApi, EventNotifications, DialogFiel
         vm.dialogUrl = `service_catalogs/${vm.serviceTemplate.service_template_catalog_id}/service_templates`;
         if (dialogs.subcount > 0) {
           DialogFieldRefresh.setupDialogData(vm.dialogs, allDialogFields, autoRefreshableDialogFields);
-                   
           angular.forEach(allDialogFields, function (dialogField) {
-            dialogField.refreshSingleDialogField = function () {
+            dialogField.refreshSingleDialogField = function (button) {
+              if (button && !dialogField.load_values_on_init) {
+                dialogField.load_values_on_init = true;
+              }
```
eclarizio: I think this is fine but I want to clarify, so for something that has `dialogField.load_values_on_init` set to false initially, if we click on the refresh button, this will cause the field to refresh and then set `load_values_on_init` to true. I don't think this would cause any weird side effects because we've already gone past the "init" stage, right?

chessbyte:
@chalettu would you please respond to comment from @eclarizio 
@simaishi would you please monitor this PR and merge it when ready for inclusion in next errata

client/app/states/catalogs/details/details.state.js
chalettu: This shouldn't cause any weird side effects because it would have been initialized once by virtue of clicking on refresh.

AllenBW:
@simaishi this one is good to go whenever you get a hot second! 🙇 🍭

simaishi:
@chalettu Only blocker PRs can be merged to Fine branch now, so this will need to wait for the next round...

PR #1351, opened 2018-01-03
Redux implementation in SUI

chalettu:
adds initial support for redux and also implements an example for nav counts for menu.


client/app/actions/nav.js
```diff
@@ -0,0 +1,10 @@
+import { NAV } from '../constants/nav'
+
+export const addCount = function (itemType, count) {
+  return {
+    type: NAV.ADD_COUNT,
+    itemType: itemType,
+    count: count
+  }
+}
+export default { addCount }
```
ohadlevy: can write this a bit shorter:
```js
import { NAV } from '../constants/nav';

export const addCount = (itemType, count) => ({
  type: NAV.ADD_COUNT,
  itemType,
  count,
});
export default { addCount };
```

client/app/core/navigation.service.js
```diff
@@ -1,12 +1,22 @@
...
   var service = {
     get: getNavigation,
     menuItems: menuItems,
     init: initNavigation,
     updateBadgeCounts: updateBadgeCounts
   }
+  const actions = NavActions
+  const mapStateToThis = function (state) {
+    return { navCount: state.nav }
+  }
```
ohadlevy: can be written as:
```js
const mapStateToThis = ({ nav }) => ({
  navCount: nav,
});
```

chalettu: @ohadlevy , these syntax changes look great.  Do you know what version of ES they were introduced.  I wanted to sit down and read up on all the different syntactical options just so I can start taking more advantage of them in the future

ohadlevy: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

chalettu:
@ohadlevy , here is where the usage of the new Redux action is being used.  https://github.com/ManageIQ/manageiq-ui-service/pull/1351/files#diff-c8366e1545d13a612d9e3938c9ad7524R94 .   We are using it for the simple counters that are used in the left nav of SUI.  For testing.  Do you have any good links you could provide me on testing the redux state changes with Jest?

client/app/core/navigation.service.js
```diff
@@ -81,7 +91,8 @@ export function NavigationFactory (RBAC, Polling, POLLING_INTERVAL, CollectionsA
     menuItems.forEach((item) => {
       if (angular.isDefined(item.badges)) {
         getBadgeCount(item.badgeQuery.field, item.badgeQuery.filter).then((count) => {
-          item.badges[0].count = count
+          service.addCount(item.state, count)
+          item.badges[0].count = service.navCount[item.state]
```
ohadlevy: will this re-trigger when a value change? e.g. is there a direct relationship between a notify event on the redux state to this function?

ohadlevy:
@chalettu I would start with the resources at https://redux.js.org/ , its probably easiest with jest, have a look at https://redux.js.org/docs/recipes/WritingTests.html. /cc @vojtechszocs

client/app/core/navigation.service.js
chalettu: @ohadlevy This is currently not reactive in the way you described.  Currently the navigation functionality for getting counts is a polling function.  The polling system calls this function  and it updates the different counts in a loop.

vojtechszocs:
Sorry for missing this, I'll review tomorrow. +1 on Redux adoption!

client/app/reducers/index.js
```diff
@@ -0,0 +1,6 @@
+import { combineReducers } from 'redux'
+import { NavReducer } from './nav'
+
+export const RootReducer = combineReducers({
+  nav: NavReducer
```
karelhala: Why do we want to rename the reducer? I think it would be better to use full name here so we know with what we are dealing with.

```javascript
export const RootReducer = combineReducers({
  NavReducer
})
```

karelhala:
Looks good, I have just couple questions about dirsctructure and how to load reducers.

The workflow importing reducer and adding it to combineReducers feels for me like too much repetetive work, which might be forgoten.

What do you think about this approach:

Given we have filder structure like:
#### `app/reducers/nav`

Here we will have our `index.js` which is written in this PR

#### `app/reducers/some-other`

Here we'd have 3 reducers named `firstReducer.js`, `secondReducer.js`, `thirdReducer.js` and `index.js`
Reducers will be function/classes with reducer function, `index.js` would look like:
```javascript
import FirstReducer from './firstReducer';
import SecondReducer from './secondReducer';
import ThirdReducer from './thirdReducer';

export {FirstReducer, SecondReducer, ThirdReducer}
// Here can be some other exports,
// no need to just import reducers from files, for instance like this
export function FourthReducer(state = {}, action) {
  //reducer actions
}
```
#### `app/reducers`
Here is the main logic of combining all reducers and it would look like

```javascript
import * as Nav from './nav';
import * as Reducers from './some-other';

export const RootReducer = combineReducers({
  nav: Nav.NavReducer, // we can rename reducer
  ...Reducers
  // this is the most important line, every
  // exported reducer from `some-other` folder will be combined to RootReducer
})
```

#### Reasoning
The reason behind this, is that whenever someone creates new PR which will have reducer inside of it, only files which are related to his work will be touched, no need to edit `index.js` inside reducers folder. Also much smaller `RootReducer` definition, since we can import multiple reducers as single const and then spread it to combine it together with other reducers.

client/app/core/navigation.service.js
```diff
@@ -1,12 +1,22 @@
...
     menuItems: menuItems,
     init: initNavigation,
     updateBadgeCounts: updateBadgeCounts
   }
+  const actions = NavActions
+  const mapStateToThis = (state) => ({
+    navCount: state.nav
+  })
+
+  const unsubscribe = $ngRedux.connect(mapStateToThis, actions)(service)
```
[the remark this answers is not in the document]
vojtechszocs: Yup, for objects that live "forever" like factories, no need to unsubscribe from Redux store changes.

For the rest, you can do something like

```
$scope.$on('$destroy', unsubscribe);
```

to ensure proper cleanup of your Angular controller (or component).

client/app/core/navigation.service.js
vojtechszocs: You'll probably want to converge to a system where the state of a unit (component, controller, service, factory, etc.) is a transformation of the whole app state, managed through the Redux store.

The app state (in Redux store) should be the single source of "truth" - whenever the state changes, all relevant units should be notified and updated according to the new data. You can consider using libraries like `reselect` to optimize this process, e.g. avoid component update if the particular bits of data relevant for that component didn't change (but other bits of data did change).

In a pure Redux architecture, a unit doesn't have its own state - all the data is managed through the Redux store. Change of the app state happens by firing actions (actions flow up) which cause state modification which then causes all subscribers to be notified (data flow down).

But I understand that adopting pure Redux arch can be challenging so you can improve step by step. The main strength of Redux is that you know exactly what the current app data is (because it's in one place, not dispersed across X units) and how exactly does it change (because you have your reducers dedicated for this purpose).

client/app/reducers/nav/index.js
```diff
@@ -0,0 +1,16 @@
...
+const initialState = Immutable({
+  services: 0,
+  orders: 0,
+  catalogs: 0
+})
+
+export function NavReducer (state = initialState, action) {
+  switch (action.type) {
+    case NAV.ADD_COUNT:
+      return state.set(action.itemType, action.count)
```
vojtechszocs: Instead of having a generic `ADD_COUNT` action type, consider being more specific, it can have positive impact on code readability, e.g. `ADD_SERVICE_COUNT`.

Alternatively, consider being more specific in terms of your action creator functions, for example:

```
export const addServiceCount = (count) => ({
  type: NAV.ADD_COUNT,
  itemType: 'services',
  count: count
})
```

This way, your code can do `addServiceCount(1)` instead of `addCount('services', 1)` or `addCount(constants.services, 1)`.

More generally, your Redux action creators and action types are the contract between your code and the Redux infra. Therefore, these things should be as sensible as possible, which in turn makes your code itself more sensible.

package.json
```diff
@@ -23,7 +23,8 @@
     "start:mock-api": "cd node_modules/@manageiq/manageiq-api-mock/ && node index.js",
     "start:mock-webserver": "cd server && node app.js",
     "start:all": "yarn start:mock-api & yarn start:mock-webserver",
-    "test": "yarn vet && yarn gettext:extract && yarn gettext:validate && karma start karma.conf.js --single-run",
+    "test": "yarn vet && yarn gettext:extract && yarn gettext:validate && karma start karma.conf.js --single-run && yarn test:jest",
+    "test:jest": "jest .*.jest.spec.js",
```
vojtechszocs: This is a nice way to integrate Jest into existing test toolchain, it allows you to gradually move tests over to (executed by) Jest.

client/app/core/config.js
```diff
@@ -1,20 +1,25 @@
...
 
   // TODO: Remove following line as per: https://docs.angularjs.org/guide/migration#migrate1.5to1.6-ng-services-$compile
   $compileProvider.preAssignBindingsEnabled(true)
 
   $qProvider.errorOnUnhandledRejections(false)
+  const storeEnhancers = []
+  if (hasDevTools) {
+    storeEnhancers.push(window.__REDUX_DEVTOOLS_EXTENSION__())
+  }
+  $ngReduxProvider.createStoreWith(RootReducer, [], storeEnhancers)
```
vojtechszocs: I see you're not going with any Redux middlewares for now, which is fine :smiley: you can add them later on, once you need to handle async operations or side effects.

At first, Redux might feel too boilerplate'ish, but once you get to the point when you really need to handle async stuff (like HTTP) or side effects (like local storage) and make those part of logical flows in your Redux architecture, you'll see the true value of Redux :yum:

vojtechszocs:
@karelhala your approach to combine reducers with object spread (...) operator makes perfect sense.

I was, traditionally, used to having a setup where each specific reducer like **`reducers/nav/index.js`** contains a `default` export, so that in your root reducer, you can simply do:

```js
import nav from './nav'
import foo from './foo'

const rootReducer = combineReducers({
  nav, foo
})
```

More generally, this reflects the idea that reducers form a hierarchy (tree) where you can theoretically use `combineReducers` in smaller (non-root) reducers, so basically an analogy of:

```js
const rootReducer = combineReducers({
  combineReducers({ // inline'd nav reducer
    // sub-reducers of nav reducer
  }),
  combineReducers({ // inline'd foo reducer
    // sub-reducers of foo reducer
  })
})

```

But I guess it's a matter of preference. Whatever makes you most productive, go with that.

client/app/core/navigation.service.js
ohadlevy: @chalettu what would it take to make sure this value is stored only in redux?

package.json
ohadlevy: /cc @himdel

client/app/actions/nav/nav.jest.spec.js
```diff
@@ -0,0 +1,12 @@
+/* global test */
```
AllenBW: Instead of a file name `nav.jest.spec.js` what about sticking with `foo.spec.js`?  Any reason to differentiate jest tests from other tests? If there is what about `foo.test.js` (another well used nomenclature)

chalettu: @AllenBW  If we did what you were suggesting  then the karma test runner would pick the tests up and they would not work being run through karma.  The tests have to be run through Jests test runner.  I know its an annoyance but the easiest way I could come up with for the two test runners to coexist peacefully.

client/app/core/navigation.service.js
```diff
@@ -1,12 +1,22 @@
+import NavActions from '../actions/nav'
+/* eslint no-unused-vars: 0 */
```
AllenBW: Looks like all the vars are used we should not need this, maybe I'm missing something 🤔

client/app/actions/nav/nav.jest.spec.js
AllenBW: Yeah I thought as much so can we go with `.test` or something equally as obvious, but not specific to the tool?

AllenBW: lookin' at the `karma.conf` `.test.js` should work as its only set to pickup on `.spec.js`
https://github.com/AllenBW/manageiq-ui-service/blob/e7c13af07adea8b5f217925697714c878bd1b3cf/karma.conf.js#L45

client/app/reducers/nav/index.js
chalettu: Thank you for the feedback on this.  I think that in this particular case I weighed your suggestion and also weighed the notion that creating a list of those other actions would make the code more verbose but the actions each has would be identical.  I think in this case just for this instance it makes sense to keep the code down to a minimum rather than creating 4+ exactly identical actions.

client/app/core/navigation.service.js
chalettu: @ohadlevy , I just made changes so that the count for each menu item is retrieved from the redux store when the menu is defined.  The other code that is already in this PR handles what happens when polling occurs.  All code for nav counts now only uses redux for setting and retrieving counts.

client/app/core/navigation.service.js
chalettu: Just took care of removing the unsubscribe variable which went ahead and removed this line.

chalettu:
@ohadlevy I am pulling this PR out of WIP.  Do you think we should have this ported go G release?

karma.conf.js
```diff
@@ -34,6 +34,9 @@ module.exports = function (config) {
       {pattern: './tests/**/*.json', included: false, served: true, nocache: false},
       {pattern: './client/assets/images/**/*', included: false, served: true, nocache: false}
     ],
+    exclude: [
+      './client/app/**/*.jest.spec.js'
```
AllenBW: `./client/app/**/*.test.spec.js`

yarn-error.log
```diff
@@ -0,0 +1,8819 @@
+Arguments:
```
AllenBW: Thinking we don't want to be adding the `yarn-error.log` to the repo... Maybe add it to the gitignore?

karma.conf.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const webpackConfig = require('./config/webpack.test.js')
+const webpackConfig = require('./config/webpack.testing.js')
```
AllenBW: NICE CATCH! 🙇 🎊  (it took a double take for me to....)

AllenBW:
So as I understand it, the TL;DR of what this work does is:
* adds redux and creates a state
* reconfigures navigation menu badge counts to be store in the state
* adds jest to test the previously mentioned work

The left nav menu object looks kinda like this

client/app/core/navigation.service.js
ohadlevy: in theory, I would the action to trigger fetching the API, meaning that the logic is initiated from the action and not the other way around

karma.conf.js
ohadlevy: nitpick: should use es6 import vs require

ohadlevy:
>  Do you think we should have this ported go G release?
@chalettu no :)

AllenBW:
So TL;DR of this work appears to be:
* adds a state to the sui
* reconfigures left navigation menu badge counts to be store in the state
* adds jest to test the previously mentioned work

This is a little bit of what that lnav object looks like:

```
      {
        title: __('Dashboard'),
        state: 'dashboard',
        iconClass: 'fa fa-dashboard',
        permissions: true
      },
      {
        title: __('My Services'),
        state: 'services',
        iconClass: 'pficon pficon-service',
        badgeQuery: {
          'field': 'services',
          'filter': 'ancestry=null'
        },
        badges: [
          {
            count: 0,
            tooltip: __('Total services ordered, both active and retired')
          }
        ],
        permissions: RBAC.has(RBAC.FEATURES.SERVICES.VIEW)
      },...
```
I have a few questions: 
1. Why only store the badge count? There are many other imitable items that could be added, or is the point to not add those things that wouldn't change to the state? 

2. Continuing off this comment: https://github.com/ManageIQ/manageiq-ui-service/pull/1351/files#r159937587 maybe rather than just calling this `navCount`, `navBadgeCount` is a bit more specific 

3. Much of this work refers to nav, without being specific to left or upper, for example https://github.com/chalettu/manageiq-ui-service/blob/57ed73c900325462b08b7435efe19aa3154a7603/client/app/core/navigation.service.js#L14 wondering if it might benefit the future to differentiate ?

4.  oh... wait... crap... forgot, ok will edit this when I remember 🍰 🤙

ohadlevy:
> Why only store the badge count? There are many other imitable items that could be added, or is the point to not add those things that wouldn't change to the state?
We can store a lot more in the state, the idea was to start with something simple?

I think that if they 100% static, it wouldnt matter much, but ideally you want your application to respond to pure data/state - this makes thing more generic and easier to reuse (but I have no idea really on the specifics in our case).

One thing important to note is that state is seralizable, e.g. you cant store function calls like rbac or even gettext __ 
> 
> Continuing off this comment: https://github.com/ManageIQ/manageiq-ui-service/pull/1351/files#r159937587 maybe rather than just calling this navCount, navBadgeCount is a bit more specific

what ever suite us best, as new stuff, I'm pretty sure we will refactor often :)
> 
> Much of this work refers to nav, without being specific to left or upper, for example https://github.com/chalettu/manageiq-ui-service/blob/57ed73c900325462b08b7435efe19aa3154a7603/client/app/core/navigation.service.js#L14 wondering if it might benefit the future to differentiate ?

Well, the state represent the data, not how its used, so in this case, if you use the same data in multiple places that's great, as you don't need to worry about synchronizing it etc.

AllenBW:
> Well, the state represent the data, not how its used, so in this case, if you use the same data in multiple places that's great, as you don't need to worry about synchronizing it etc.

OH OH OH waiiiiiitttttt just a sec, I had a 💡 moment... So really we shouldn't be calling these navBadgeCount, or even navCounts , we should call them what they are (forgive the absence of a real name) parent `services`, ordered `service_orders`, and visible `service_templates`. These count are also the going to be the default number of total results for each of those explorer views and also used on the current dashboard (well `services` is).

AllenBW:
@chalettu Dig the name changes! Yeah we could use these counts on the dashboard... default total results for the explorers... a number of places!  
I see two last pieces to complete (probably am overlooking a bit, prepared to eat my words) and I'll be g2g to approve:
* import vs require comment
* new yarn.lock

chalettu:
@AllenBW , I updated the lock file and looked at the import vs require comment.  This code in question is node run code not clientside javascript.  That import syntax doesn't work on this code.  I tried it and get an unexpected token error when trying to use the import syntax.

client/app/reducers/appCounters/index.js
```diff
@@ -0,0 +1,18 @@
...
+import Immutable from 'seamless-immutable'
+const initialState = Immutable({
+  SERVICES_COUNT: 0,
+  CATALOGS_COUNT: 0,
+  ORDERS_COUNT: 0
+})
+
+export function AppCountersReducer (state = initialState, action) {
+  switch (action.type) {
+    case APPCOUNTERS.SERVICES_COUNT:
```
AllenBW: Quick question.. camel case, snake case, any reason for changing it up here?

chalettu:
@AllenBW , can you take a look. I renamed the offending variables to to consistently use camel case.

AllenBW:
@chalettu right on! okie dokie, lemme jump in the branch real quick before.....

chalettu:
@ohadlevy , any final input before we close out this PR?

ohadlevy:
The pr looks good, I do have two questions:

1. What is the next step? Should we actually drive the API call from redux
in order to maintain all of that as a state?

2. Is the counter a first class store state? Or rather it's a computation
(e.g. count the number of services we got from the API)? I would argue that
if that's the case we shouldn't (maybe not in this pr) store the actual
number in the state but rather use something like reselect (
https://github.com/reactjs/reselect).

Overall :+1:

chalettu:
@ohadlevy , I think the next step might be to possibly look into drive the API queries into redux.  I believe at this point that should be considered out of scope for this PR.  We also have counters that exist on our dashboard.  I would say migrating those counters to utilize redux would be another priority we should set.  
I would also like to work on redux state persistence into session because we have some outstanding issues with how our language logic works and I think a bit of it could be simplified by moving the logic into redux.

2. The API is returning a count and we aren't responsible for any computation of those counts.

ohadlevy:
Thanks!

client/app/actions/appCounters/index.js
```diff
@@ -0,0 +1,16 @@
...
+
+export const addServicesCount = (count) => ({
+  type: APPCOUNTERS.SERVICES_COUNT,
+  count: count
+})
+export const addCatalogsCount = (count) => ({
+  type: APPCOUNTERS.CATALOGS_COUNT,
+  count: count
+})
+
```
ohadlevy: nitpick: extra line (or just inconsistent between the rest of the functions).

ohadlevy:
thanks!

PR #1360, opened 2018-01-17
Removes unnecessary header x-miq-group

AllenBW:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1531626

The sister has been slain, no longer a blocker, this pr is the winner ⚔️ 
~~This pr has a sister, https://github.com/ManageIQ/manageiq-api/pull/287 , another way to solve the above bz~~

 
In the past, was used for group switching: http://manageiq.org/docs/reference/latest/api/overview/authorization

This is now done through the api, this code is no longer needed and in fact creates the issue (bz above), tl;dr - special characters in headers are bad and not to spec

PS: Totally tested group switching and logging in with invalid groups, everything behaves identical to master

### DIFF
--- a/client/app/core/authentication-api.factory.js
+++ b/client/app/core/authentication-api.factory.js
@@ -11,8 +11,7 @@ export function AuthenticationApiFactory ($http, $base64, API_BASE, Session, Not
       $http.get(API_BASE + '/api/auth?requester_type=ui', {
         headers: {
           'Authorization': 'Basic ' + $base64.encode([userLogin, password].join(':')),
-          'X-Auth-Token': undefined,
-          'X-Miq-Group': undefined
+          'X-Auth-Token': undefined
         }
       }).then(loginSuccess, loginFailure)
 
--- a/client/app/core/authentication-api.factory.spec.js
+++ b/client/app/core/authentication-api.factory.spec.js
@@ -19,8 +19,7 @@ describe('Authentication API', () => {
         'http://localhost:9876/api/auth?requester_type=ui', {
           'headers': {
             'Authorization': 'Basic dGVzdDp0ZXN0',
-            'X-Auth-Token': undefined,
-            'X-Miq-Group': undefined
+            'X-Auth-Token': undefined
           }
         })
     })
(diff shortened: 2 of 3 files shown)


jntullo:
I really like this approach, thanks @AllenBW for doing this / testing! If this is removed from the SUI, i think it will be a good idea to revisit the group header at all in the API.

h-kataria:
@jntullo does this mean https://github.com/ManageIQ/manageiq-api/pull/287 should be closed

jntullo:
@h-kataria no, we will still fix the API issue, in the event that users are still using this and have a special character in the group name

himdel:
Well ... the good thing is, this looks like it fixes the accent bug, and definitely doesn't make things works :).

Given this is a blocker, merging.

The bad thing is: group switching is completely broken in SUI.
Switching between super-administrator and user doesn't update the menus, you have to switch the group, log out and log in again for the group switch to take effect in the UI.
It does seem to switch the group for API.

himdel:
@AllenBW can you please look into fixing that? The original implementation used to cause the whole app to reload as if you were logging in, for the changes in the UI to take effect. (Or is it that we're caching rbac even after group switching?)

AllenBW:
@himdel To say completely broken is a bit of an overstatement, numbers and visibility of sections is restricted, but yeah, looks like the lnav menu no longer updates on the change.  That behavior is unrelated to this bug though.

AllenBW:
Yah know... it looks like the whole menu no longer respects rbac, no matter what, logging in/out yah always see the same items no matter which user 😱

simaishi:

```
$ git log -1
commit 1d0be6466282425bcd290da4cec8bbde34511f0b
Date:   Thu Jan 18 12:17:17 2018 +0100

    Merge pull request #1360 from AllenBW/bug/master/#1531626-x-group-removal
    
    Removes unnecessary header x-miq-group
    (cherry picked from commit 9c9a5d89e4ee82b3c1b58c85d32d828a4db41a6e)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1536047
```

PR #1371, opened 2018-01-23
Refactors deps

AllenBW:
* updates and refactors sui deps
* remediates missing dependency warnings
* removes unused deps (ie use of the module `patternfly.table` is the only reason to have any data tables deps)
* improves readability of inter-dependency dependencies 

Ones that are seemingly innocuous but required by other libs are grouped together and annotated as being related, the following links rather helpful in this work:
https://github.com/patternfly/angular-patternfly/blob/master/README.md
https://github.com/ManageIQ/ui-components/blob/master/package.json#L89

Cleans up some awkward (service-details) styling that was hidden by a magic bug, long story short, we `require('bootstrap')` but which didn't resolve due to the packages package.json (it doesn't have a root `main`)

### see, look, its not broken, its the same! 
screen shot 2018-01-31 at 9 41 17 am


package.json
```diff
@@ -32,138 +32,131 @@
...
     "jquery-ui-bundle": "1.12.1",
-    "lodash": "4.17.4",
+    "lodash": "4.17.5",
     "merge": "1.2.0",
+    "moment": "2.20.1",
     "moment-timezone": "0.5.14",
     "ng-annotate-loader": "0.6.1",
     "ng-redux": "3.5.2",
-    "ngprogress": "VictorBjelkholm/ngProgress#v1.1.3",
+    "ngprogress": "VictorBjelkholm/ngProgress#07f8fab",
```
himdel: This a `noVNC` look surprising - any reason to depend on a particular commit here?

himdel:
Removed `require`s:

  * `angular-datatables`
  * `angular-datatables/dist/plugins/select/angular-datatables.select`
  * `angular-file-saver`
  * `bootstrap`
  * `bootstrap-select`
  * `bootstrap-switch`
  * `bootstrap-touchspin`
  * `datatables.net-select/js/dataTables.select`
  * `datatables.net/js/jquery.dataTables`

client/app/services/service-details/service-details.html
```diff
@@ -258,7 +258,7 @@ <h2 translate>Resources</h2>
                                     </div>
                                 </div>
                                 <div ng-if="!item.retired " class="list-view-pf-actions">
-                                    <div class="btn-group dropdown-kebab-pf" uib-dropdown dropdown-append-to-body
```
himdel: Does this change belong here?

Seems like there's plenty more uses of `dropdown-kebab-pf`, including in the same file.

himdel:
So.. what I'm still confused about:

  * depending on commits
  * `dropdown-kebab-pf`
  * the new dependencies
  * adding bootstrap as a dependency but removing it from requires

AllenBW:
Totally not working today but wanted to get in a response...

* you'll notice the release commit hash refers to the release tag, (for no vnc tag v0.6.2 === e8986fa) while tags can always be fiddled with, commit hashes can't, its a better way to lock it down (also when referring to git urls, you'll notice the absence of branch tags https://docs.npmjs.com/files/package.json#github-urls)

* yah can trace back the styling to some pf-sass stuff dealing with `class="list-view-pf-actions"` it yields a 16px font size applied to the dropdowns

* new dependencies, including boostrap are/were missing peer dependencies that yarn and webpack would whine about when building/ installing  so we add bootstrap because its required by others, but not needed for our app directly, as we use it through others

* good question, i wasn't aware we had any mechanism for running the specs in the browser, yarn test (only thing for test running) is a console only kinda thing

* yeah so dev dependencies (as refactored in this pr) are all the things we don't for `yarn build` to run successfully...  so build vs tooling as it stands presently, except, really really, meaning someone intentionally checked 😏

himdel:
Aah, ok, makes sense, I guess :)

Except for:

> you'll notice the release commit hash refers to the release tag, (for no vnc tag v0.6.2 === e8986fa) while tags can always be fiddled with, commit hashes can't, its a better way to lock it down (also when referring to git urls, you'll notice the absence of branch tags https://docs.npmjs.com/files/package.json#github-urls)

Please don't :) From a commit hash, you can't tell which version it actually is supposed to be, and there's already `yarn.lock` for that.

AllenBW:
So ya wanta keep them on the tags? That change and this is okie dokie?

himdel:
Yes please, because we keep the version info that way..  I'm assuming most of those are github links because the npm package is not released, something we can fix perhaps :).

As for the rest, I don't think there's anything problematic assuming nothing breaks :). I'm still surprised that the bootstrap stuff came up now, but.. hey, it does seem to work :).

AllenBW:
@himdel looooooooooks like da thing went 💚 🥗 🍏 📗  🙇‍♀️

himdel:
Tested, it works! Hwah hwah hwah hwah hwah :)

PR #1373, opened 2018-01-25
Refactors dashboard state to a component

AllenBW:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1540733

In support of the longer term fix brought on by https://github.com/ManageIQ/manageiq-ui-service/pull/1372

So there's a few bzs this fixes (for real this time): 
* https://bugzilla.redhat.com/show_bug.cgi?id=1536382
* https://bugzilla.redhat.com/show_bug.cgi?id=1533222
* https://bugzilla.redhat.com/show_bug.cgi?id=1536468

### look, its working 🍰 
rbac

Goal of this is to be identical to original, cept no more 403 kick ya outta the app state change errors

TODO 👍 : (In this)
- [x] add the basic component, put all the existing logic in the right places
- [x] build out the dashboard.component.service
- [x] refactor dashboard.component.spec
- [x] ensure dashboard.component is working as intended


AllenBW:
Thinking its probably a good idea to get @serenamarie125 to check this out also... specifically the loading of the state before the widgets are present, maybe each column should have a loading spinner?

serenamarie125:
@AllenBW thanks for bringing this up ! From a PatternFly perspective, the cards should be displayed with their loading state.  

The base card describes this state here: http://www.patternfly.org/pattern-library/cards/base-card/#design

serenamarie125:
FYI - After chatting with @AllenBW as well as some of the visual designers, we will look into an alternate solution from a PF perspective - possibly a way to tie cards together, so that if data isn't available, you can only show a single spinner.  This won't be trivial and a design should be done up front, implemented in PF and then be brought back to product.  So for now, let's stick with the PF solution.  Sound good ?

AllenBW:
@serenamarie125 so present state, these aren't actually pf cards... you ok with the top two "header" cards being swapped for the design of the pfInfoStatusCard? http://www.patternfly.org/angular-patternfly/#/api/patternfly.card.component:pfInfoStatusCard

AllenBW:
@serenamarie125 updated the first comment to have a new gif, showing loading state! also... if the user doesn't have permissions to view the card, its removed when loading fails

serenamarie125:
@miq-bot add_label ux/approved

serenamarie125:
@miq-bot remove_label ux/review

AllenBW:
no need for a new gif, ensures cards are always as left as possible (for the case of when services fails (due to permissions) 
screen shot 2018-01-30 at 3 21 21 pm

client/app/core/navigation.service.js
```diff
@@ -123,4 +130,10 @@ export function NavigationFactory (RBAC, Polling, POLLING_INTERVAL, $ngRedux, Co
       })
     })
   }
+
+  function setPermissions () {
+    menuItems[1].permissions = RBAC.has(RBAC.FEATURES.SERVICES.VIEW)
```
chalettu: Can we create constants for these array positions. I know its simple but it will make the code a little easier to read.

AllenBW: The assignment is self informing, I don't so much agree with creating variables that server no other purpose than to increase verbosity .

dclarizio:
@himdel your attention required on this blocker PR . . . thx, Dan

himdel:
So... before reviewing this any further... do we *need* to `gaprindashvili/yes` this?

I mean, it's great to see code rewritten to components.
It's just not so great as a blockers-only bugfix...

client/gettext/po/manageiq-ui-service.pot
```diff
@@ -4,7 +4,7 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Project-Id-Version: \n"
 
-#: ./client/app/states/dashboard/dashboard.html:88
+#: ./client/app/components/dashboard/dashboard.html:88
```
himdel: recreating `pot` files should not be a part of the PR

AllenBW: @himdel will remove, but why not? especially when file refactoring takes place?

AllenBW:
@himdel agreed its a large change to be a bugfix, but removing the "application gateway" (the three asset resolves that challenged a user enters entering dashboard state) state resolves is the best way to fix the bz.. 

When any of those fails, the app router throws us into a  `$stateChangeError` loop. Unfortunately throwing a `event.preventDefault()` will not address this infinite loop.  These calls have to be allowed to fail, means we need to get em outta the resolve, means dashboard now has to be a component.

himdel:
@AllenBW So.. if we just backported a change that prevents the resolve from erroring..?

Something like 

```diff
 function resolveServicesWithDefinedServiceIds (CollectionsApi, RBAC) {
   if (RBAC.has('service_view') && RBAC.has(RBAC.FEATURES.SERVICES.VIEW)) {
     const options = {
       expand: 'resources',
       filter: ['service_id=nil'],
       attributes: ['chargeback_report']
     }
 
     return CollectionsApi.query('services', options)
+      .catch(function() {
+         return null;
+      });
   }
 
   return true
 }
```

?

himdel:
@AllenBW One thing I do notice..

1. log in with EvmGroup-desktop
1. theres no count shown for My Services in the menu (as that role can't access services)

1. log in with any other role
1. switch to EvmGroup-desktop
1. the count is there (but clicking the menu item still gives me "There was an error loading the services. Use of the read action is forbidden")


Not sure if this is just about removing the count (but that doesn't sound blocker-worthy), or if the menu item should not be there (but that would be a separate bug entirely).

simaishi:
> So.. if we just backported a change that prevents the resolve from erroring..?

@AllenBW @himdel since I don't see a reply to that question... can either one of you confirm that we will be backporting the entire PR to Gaprindashvili branch?

AllenBW:
@simaishi yes, whole pr, sorry for delayed response 😭

simaishi:
@AllenBW @himdel @dclarizio 

This is built on top of https://github.com/ManageIQ/manageiq-ui-service/pull/1375, which is not in Gaprindashvili yet (not `blocker`), and can't be backported as is.  I think we have 2 options here... Either 1) we create a separate PR for Gaprindashvili without https://github.com/ManageIQ/manageiq-ui-service/pull/1375 (not sure how much work that is...), or 2) we take https://github.com/ManageIQ/manageiq-ui-service/pull/1375 now and backport this PR as well.

Two PRs touch many files/lines and I'm not sure how 'safe' it is to get either/both PRs to gaprindashvili branch now. Looking for a guidance from someone familiar with both changes...

cc @chrispy1

AllenBW:
To answer the how much work question (@simaishi ), wouldn't say more than a few hours of effort to copy this into a Gaprindashvili specific branch, the only overlap is a single file, and even that is not more than 5 or so lines

himdel:
@AllenBW Can you take care of creating a gaprindashvili version of this?

simaishi:
Backported to Gaprindashvili via https://github.com/ManageIQ/manageiq-ui-service/pull/1378

PR #1375, opened 2018-01-29
BZ#1501098 Fix for Service UI not taking default language

chalettu:
@miq-bot add_label bug


chalettu:
This PR is accomplishing a handful of things.
1. Change the order of precedence for how language is set and persisted in service UI.
2. Conversion of the language-switcher directive into an angular component
3. Made sure the language selections made on the login page of Service UI get persisted and are honored. 

Here is how the language selection order of precedence works.  
If in Classic UI a language is set as default and that is the only setting made, then SUI will assume this language.
If in Classic UI a user changes their default language selection that language will be used by SUI. 
If you choose to use the language switcher on the Service UI login page or on the nav menu switcher, those changes will change your language settings for both Service UI and Classic UI.

chalettu:
So this PR contains some CSS changes but they are all made in support of converting the directive into an angular component.  Still, I decided to do a quick SS of the changed code.  No UX changes should be required for this PR
language-selector

AllenBW:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1501098

client/app/core/core.module.js
```diff
@@ -22,7 +22,7 @@ import { DialogFieldRefreshFactory } from './dialog-field-refresh.service.js'
 import { EventNotificationsFactory } from './event-notifications.service.js'
 import { ExceptionModule } from './exception/exception.module.js'
 import { LanguageFactory } from './language.service.js'
-import { LanguageSwitcherDirective } from './language-switcher/language-switcher.directive.js'
+import { LanguageSwitcherComponent } from './language-switcher/language-switcher.component.js'
```
AllenBW: Thoughts on moving this to `./app/components`?

chalettu: I thought we were going to get rid of that folder, how about /app/shared ?

client/app/layouts/_application.sass
```diff
@@ -85,3 +86,22 @@ li
     height: 38px
     margin-top: 10px
     width: 320px
+
+.language-switcher
```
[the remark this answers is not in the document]
chalettu: good feedback, I had it there originally because of some other css but didnt look at it again at the very end.

client/app/core/core.module.js
AllenBW: OOoooo Shared kinda makes sense... it is after all used in two distinct places, login.html and application.html (one is a state the latter a layout)

chalettu: moved the component to /app/shared

chalettu:
Demos of language selection on login page
language-selector

chalettu:
Demo of changing language while logged in 
language-selector-logged-in

chalettu:
change_language_reload
Screenshot of fix for navigating to another page after changing language and hitting reload page.  @AllenBW take a look.  Good find by the way.

simaishi:
@chalettu Changes in `client/app/core/navigation/navigation-controller.js` is conflicting on backport because similar change was made in gaprindashvili PR https://github.com/ManageIQ/manageiq-ui-service/pull/1378. I'm not sure what to take, please clarify.

himdel:
Second BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1672324

Will take care of the conflict...

himdel:
Created https://github.com/ManageIQ/manageiq-ui-service/pull/1526 for gaprindashvili

PR #1387, opened 2018-02-19
Added numeral dependency

chalettu:
Change that was needed to support. https://github.com/ManageIQ/ui-components/pull/244
@miq-bot add_label bug

### DIFF
--- a/client/app.js
+++ b/client/app.js
@@ -14,7 +14,7 @@ window.sprintf = require('sprintf-js').sprintf
 window.c3 = require('c3/c3.js')
 window.d3 = require('d3/d3.js')
 window.patternflyVersion = 4
-window['numeral'] = numeral
+window.numeral = numeral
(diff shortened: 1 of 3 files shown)


chalettu:
@miq-bot add_label gaprindashvili/no

client/app.js
```diff
@@ -13,6 +14,7 @@ window.sprintf = require('sprintf-js').sprintf
 window.c3 = require('c3/c3.js')
 window.d3 = require('d3/d3.js')
 window.patternflyVersion = 4
+window['numeral'] = numeral
```
AllenBW: thinkin' this oughta follow the pattern from lines 15 and 14... just cuz... ya know, keeping things consistent

soooo...

`window.numeral = require('numeral')` (especially considering cuz we're bringing in the whole lib)

PR #1396, opened 2018-03-06
Updates nonbreaking deps

AllenBW:
Updates nonbreaking deps


himdel:
(Wondering why we have all those dependencies exactly, no `~` or `^`, but if you like to do these things manually.. LGTM :))

AllenBW:
Yeah! so the mentality with this was, in order to ensure everything is working (dep wise) precisely as intended, versions would be exact as yah can never 100% trust other libraries. Sure it might be a bug change, or a an enhancement that isn't breaking, but whats assurances does one have there isn't some surprise conflict that breaks all of your code.

PR #1399, opened 2018-03-14
ui-comp dialog select/multi-select assumes correct styling

AllenBW:
So this spawned from an observation made over here: https://github.com/ManageIQ/ui-components/pull/266#issuecomment-372790492

tl;dr - dialog selects are broken on master, this work fixes them

This work restores the odd voodoo combination of relying on bootstrap/select from ui components 

This work rolls back jquery, which introduces a bug rendering selects kinda useless

This work updates ui-components to a version with dialog select that supports pf4

## Looooooook it works now
screen shot 2018-03-14 at 12 34 53 am


himdel:
Can confirm both single-select and multiselect look good with this one :).

But... something's indeed fishy.. If I remove `yarn.lock` and `yarn` again, it stops working completely, back to the browser version, and error in the console.

So.. I'm thinking something in yarn.lock either doesn't quite match package.json now, or we still haven't nailed some dependency down properly.

AllenBW:
oh curious... ok... will look into that...

himdel:
Thanks.. I tried picking just some changes in yarn.lock and not others, but half the time when trying I get an error about not being able to find bootstrap.. So IDK.

If you don't find anything, maybe we can just merge, and hope it doesn't break again, but .. feels like there's a surprise lurking still :)

AllenBW:
OK sooooo maybe your yarn.lock was showing its age? in sore need of a rebuild 🤔 

deleted it... reinstalled, everything looks good, 's that was done was restore the updated jquery version... but maybe it cleaned out some odd collisions that was happing ? not entirely sure... but give this a try...

himdel:
Confirmed this is now behaving consistently :) `yarn ; yarn start` .. it works :), `rm -rf node_modules yarn.lock ; yarn ; yarn start` .. it still works :).

Also, fixes the bug .. merging :)

PR #1410, opened 2018-03-21
Fixes disk size misreporting for some disks

AllenBW:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1561160
(Resource details view) Some disks total size was being misreported in the tooltip.

### before
screen shot 2018-03-21 at 8 34 40 am

### after
screen shot 2018-03-21 at 8 34 09 am


AllenBW:
cc @nachandr 🙇 ❤️ 🎿

PR #1411, opened 2018-03-26
AuthenticationAPI - use base64encode which can deal with UTF8 passwords

himdel:
This is the same change as https://github.com/ManageIQ/manageiq-ui-classic/pull/3682, except for SUI.

---

This adds 2 npm packages:

* `base64js` - JS implementation of base64 which can deal with arbitrary encodings
* `text-encoder-lite` - a shim for the TextEncoder api, useable for utf8 strings (even in IE)

and replaces the `AuthenticationAPI` use of `$base64.encode` with a custom `base64encode` function implemented using the 2 libraries.

This way, users can use passwords with unicode chars.

Note: on modern browsers, this will handle any unicode password, including emoji.
On IE, more modern unicode versions may not be fully supported (czech or japanese chars should be fine, emoji may not).

The difference:

```
btoa('babicka')
"YmFiaWNrYQ=="

base64encode("babicka")
"YmFiaWNrYQ=="

btoa('babička')
VM7832:1 Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.

base64encode("babička")
"YmFiacSNa2E="
```

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1527316


himdel:
Testing:

add a user with special (unicode) characters in the password
try to log in :)

AllenBW:
oh my, now those are annoying failures... lemme see what can be done to help out travis... 🤔

himdel:
So.. this does fail with PhantomJS launcher but not with Chrome launcher .. also, looks like this kind of `import`  is not really used in SUI, almost everything is `require`d in `client/app.js`.

Also also, the specs kinda don't even run... looks like a runtime error early on, before any of the real code gets run (otherwise we'd see the promise either resolve, or catch, but if I add a `.then(done, done)`, it still times out.

That leads me to think this could be some webpack config error (which we could see as not being able to login in older browsers) or a difference in how the specs are compiled (babel vs webpack?).

Either way, I'll try fixing in the morning by moving that import to `client/app.js`, and seeing if that works :).

himdel:
Never mind, this is actually a different problem .. `TextEncoder` works just fine (in browsers that support it), but `TextEncoderLite` doesn't seem to have an `encode` method we rely on.

himdel:
@AllenBW looks like this fixes the specs, finally :) PTAL :)

package.json
```diff
@@ -112,6 +112,7 @@
     "sprintf-js": "1.1.1",
     "standard-loader": "6.0.1",
     "style-loader": "0.20.2",
+    "text-encoder-lite": "git://github.com/coolaj86/TextEncoderLite.git#e1e031b",
```
AllenBW: DANG that's annoying... update the to the version we need... but don't release it, 1.0.2 looks like what we want... but its nowhere cept on the commit...

@himdel yah ok with rewording this as follows? `    "text-encoder-lite": "coolaj86/TextEncoderLite#e1e031b",`

AllenBW:
@simaishi gooooood to get back! 👍 🌮

simaishi:

```
$ git log -1
commit 98c826b38d0982221658923c5f6ee4392294e571
Date:   Wed Mar 28 10:00:30 2018 -0400

    Merge pull request #1411 from himdel/password-utf-bz1527316
    
    AuthenticationAPI - use base64encode which can deal with UTF8 passwords
    (cherry picked from commit 0ae3d013fe0c6211080adb28f9c8130ba3ed1bdc)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1562797
```

himdel:
Another BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1579501

simaishi:
Fine backport details:

```
$ git log -1
commit 577c5533bdc412df10b2730c283253f36ff658ae
Date:   Wed Mar 28 10:00:30 2018 -0400

    Merge pull request #1411 from himdel/password-utf-bz1527316
    
    AuthenticationAPI - use base64encode which can deal with UTF8 passwords
    (cherry picked from commit 0ae3d013fe0c6211080adb28f9c8130ba3ed1bdc)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1562798
```

simaishi:
@himdel Can you please take a look at Travis failure in Fine branch?

https://travis-ci.org/ManageIQ/manageiq-ui-service/jobs/380680697

himdel:
Aah, looks like the failure is just linters .. doing a FINE PR to fix...

PR #1417, opened 2018-04-03
Refactors ansible credentials to be more generic, override with dialog input

AllenBW:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1557504
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1540689

Also displays any config_info credential object with 'credential' in the key

This work follows the following assumptions: 
1. dialog objet will always over config info - for provision does not exist in reirement
2. dialog_credientail will always override credential_id
3. from the dialog object, to the config_info -  dialog_cloud_credential -> (normalized) cloud_credential - will override cloud_credential
4. when avaliable, all possible credential types will have "credentail" in the object key, and be contained in config info (if not overridden)

cc @bzwei - yer a flippin 🧙‍♂️ ❤️

### DIFF
--- a/client/app/services/service-details/service-details-ansible.component.js
+++ b/client/app/services/service-details/service-details-ansible.component.js
@@ -37,7 +37,6 @@ function ComponentController (ModalService, ServicesState, lodash) {
 
   function fetchResources () {
     vm.loading = true
-    const credentialTypes = ['credential_id', 'network_credential_id', 'machine_credential_id', 'cloud_credential_id']
 
     if (angular.isDefined(vm.service.options.config_info)) {
       vm.orcStacks = {}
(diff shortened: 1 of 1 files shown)


AllenBW:
cc @kedark3 gonna wanta work with you to ensure this works as intended before merging (but maybe tomorrow 😏 )

kedark3:
@AllenBW let me know if you want me to review. Happy to help!

client/app/services/service-details/service-details-ansible.component.js
```diff
@@ -106,4 +112,15 @@ function ComponentController (ModalService, ServicesState, lodash) {
...
+
+  function processCredential (item) {
+    item.type = item.type.substring(item.type.lastIndexOf('::') + 2, item.type.lastIndexOf('Credential'))
+    return item
+  }
+
+  function normalizeDialogKeys (key) {
+    const normalizedKey = key.replace(/dialog_/i, '')
+
+    return normalizedKey === 'credential' ? 'credential_id' : normalizedKey
```
bzwei: To include other credential type, should it be something like
```
  return normalizedKey.contains('credential') ? normalizedKey+'_id' : normalizedKey
```

AllenBW: @bzwei if at dialog has a credential override of `dialog_cloud_credential_id` the change you are proposing would result in a normalized key of `cloud_credential_id_id`

This conditional is to catch the edge case of `dialog_credential` which would be normalized to `credential` which would not override the desired machine credential which is called `credential_id`

AllenBW:
@kedark3 was gracious enough to make a machine, looks like we're lookin' gooooood!!!

screen shot 2018-04-04 at 11 05 02 am

🌮 💃

client/app/services/service-details/service-details-ansible.component.js
AllenBW: >  Since you only need to look for one and only credential, you do not need to use a map.

unclear watcha mean by this ☝️ 😕

bzwei: I thought you use a map from `credential` to value. In this map it can be only one pair. why bother?

AllenBW: In the case of `provision ` and `dialog` present, we map through each `dialog` keys, normalize them and assign them to the config_info object. Without this we miss all overrides 😬

simaishi:

```
$ git log -1
commit 91a990edc68d01ec8d8b1fa50c885091f4b4daba
Date:   Thu Apr 5 16:13:29 2018 +0200

    Merge pull request #1417 from AllenBW/BZ/#1557504-ansible-credentails
    
    Refactors ansible credentials to be more generic, override with dialog input
    (cherry picked from commit 9eeaab534f422a4f5bb2b3517f689dd5597ae127)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1565160
```

simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq-ui-service/pull/1418

PR #1425, opened 2018-04-23
Add .yamllint from provider generator

jrafanie:
Add .yamllint from provider generator


AllenBW:
@jrafanie no .yaml here, yah ok with this being closed?

PR #1446, opened 2018-06-19
BZ#1589409 - Support for disabling Dashboard

chalettu:
@miq-bot add_label bug

### DIFF
--- a/client/app/components/dashboard/dashboard.component.js
+++ b/client/app/components/dashboard/dashboard.component.js
@@ -18,7 +18,7 @@ function ComponentController ($state, DashboardService, EventNotifications, loda
   }
   vm.$onInit = function () {
     vm.permissions = {
-      'monthlyCharges': RBAC.has(RBAC.FEATURES.DASHBOARD.VIEW.MONTHLY_CHARGES)
+      'monthlyCharges': RBAC.has(RBAC.FEATURES.DASHBOARD.MONTHLY_CHARGES)
     }
 
     angular.extend(vm, {
--- a/client/app/core/authorization.config.js
+++ b/client/app/core/authorization.config.js
@@ -58,7 +58,11 @@ export function authInit ($rootScope, $state, $log, Session, $sessionStorage, La
     if (angular.isDefined(toState.data)) {
       if (angular.isDefined(toState.data.authorization) && !toState.data.authorization) {
         event.preventDefault()
-        $state.transitionTo('dashboard')
+        if (toState.name === 'dashboard' && RBAC.has(RBAC.FEATURES.SERVICES.VIEW)) {
+          $state.transitionTo('services')
+        } else {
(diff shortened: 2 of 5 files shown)


chalettu:
Screenshot of what it looks like if you don't have access to the Dashboard.  
demo_screenshot

AllenBW:
closes https://bugzilla.redhat.com/show_bug.cgi?id=1589409

serenamarie125:
I'm not sure of the RBAC model for the SUI, but I wonder if utilizing something like this might make more sense than only handling a single use case:
- if user access to 1 or more of the nav items, navigate to the first navigation item by default
- if user access to none of the nav items, don't let them in ( i think we previously handled this case )

@chalettu @AllenBW @Loicavenel @ohadlevy thoughts?

serenamarie125:
I think that makes sense @chalettu but you should def get the 👍from either Ohad or Loic first!

AllenBW:
Oh oh yeah so if a user can view say, the service catalog, and *ONLY* the service catalog, thats the page they should be pushed to 😋

chalettu:
Updated the logic to go down the list of pages a user has access to and pick the firsts one they can access and sends them there .  So if a person tries to go to a page they do not have access to, they will be redirected to the first menu item they do have access to.

AllenBW:
So I jumped in with a few users, ones that only had the service catalog role enabled then my orders enabled, the role looking a lil like this...

screen shot 2018-06-20 at 8 07 08 am


And for both users, when attempting to login, ended up with this...
screen shot 2018-06-20 at 8 07 52 am

chalettu:
@AllenBW , I tried reproducing the error you are seeing with the role you had pictured above and  can't seem to reproduce it.  Did you close your browser out and try fresh after saving the role changes?  We cache permissions so if you didnt log out first of SUI, sometimes it causes weird things to happen.

AllenBW:
@chalettu yeah tried it across a number of browsers, bounced the miq instance between changes, same result. Gonna do a box bounce and see if that changes anything 🤷‍♂️

AllenBW:
Yeah bounced the box, still see it, furthermore this is what yah get when working from master, only have the "Service Catalog" and "Desktop" lnav tab
screen shot 2018-06-20 at 11 08 23 am

simaishi:

```
$ git log -1
commit 307c96833fe66b7bfccb66ea7ced8c767182b412
Date:   Wed Jun 20 11:27:49 2018 -0400

    Merge pull request #1446 from chalettu/dashboard-permissions
    
    BZ#1589409 - Support for disabling Dashboard
    (cherry picked from commit ae385823137b78b78b1ea6b8abdee77b0b9b2ac4)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1595454
```

PR #1451, opened 2018-07-05
BZ#1597393 - VMware VNC Remote Console does not work in SSUI

chalettu:
@miq-bot add_label bug


chalettu:
Screenshot of the fix in action.  It would error out without the change implemented in this PR 
console_demo

simaishi:
@chalettu @AllenBW As per BZ, this is not an issue in 5.9.z, but this PR has `gaprindashvili/yes` label. Can you please clarify?

AllenBW:
@chalettu can you weigh in here? BZ doesn't have the flag indicating this ought to be backported... should we we remove `gaprindashvili/yes`?

chalettu:
@miq-bot add_label gaprindashvili/no

PR #1454, opened 2018-07-12
Fix for refresh of dialog fields not being called in some circumstances

eclarizio:
The root cause is that we were defaulting to use the first `resourceAction` as long as the `action` was `'Provision'`. However, in this case, `serviceTemplate.resourceActions` returned two, and the `action` property for the first one was `'Retirement'` and the second was `'Provision'`. However, since we were only checking the `action` for the first in the list, the id was ending up as an empty string and thus the API call was simply failing in the backend.

So, this piece of code now finds the correct provision `resourceAction`.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1598475

@miq-bot add_label gaprindashvili/yes, bug, blocker
/cc @tinaafitz , @d-m-u 

@d-m-u Can you review?

@dclarizio Not sure who this should be assigned to, haven't done service-ui work in a while, please advise!

### DIFF
--- a/client/app/states/catalogs/details/details.state.js
+++ b/client/app/states/catalogs/details/details.state.js
@@ -125,10 +125,7 @@ function Controller ($stateParams, CollectionsApi, EventNotifications, ShoppingC
  */
   function refreshField (field) {
     const resourceActions = vm.serviceTemplate.resource_actions
-    let resourceActionId = ''
-    if (resourceActions.length > 0 && resourceActions[0].action === 'Provision') {
-      resourceActionId = resourceActions[0].id
-    }
+    let resourceActionId = lodash.find(resourceActions, ['action', 'Provision']).id
 
     let idList = {
       dialogId: vm.parsedDialogs[0].id,
--- a/client/app/states/catalogs/details/details.state.spec.js
+++ b/client/app/states/catalogs/details/details.state.spec.js
@@ -58,20 +58,14 @@ describe('State: catalogs.details', () => {
         controller = $controller($state.get('catalogs.details').controller)
         expect(controller).to.be.defined
       })
-      it('it allows a field to be refreshed', (done) => {
-        controller = $controller($state.get('catalogs.details').controller)
-        controller.serviceTemplate = serviceTemplate
-        controller.setDialogUrl()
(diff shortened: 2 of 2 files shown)


dclarizio:
@AllenBW or @chalettu can you please review and look at the travis failure?  @eclarizio doesn't think the travis failure has anything to do with the code changes.  Thx, Dan

chalettu:
So one issue I see is that we have dialogs that are used in 3 places in Service UI and this PR seems to only cover 1 of those places.  The other two places are custom button detail page and also service reconfigure page.  
Here is the custom button detail page
https://github.com/ManageIQ/manageiq-ui-service/blob/master/client/app/states/services/custom_button_details/custom_button_details.state.js
Here is the reconfigure service page
https://github.com/ManageIQ/manageiq-ui-service/blob/master/client/app/states/services/reconfigure/reconfigure.state.js

Please take a look and ensure that this change if relevant is applied to these pages as well.

AllenBW:
@dclarizio, Travis sometimes fails, just gotta give it a kick, looks good now

chalettu:
@AllenBW , I kicked the job back off and it seems to be going ok.

eclarizio:
@chalettu I think we're good to go here, those two places are unrelated to this change.

AllenBW:
@eclarizio if resourceActions is null, does looking for .id blow this up?

eclarizio:
@AllenBW You mean if the `lodash.find` ends up returning undefined? Theoretically, yes, but I don't think there's any instance where a service template would end up not having a `'Provision'` resource action. If there is, then I think we'd have more problems (backend-wise) than just a missing ID, haha.

simaishi:

```
$ git log -1
commit 4818c77231c790d1127e5c789e5b5fb91df3eea5
Date:   Thu Jul 12 16:50:24 2018 -0400

    Merge pull request #1454 from eclarizio/BZ1598475
    
    Fix for refresh of dialog fields not being called in some circumstances
    (cherry picked from commit 8bbdd5e40637edf06ff6f4293704e46fcbb0602a)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1600738
```

PR #1458, opened 2018-07-16
[GAPRINDASHVILI] BZ#1601587 - Fixed issue with dialogs that causes duplicate values

chalettu:
@miq-bot add_label bug

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@angular/platform-browser-dynamic": "4.4.6",
     "@angular/upgrade": "4.4.6",
     "@manageiq/font-fabulous": "1.0.0",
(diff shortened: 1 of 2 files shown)


chalettu:
This is the SUI implementation that fixes the bug addressed here, https://github.com/ManageIQ/ui-components/pull/316

simaishi:
@chalettu @AllenBW Are we not updating `yarn.lock`?

AllenBW:
Ooooo good point, we ought to update that as well, @chalettu yah want me to drop in?

chalettu:
Repushed branch with the yarn.lock file updates

PR #1462, opened 2018-07-18
BZ#1602108 - Changed Retire services to a modal instead of confirmation

chalettu:

### DIFF
--- a/client/app/services/services-state.service.js
+++ b/client/app/services/services-state.service.js
@@ -156,13 +156,6 @@ export function ServicesStateFactory (ListConfiguration, CollectionsApi, RBAC) {
           actionName: 'retireService',
           actionFn: retireServiceFn,
           icon: clockIcon,
-          showConfirmation: true,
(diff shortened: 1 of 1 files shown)


chalettu:
Screenshot of update
demo_retire_again

chalettu:
Fixes BZ https://bugzilla.redhat.com/show_bug.cgi?id=1602108

AllenBW:
@chalettu would yah mind posting what it look like when we have 20 items selected? 

Just wanta make sure we're not creating more problems than we're looking to solve...

chalettu:
Here is a demo of what it looks like when selecting 50 services at once to retire
demo_2

chalettu:
@serenamarie125 , do you think this UI interaction for this modal is appropriate.  When you try to retire a lot of services at once in SUI, we list them all and make them scrollable.  Do you think we should have a threshold where we should change this?  Let us know your thoughts.

PR #1474, opened 2018-09-12
BZ#1624573 - Fix issue where error message is not displayed for consoles

chalettu:
Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1624573.   Error messages not being displayed properly for console launch errors. 
@miq-bot add_label bug

### DIFF
--- a/client/app/services/consoles.service.js
+++ b/client/app/services/consoles.service.js
@@ -47,8 +47,15 @@ export function ConsolesFactory ($window, CollectionsApi, $timeout, $location, E
     }
   }
 
-  function consoleError (error) {
-    EventNotifications.error(__('There was an error opening the console. ' + error.message))
+  function consoleError (response) {
+    let message = __('undefined')
+    if (response.data) {
+      message = response.data.error.message
+    } else if (response.message) {
+      message = response.message
+    }
+
+    EventNotifications.error(__(`There was an error opening the console. ${message}`))
   }
 
   function consoleOpen (results) {


chalettu:
Before SS
before

After SS
after

chalettu:
Thanks @AllenBW .  Seems this one caused a test to fail.  Working on fixing it now.

client/app/services/consoles.service.js
```diff
@@ -47,8 +47,15 @@ export function ConsolesFactory ($window, CollectionsApi, $timeout, $location, E
     }
   }
 
-  function consoleError (error) {
-    EventNotifications.error(__('There was an error opening the console. ' + error.message))
+  function consoleError (response) {
+    let message = __('undefined')
+    if (response.data) {
+      message = response.data.error.message
+    } else if (response.message) {
```
AllenBW: When would this case happen (the response.message)? I thought error messages were always of the `response.data.error.message` format?

chalettu: I dug through our test cases and it seems that in the use case that the response was a 200 and status was != success .  I am assuming the test cases were accurate as I didn't have a way to forcibly replicate this http response scenario.  This change I made supports that scenario and seems harmless enough in the case the scenario doesn't happen.

AllenBW: Hokay, yeah no harm, merging.

JPrause:
@miq-bot add_label blocker

PR #1497, opened 2018-11-16
BZ#1650693 If tagging is readonly we should not show the edit dropdown

AllenBW:
But if tagging is not read only, behavior is nominal

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1650693

### see, no more edit dropdown here (where we shouldn't have one)
screen shot 2018-11-16 at 4 59 04 pm

### but we can still do tagging where we need to !
screen shot 2018-11-16 at 4 59 16 pm

### DIFF
--- a/client/app/shared/tagging/tagging.component.js
+++ b/client/app/shared/tagging/tagging.component.js
@@ -101,7 +101,7 @@ function TaggingController ($scope, $filter, $q, $log, CollectionsApi, TaggingSe
     if (vm.tags.filtered) {
       vm.tags.filtered[0] = placeholderCategorization
       vm.tags.selectedTag = vm.tags.filtered[0]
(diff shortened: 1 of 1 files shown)


simaishi:
Hammer backport details:

```
$ git log -1
commit af29bdaaf19c57998565b1bf6a4f69fc13fe6ee5
Date:   Mon Nov 19 16:31:43 2018 +0100

    Merge pull request #1497 from AllenBW/BZ/#1650693-fix-tagging-readonly-issue
    
    BZ#1650693 If tagging is readonly we should not show the edit dropdown
    
    (cherry picked from commit 953635b5b8a21387f775f94af6b6c325200992de)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1650693
```

PR #1500, opened 2018-11-27
 BZ#1525883 CFME SSUI My Orders - Oder page has untranslated entries

AllenBW:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1525883

PR #1501, opened 2018-11-27
Update .pot on master to capture a number of untranslated strings

AllenBW:
a follow-on to #1500, pot file on master isn't up to date (hammer .pot is current to its code)


himdel:
Cc @mzazrivec 

(LGTM, but I'm not sure what the process with pot files is.)

PR #1508, opened 2018-12-05
[HAMMER] Switch zh-Hans-CN translation identification string to delimiting dashes

AllenBW:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1655584

Looks like when we pull down translations `-` are changing to `_` which breaks this translation, but no longer! Though something to watch out for in the future...


himdel:
Aah, good catch but the wrong place to fix it.

The problem is that rails is using a different dash/underscore convention than .. everybody else, so we're manually transforming the locale string coming from rails.

And that code likely never expected 2 of those in the same locale.

(Or, to put it differently, `zh_CN` is a valid locale, `zh-Hans-CN` is *something*.)

mzazrivec:
https://github.com/ManageIQ/manageiq-ui-service/commit/2f356c85db60604ac26ccc711f8687b03d91ce83

himdel:
The problem:

```
jq 'keys' client/gettext/json/manageiq-ui-service.json > locales
jq 'keys' client/gettext/json/available_languages.json > available

diff -Naur locales available
```

```diff
--- locales	2018-12-05 13:45:39.914921612 +0000
+++ available	2018-12-05 13:45:43.606947052 +0000
@@ -5,8 +5,8 @@
   "ja",
   "ko",
   "nl",
+  "pt-BR",
   "pt-PT",
-  "pt_BR",
-  "zh-TW",
-  "zh_Hans_CN"
+  "zh-Hans-CN",
+  "zh-TW"
 ]
```

(Whereas on master, the diff is empty.)

mzazrivec:
So we just need to update `available_languages.json`, correct?

himdel:
And `client/gettext/json/available_languages.json` is an auto-generated file... `config/available-languages.js`.

So.. you should be able to run `yarn run available-languages` to get this fixed.

AllenBW:
Oh, darn, saw the bz comment before this, doing doing...

AllenBW:
hmmm running `yarn available-languages`  generates 

```
{"es-ES":"","es":"locale_name","fr":"Français","ja":"locale_name","ko":"","nl":"","pt_BR":"Português (Brasil)","pt-PT":"","zh_Hans_CN":"locale_name","zh-TW":""}
``` 
which for sure isn't gonna work

himdel:
Oh well :(.

@mzazrivec will have to fix that.

Every single locale must have a `locale_name` string that translates to the name of that locale. That's used by the script to generate those.

Seems like half the language don't have that string at all, and the other half translates it as `locale_name` .

AllenBW:
This is because translators didn't include the locale_name in the translation files, most of the time `locale_name` was entered as the translation 😭

himdel:
Yup, exactly :(

mzazrivec:
I just updated / fixed `locale_name` in Zanata for `es fr ja pt-BT zh-Hans-CN` locales (hammer branch).

mzazrivec:
So I think what needs to be done is download latest po catalogs from Zanata (with my changes), compile them (into json format) and update `available_languages.json`.

AllenBW:
closing in favor of #1509

PR #1509, opened 2018-12-05
[HAMMER] Fix locale names for hammer

mzazrivec:
This PR:
* contains updated catalogs with fixed `locale_name`
* contains compiled catalogs
* contains updated `available_languages.json`

@AllenBW @himdel 

https://bugzilla.redhat.com/show_bug.cgi?id=1655584


mzazrivec:
@himdel @AllenBW for some reason, switching the UI locale to any of `es fr ja` works correctly. Switching to either `pt-BT` or `zh-Hans-CN` fails :-/

himdel:
> Switching to either pt-BT or zh-Hans-CN fails :-/

Makes sense, both still use underscores in `available_languages` but dashes in the translation.
Need to make sure the keys match.

mzazrivec:
> > Switching to either pt-BT or zh-Hans-CN fails :-/
> 
> Makes sense, both still use underscores in `available_languages` but dashes in the translation.
> Need to make sure the keys match.

`available_languages.json` contains `pt_BR`, so does `manageiq-ui-service.json` (the file that matters here). So what needs to match with what here?

AllenBW:
maybe we need to 🔪 https://github.com/ManageIQ/manageiq-ui-service/commit/2f356c85db60604ac26ccc711f8687b03d91ce83 ?

himdel:
Aah, yeah, exactly... it will only match if both are dashes. If both are underscores, that code breaks it.

mzazrivec:
The true fix here was to fix language codes in the `.po` catalogs: from `pt_BR` to `pt-BR` and from `zh_Hans_CN` to `zh-Hans-CN`.

Now things should work correctly.

PR #296, opened 2016-10-28
Enhancement/251 migrate from bower to npm

AllenBW:
☠️  bower, long live bower 👑 

Re-configures process for adding dependencies, need to add documentation on this, was thinking of starting a `guides` folder in where I can explain the process (long story short npm install, add fep location in the stylesheet.html or javascript.html, update test config dependencies to reflect)

Renames a number of gulp tasks to better align with actual function
Removes a number of unnecessary gulp tasks relating to building the test environment

End of the day, not much changed here! Though I would love some 👀 from the following :godmode: :
@chriskacerguis @himdel @jeff-phillips-18 @dtaylor113 

closes #251

PR #1517, opened 2019-01-04
Drop /server, and dependencies, not used anymore

himdel:
`server/app.js` is not being used anymore, since `yarn run start` just runs `webpack-dev-server` instead.

This seems to have been the case since somewhere between https://github.com/ManageIQ/manageiq-ui-service/pull/296 (drops the `start` script running `server/app.js`) and https://github.com/ManageIQ/manageiq-ui-service/pull/415 (adds a `start` script running `webpack-dev-server`).

Removing the old server, and its dependencies... and updating the skin-sample to be able to live without `server/`.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -16,7 +16,8 @@ If you would like to run unit tests for the project, you can run
 If you would like to run the integration tests locally for the project please be sure to first have a running copy of the frontend and a running backend server.  You can either use a ManageIQ backend or test with our Mock API running locally.  These are the commands to launch the frontend and our mock api
   
 ``` yarn start ```  (Starts a frontend web server - Port 3001)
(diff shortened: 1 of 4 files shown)


AllenBW:
LOVE THIS!!! less is more for sure @himdel well the only real place [it is being used is here](https://github.com/ManageIQ/manageiq-ui-service/blob/1d69877e0e4195830b3f81fbbad501338f7679f7/package.json#L24-L25)

but we could 🔨 those two, cuz no one uses `start:all` could also get rid of `start:mock-api` 🤔

himdel:
Ooh, I missed those, sorry O:-).

Ok, so, you're positive nobody is using those? I can just drop them too...

himdel:
..and done :)

PR #1513, opened 2018-12-18
BZ#1471301 - Login/nav brand and logo are set by api

AllenBW:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1471301

In dev, these things will appear as broken images, but upon build, they show up just fine

Styling on login is a little jacked up atm... but we're applying the same classes so 🤷‍♂️ 

### What we look like (note the url `ui/service`)
screen shot 2018-12-18 at 3 16 47 pm
screen shot 2018-12-18 at 3 16 34 pm

PR #1521, opened 2019-01-30
Support custom favicons

himdel:
Follow up to https://github.com/ManageIQ/manageiq-ui-service/pull/1513 which made it possible to use the same (possibly custom) brand & logo as the ops ui,
this adds the same support for favicons.

Depends on https://github.com/ManageIQ/manageiq-api/pull/551 => `needs api`.

Using `$rootScope.favicon` because `<head>` is outside any controllers.

Cc @AllenBW , @skateman 

https://bugzilla.redhat.com/show_bug.cgi?id=1578076


himdel:
For testing:

`ln -s ~/manageiq/public/upload ~/manageiq-ui-service/client/assets/` 
creates a symlink that allows devel-mode SUI to see custom images.

(Won't work for non-custom in development, because hashes.)

AllenBW:
yup and there it is, a glorious favicon 🤗 
screen shot 2019-01-31 at 12 56 56 pm

PR #1524, opened 2019-02-01
Suspended power_state is "suspended"

AllenBW:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1670373

### look now its all fixed n stuff, you can start a service that is of an unknown or suspended power state
screen shot 2019-02-01 at 8 29 38 am

### DIFF
--- a/client/app/services/poweroperations.service.js
+++ b/client/app/services/poweroperations.service.js
@@ -92,7 +92,7 @@ export function PowerOperationsFactory (CollectionsApi, EventNotifications, spri
   }
 
   function powerOperationSuspendState (item) {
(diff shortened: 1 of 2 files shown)


himdel:
This changed in https://github.com/ManageIQ/manageiq-ui-service/pull/430,

```diff
-      return item.power_state === "off" && item.power_status === "suspend_complete";
+      return getPowerState(item) === 'off';
```

and I'm not seeing any `suspended` in manageiq core.


Are you sure the state is really consistently `suspended` for all the providers, or should we check `power_status` as well?

PR #1526, opened 2019-02-11
[GAPRINDASHVILI] NavigationController - regenerate menu when language changed

himdel:
This is *technically* a backport of https://github.com/ManageIQ/manageiq-ui-service/pull/1375,
except most of the PR had nothing to do with the bug.

---

When the language changes, we need to call `Navigation.init` again, to popule the menu items with the newly translated values.

This would happen on login, but not after reloading the page.
(Because Navigation.init is first called before the language gets set.)

https://bugzilla.redhat.com/show_bug.cgi?id=1501098
https://bugzilla.redhat.com/show_bug.cgi?id=1672324


simaishi:
We're no longer building gaprindashvili, so closing this PR.

PR #1530, opened 2019-03-20
Speed up service_dialog initialization

KarinaTiurina:
This speeds up dialog initialization with a big number of dynamic fields. 

**How it works:**

1. Gets empty dialog with only static fields initialized (API PR: https://github.com/ManageIQ/manageiq-api/pull/567)
2. Creates hash `dynamicFieldsSummary` with the following keys: 
  `allDynamicFieldsToInit ` - all dynamic fields that should be initialized
  `allFieldResponders` - all dependent fields

3. Sends api requests for each dynamic field in `allDynamicFieldsToInit` apart from dependent fields
4. Updates `dynamicFieldsSummary`: removes all initialized fields from `allDynamicFieldsToInit` and updates dependent fields
5. Repeats 3-4 until `allFieldResponders` is empty (`initDialogFields` is a recursive function)

- - - - - - - 

It works with multiple dialogs and tabs and allows to set fields in any order

### DIFF
--- a/client/app/states/catalogs/details/details.state.js
+++ b/client/app/states/catalogs/details/details.state.js
@@ -75,7 +75,6 @@ function Controller ($stateParams, CollectionsApi, EventNotifications, ShoppingC
       const DIALOGS_RESPONSE = 0
       dialogs = data[DIALOGS_RESPONSE]
       vm.serviceTemplate = data[SERVICE_TEMPLATE_RESPONSE]
-      vm.serviceTemplate.long_description = parseDescription(vm.serviceTemplate)
       setDialogUrl()
 
       if (dialogs.subcount > 0) {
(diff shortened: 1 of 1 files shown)


himdel:
Cc @eclarizio I guess you'll want to take a look :) (also https://github.com/ManageIQ/manageiq-api/pull/567)

himdel:
Added `pending core`, since this depends on the API PR.

@eclarizio Aah, I missed that, yes, a similar change would have to happen in ui-classic as well, and maybe we could refactor the edges a bit, to make it possible to share more of the loading logic between the repos.

@KarinaTiurina is that something you'd be interested in doing?

martinpovolny:
still waiting for https://github.com/ManageIQ/manageiq-api/pull/567

himdel:
This is already achievable by setting any individual dynamic field's "Load values on init" to false.

Unless anybody is willing to take on the work, tentatively closing.

PR #1531, opened 2019-04-03
Fix centos image filename

himdel:
have a service with an associated *centos* vm,
open the service in My Services,
click on the VM detail,
an icon is missing in the leftmost card.

The file was named `os-centos.svg`, but `vm.os_image_name` says `linux_centos`, so we need `os-linux_centos.svg`

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1687059

---

Before:

before

After:

now


himdel:
Reproducing in rails console:

find a centos vm: `vm = Vm.all.find { |v| v.os_image_name == 'linux_centos' }`
find a service: `s = Service.first`
associate: `s.add_resource(vm); s.save!`

Open `/services/#{s.id}/resource=#{vm.id}`

PR #1533, opened 2019-04-03
Services - add "Select All" checkbox

himdel:
Added support for a Select All checkbox to ExplorerPagination:

    check-all="{ checked: false, trigger: function(newValue)... }"

And enabled for the My Services screen.
Styled identical to the ui-classic version of the same.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1498996

---

checkall2

### DIFF
--- a/client/app/shared/pagination/pagination.html
+++ b/client/app/shared/pagination/pagination.html
@@ -1,5 +1,5 @@
 <div class="pagination-footer">
-  <span class="pull-left checkbox span-right-border" style="margin-top: 2px; margin-left: 18px" ng-if="vm.checkAll">
+  <span class="pull-left checkbox span-right-border checkall-margin-override" ng-if="vm.checkAll">
     <label class="checkbox-inline">
       <input type="checkbox" ng-model="vm.checkAllState" ng-change="vm.onCheckAll()">
       <span translate>
--- a/client/app/shared/pagination/pagination.html
+++ b/client/app/shared/pagination/pagination.html
@@ -1,5 +1,5 @@
 <div class="pagination-footer">
-  <span class="pull-left checkbox span-right-border" style="margin-top: 2px" ng-if="vm.checkAll">
+  <span class="pull-left checkbox span-right-border" style="margin-top: 2px; margin-left: 18px" ng-if="vm.checkAll">
     <label class="checkbox-inline">
       <input type="checkbox" ng-model="vm.checkAllState" ng-change="vm.onCheckAll()">
       <span translate>
--- a/client/app/shared/pagination/pagination.html
+++ b/client/app/shared/pagination/pagination.html
@@ -1,4 +1,12 @@
 <div class="pagination-footer">
+  <span class="pull-left checkbox span-right-border" style="margin-top: 2px" ng-if="vm.checkAll">
+    <label class="checkbox-inline">
+      <input type="checkbox" ng-model="vm.checkAllState" ng-change="vm.onCheckAll()">
+      <span translate>
+        Select All
+      </span>
+    </label>
+  </span>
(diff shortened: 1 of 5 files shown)


terezanovotna:
@himdel I know there is not much pf documentation on this topic, but could we move the Select All to the toolbar area? it seems odd having it below the list view area.

himdel:
@terezanovotna Probably .. can you provide a screenshot please? (or a mock or something)

I'm not sure what you call a toolbar... the only one I see is on the top, with filters, and that doesn't sound like a likely place to find a check all checkbox, is it?

(I mean, left of the filter sounds wrong, right of the filter sounds just surprising... is that where you meant?)

terezanovotna:
I found the [right link!](https://www.patternfly.org/pattern-library/forms-and-controls/bulk-selector/).


This is the toolbar:
Screen Shot 2019-04-03 at 15 04 50


We have two options
1) either have a checkbox in the toolbar
Screen Shot 2019-04-03 at 15 04 43

or

2)have Select All (as you suggested) in the toolbar
Screen Shot 2019-04-03 at 15 04 34


My main point is to move it to the toolbar, feel free to pick whatever option you want.

himdel:
Ugh, ok then. Thanks! :)

(This would have been very nice to have before, this pretty much means closing this PR and rewriting a different component for that.)

terezanovotna:
Sorry, feel free to ping me next time, will try to get back to you sooner.

himdel:
So...

the bulk selector pattern ... according to https://www.patternfly.org/pattern-library/forms-and-controls/bulk-selector/, there is no implementation, only a design. That's useless.

We're using the toolbar from angular-patternfly. According to https://patternfly.github.io/angular-patternfly/#!/api/patternfly.toolbars.directive:pfToolbar , there is no support for adding custom elements to the toolbar, or to add the bulk selector.

So... we would need to rewrite the toolbar component to put the checkbox in the suggested place.


Cc @h-kataria @martinpovolny @terezanovotna 

Frankly I don't see a point in rewriting the whole toolbar handling in SUI just to add a single checkbox.

Can we live with the existing design instead? (As shown in this PR's description.)

himdel:
@miq-bot remove_label wip

Removing wip, if the current design can be used, this is done.
If it can't, I would probably look into using @martinpovolny react toolbar code in SUI.

martinpovolny:
> So... we would need to rewrite the toolbar component to put the checkbox in the suggested place.

No. We surely do not want to rewrite the toolbar. Not in Angular. That's a wasted effort.

We have to use what is available in Angular. If there are any significant requirements, we need to talk to @Loicavenel about merging SUI back to Classic.

@terezanovotna, @himdel : please, find a solution that is acceptable by UX and doable with the component as we have it.

terezanovotna:
consulted with @himdel and we are not going to rewrite the toolbar. we agreed the select all checkbox will be located in the left side of the footer.

himdel:
Updated the alignment so that the Select All checkbox is aligned under all the other checkboxes, not slightly to the left.

blee

client/app/shared/pagination/pagination.html
```diff
@@ -1,4 +1,12 @@
 <div class="pagination-footer">
+  <span class="pull-left checkbox span-right-border" style="margin-top: 2px; margin-left: 18px" ng-if="vm.checkAll">
```
AllenBW: @himdel any chance we can kill the inline styles for a class (`.checkallMarginOverride`? ) in _pagination.sass ?

himdel: Yeah, makes sense :)

himdel: Should be ready :)

PR #1542, opened 2019-04-30
VM custom button dialog - use the VM instead of the parent service

himdel:
Triggering a VM-based custom button on a Service's VM Resource causes the right dialog to appear..

But the inital data assumes the associated resource is the service, not the VM.
Subsequent field refresh requests and submit already use the VM, but the initial load was missed in https://github.com/ManageIQ/manageiq-ui-service/pull/1022.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1687061

### DIFF
--- a/client/app/states/services/custom_button_details/custom_button_details.state.spec.js
+++ b/client/app/states/services/custom_button_details/custom_button_details.state.spec.js
@@ -1,7 +1,10 @@
 /* global $state, $controller, CollectionsApi, Notifications, DialogFieldRefresh */
 /* eslint-disable no-unused-expressions */
 describe('State: services.custom_button_details', () => {
-  let dialog, dialogData, dialogFields, button
+  const dialogId = 213
+  const serviceId = 123
+
+  let dialog, dialogData, dialogFields, button, dialogResponse
 
   beforeEach(() => {
     module('app.states')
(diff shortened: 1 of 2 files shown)


himdel:
Have a Custom Button for VMs, associated with a service, which uses the VM to populate dropdown options (see the bz for exact steps).

Go to My Services
find a service with at least 1 VM associated
open the service detail (/services/4)
open the vm detail (/services/4/resource=2)
toolbar kebab custom button (goes to /services/4/custom_button_details)

does get /api/service_dialogs/110?expand=resources&attributes=content&resource_action_id=6572&target_id=4&target_type=service
dropdown says Script Error

after this, it does
get /api/service_dialogs/110?expand=resources&attributes=content&resource_action_id=6572&target_id=2&target_type=vm

h-kataria:
@eclarizio do you have a setup to review/test this?

himdel:
Thanks @eclarizio, working on a spec now...

As for the environment, yeah, it's not exactly trivial, this should help:

So, steps to reproduce - setup:

```
Automation > Automate > Explorer
choose a class
tab Methods
toolbar Configuration > Add a new Method
type inline
name bz1687061
data:
  @vm = $evm.root['vm']
  dialog_hash = {}
  dialog_hash[@vm.id] = @vm.name
  $evm.object['default_value'] = dialog_hash.first[0]
  $evm.object['values'] = dialog_hash
save

tab Instances
toolbar Configuration > Add a new Instance
name bz1687061
runcheck value: bz1687061

Automation > Automate > Customization
accordion Service Dialogs
toolbar Congfiguration > Add a new Dialog
name bz1687061
add a *Dropdown*, edit
  Field Information
    Dynamic Yes
  Options
    Entry Point: .../bz1687061
  save
save

accordion Buttons
create/choose a button group under "VM and Instance"
toolbar Configuration > Add a new Button
text bz1687061
dialog bz1687061
save

Have a service which is associated with a least 1 VM/Instance:
  (in rails console)
  s = Service.first
  v = Vm.first
  s.add_resource(v)
  s.save!
```

Then,

```
In ui-service:

My Services
open the service detail (/services/4/resource=2)
toolbar kebab custom button bz1687061 (goes to /services/4/custom_button_details)

does get /api/service_dialogs/110?expand=resources&attributes=content&resource_action_id=6572&target_id=4&target_type=service
dropdown says Script Error
```

himdel:
Tests should be ready :)

client/app/states/services/custom_button_details/custom_button_details.state.spec.js
```diff
@@ -159,48 +161,84 @@ describe('State: services.custom_button_details', () => {
...
-      const dialogResponse = {content: [dialog], id: 213}
-      const apiQueries = sinon.stub(CollectionsApi, 'get')
-      apiQueries.onFirstCall().returns(Promise.resolve(dialogResponse))
-      apiQueries.onSecondCall().returns(Promise.resolve({}))
+    const dialogResponse = {
+      content: [dialog],
+      id: 213
+    }
+
+    it('sends target_type=vm service dialog query on init', (done) => {
```
eclarizio: I think we're still missing the spec for when the `vmId` doesn't exist, right? There are still cases where the service dialogs endpoint is going to request with the `target_type` of `'service'` and the `target_id` with that service's id.

I guess maybe this spec didn't exist before, but I think it needs to be added now that we have these two different calls happening.

himdel: Good catch :+1:. The tests should probably be reorganized a bit to be more symmetrical, and looks like we have some SUI test inconsistencies (3 tests randomly in jest not karma), can't use object spread there, etc.

But for now, I just added a similar test for init *without* `vmId` :).

simaishi:
Hammer backport details:

```
$ git log -1
commit 5fe64d26c33c6e1af15b6194fd54cd2b268f4096
Author: Harpreet Kataria <hkataria@redhat.com>
Date:   Fri May 3 11:24:30 2019 -0400

    Merge pull request #1542 from himdel/bz1687061-custombutton
    
    VM custom button dialog - use the VM instead of the parent service
    
    (cherry picked from commit 61c583c5a7cf63d86a6c30166f029559e060e8f1)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1722817
```

PR #1546, opened 2019-05-31
Drop unused ui-select dependency

himdel:
this used to be needed for the dialog-user component,
but was later replaced by pf-select and later miq-select.

Unused since https://github.com/ManageIQ/ui-components/pull/249

(The dependency is also being removed from the other repos in https://github.com/ManageIQ/ui-components/pull/390 and https://github.com/ManageIQ/manageiq-ui-classic/pull/5607.)

### DIFF
--- a/client/app.js
+++ b/client/app.js
@@ -38,7 +38,6 @@ require('bootstrap-datepicker')
 require('angular-bootstrap-switch')
 require('angular-ui-sortable')
 require('bootstrap-combobox')
-require('ui-select')
 require('@manageiq/ui-components/dist/js/ui-components')
 
 // Libraries: angular-patternfly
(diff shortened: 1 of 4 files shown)


himdel:
(travis is red because we need to release the ui-components change first, and maybe explicitly bump the version)

AllenBW:
@himdel can we get a lockfile update with this pr?

himdel:
OK, merge https://github.com/ManageIQ/manageiq-ui-service/pull/1557 please and I'll update the lockfile after that? :)

himdel:
Never mind, there's no conflict, updated :).

himdel:
No problem, thanks :)

PR #1548, opened 2019-06-03
[WIP] Power state - support Suspended, show actual state for Unknown

himdel:
SUI is hardcoded to understand only 2 values of power_state: `on`, `off`
Everything else is "Unknown".

Adding an extra power_state: `suspended` ("Suspended")
Changing "Unknown" to "Unknown (actual value)"
Changing the icons for `on`, `off`, `unknown` and `suspended` to match [QuadiconHelper](https://github.com/ManageIQ/manageiq-decorators/blob/master/app/helpers/quadicon_helper.rb#L9)

~~Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1704226~~ (the BZ was closed WONTFIX)

---

Icons + "Unknown" tooltip - before:

before

After:

now

(TODO add suspened icon too)

### DIFF
--- a/client/app/services/poweroperations.service.js
+++ b/client/app/services/poweroperations.service.js
@@ -64,6 +64,8 @@ export function PowerOperationsFactory (CollectionsApi, EventNotifications, spri
         powerState = 'on'
       } else if (powerStatesMatch(item.power_states, 'off')) {
         powerState = 'off'
+      } else if (powerStatesMatch(item.power_states, 'suspended')) {
+        powerState = 'suspended'
       } else if (powerStatesMatch(item.power_states, 'timeout')) {
         powerState = 'timeout'
       }
--- a/client/app/services/service-details/service-details.html
+++ b/client/app/services/service-details/service-details.html
@@ -175,19 +175,24 @@
                                              tooltip-popup-delay="1000"
                                              tooltip-placement="bottom"></i>
                                           <span ng-if="!item.retired">
-                                          <i class="pficon fa fa-power-off" ng-if="item.power_state === 'off'"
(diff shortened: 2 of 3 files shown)


himdel:
TODO VM detail still wrong icon
maybe also use quadhelper colors

TODO2 use raw_power_state, power_state is unknown after setting raw_power_state to suspended

bekasovoleg:
Could you add states for powering on/off?

himdel:
@bekasovoleg Not with the current backend support.

If you can make it so that those states are consistently (for all providers) saved in the `power_state` field then absolutely.

But unfortunately, right now, this information is not available via the API.

bekasovoleg:
@himdel we have `power_status` attribute with `starting`, `stopping`, `suspending` values. Can we use this? This values  are used to block `Power On`, `Power Off` buttons. I guess it's possible to use this values.

himdel:
@bekasovoleg Sorry, I have no intention of implementing backend logic in SUI.

We have `power_state`, `power_states`, `raw_power_state` and `power_status`. 

So.. this really needs to happen on the backend, but if you can merge those into 1 place where we can always look, I'll update this PR (and ui-classic) accordingly :).

himdel:
@bekasovoleg (an alternative to that is that we would have the logic documented somewhere publicly - because I really know how to interpret each combination of those)

bekasovoleg:
@himdel, ок, thx, got it now

himdel:
(If we can expose QuadiconHelper from ui-classic via the api, this could be another alternative)

himdel:
This info *may* be available from the API after all, with `?attributes=normalized_state`

Issue #1520, opened 2019-01-29
SUI to support support webmks 


AllenBW:
Description of problem:
:point_up: [January 29, 2019 9:51 AM](https://gitter.im/ManageIQ/manageiq/providers?at=5c5068888aa5ca5abf4edda7)

https://bugzilla.redhat.com/show_bug.cgi?id=1532720

skateman:
So as the WebMKS console isn't shipped OOTB and we cannot precompile assets on an already existing appliance, it cannot be included into the asset pipeline nor a webpack pack (sounds funny, you seriously say it like this?). So I think we could do the following:
```js
// Send a HEAD request to determine if the file exists, IMO it's enough to check for the JS
fetch('webmks/wmks.min.js', {
  method: 'HEAD'
}).then(() => new Promise((resolve, reject) => { // Create and return with a promise
  // Create a <link> tag that loads the CSS file and append it to the page
  let link = document.createElement('link');
  link.rel = 'stylesheet';
  link.type = 'text/css';
  link.href = '/webmks/css/wmks-all.css';
  document.head.appendChild(link);

  // Create a <script> tag that loads the JS file
  let script = document.createElement('script');
  script.src = '/webmks/wmks.min.js';

  script.onload = () => resolve(); // Set an onload function that resolves the promise

  document.head.appendChild(script); // Append the <script> tag to the page
})).then(() => {
  // The code is loaded, we can do stuff with WebMKS
}).catch((error) => {
  // some error handling
});

```

himdel:
I suppose so..., which implementation of `fetch`?

ahrechushkin:
Is there any new about this feature?

skateman:
@AllenBW when will you have time to take a look? Do you have any questions about how and where to start?

AllenBW:
Heyoh pto caused the delayed response, if time is available later this week I can begin to give this some 👀 in earnest! 🤞

ahrechushkin:
up

skateman:
@AllenBW ping?

AllenBW:
Yah, haven't had the time. 😔

ahrechushkin:
up

AllenBW:
no update on my end

skateman:
Yes update on my end, final PR coming soon :smiling_imp:

PR #1553, opened 2019-07-01
Update the VM remote console button to the new unified HTML5 console

skateman:
In order to support [webmks](https://github.com/ManageIQ/manageiq-ui-service/issues/1520) here, there were some [changes](https://github.com/ManageIQ/manageiq/pull/18927) in the backend for the new HTML5 console meta type. This PR adjusts the codebase to these changes. 

Depends on: https://github.com/ManageIQ/manageiq/pull/18927 and https://github.com/ManageIQ/manageiq-api/pull/621

https://bugzilla.redhat.com/show_bug.cgi?id=1532720

PR #1554, opened 2019-07-02
 Add support for WebMKS remote consoles via a custom template

skateman:
This is the very last piece of the puzzle for having WebMKS support in the SUI...

@miq-bot add_reviewer @himdel 
@miq-bot add_label enhancement, hammer/no

Resolves https://github.com/ManageIQ/manageiq-ui-service/issues/1520
Depends on https://github.com/ManageIQ/manageiq-ui-service/pull/1553 https://github.com/ManageIQ/manageiq-api/pull/621 https://github.com/ManageIQ/manageiq/pull/18927
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1532720

### DIFF
--- a/client/app/core/product-features.constants.json
+++ b/client/app/core/product-features.constants.json
@@ -30,7 +30,7 @@
         "ADD": "sui_vm_snapshot_create",
         "DELETE": "sui_vm_snapshot_delete"
       },
-      "CONSOLE": "sui_vm_console",
+      "HTML5_CONSOLE": "sui_vm_html5_console",
       "WEB_CONSOLE": "sui_vm_web_console",
       "TAGS": "sui_vm_tags",
       "RETIRE": "sui_vm_retire",
--- a/client/app/services/consoles.service.js
+++ b/client/app/services/consoles.service.js
@@ -63,6 +63,9 @@ export function ConsolesFactory ($window, CollectionsApi, $timeout, $location, E
       case 'spice':
         openSpice(results)
         break
+      case 'webmks':
+        openWebmks(results)
+        break
       case 'vnc':
         openVnc(results)
         break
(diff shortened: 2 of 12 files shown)


skateman:
@himdel @AllenBW somehow the remote console file isn't accessible through `/ui/service` in development. Neither in this PR and in the master. However, I assume this thingie worked before I toughed anything and it will work in production as it should. I manually dropped the `/ui/service` prefix from the URL after opening the console popup and it worked.

himdel:
> the remote console file isn't accessible through /ui/service in development

Yeah, the current consoles have the same problem. There is no directory to serve those files from, before you build it. And webpack-dev-server only supports proxying, not custom data routes.

Testing should still be possible by running `yarn run build`, running ops ui, and going to http://localhost:3000/ui/service

himdel:
@skateman you want to add `/* eslint comma-dangle: 0 */` on the first line of the file.

AllenBW:
> @skateman you want to add `/* eslint comma-dangle: 0 */` on the first line of the file.

Should we just turn it off in the eslint config? we already turn this off in 4 other places 🤔

himdel:
Oh, @AllenBW I was under the impression that standardjs is involved and disabling globally would be non-trivial. If that's not so, yeah, by all means let's switch that off :)

AllenBW:
Ah that's right, we could rework linting to extend standard, have all the benefits of with the ability to edit easily 🤔  https://github.com/standard/eslint-config-standard

skateman:
This is ready for review, the test and the noVNC dependency is still relevant, the old consoles should work as before.

himdel:
LGTM, I've seen this work :)
(Waiting for core)

PR #1562, opened 2019-07-24
[IVANCHUK] Update gettext catalog for ivanchuk

mzazrivec:


himdel:
What's the reason behind introducing `es` in addition to the `es-ES` we already have?

Should `es` be *replacing* `es-ES` instead of having 2 versions?

(I don't think it makes sense to have a "spanish (generic)" and "spanish (spain)" as 2 variants, maybe one of them should be `es-419` for latin america?)

mzazrivec:
The `es` locale, that seems like it's been added, is a merge of the `es` translation we have in `hammer` branch. It just seems like a new locale, because we did not have it in `master` branch.

The `es` translation has been contributed by an external translating company, `es-ES` is a community contribution. And I guess that we can consider `es` a generic spanish translation, while `es-ES` contains Spain-specific translation.

himdel:
OK, as long as both are a valid up to date translations.. :) Thanks :)

PR #1565, opened 2019-07-25
Explicitly add the HTML for WebMKS into the webpack production config

skateman:
To be honest, I have no idea how to test this, but @himdel says this is *The Right Way&trade;* to fix this. The file was present in development, but not in production. Even though we're including the development file in production :rofl: 

@miq-bot add_label bug, ivanchuk/yes, hammer/no, changelog/no
@miq-bot assign @himdel 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1732730

### DIFF
--- a/config/webpack.prod.js
+++ b/config/webpack.prod.js
@@ -25,6 +25,14 @@ config.plugins.push(
     chunks: ['app']
   }),
 
+  // Build the HTML required for the WebMKS console
+  new HtmlWebpackPlugin({
+    base: '/',
+    filename: 'console/webmks.html',
+    template: '../client/console/common.ejs',
+    chunks: ['console_webmks']
+  }),
+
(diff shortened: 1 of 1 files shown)


himdel:
Closing, not the issue.
The file is present on the appliance.

skateman:
Okay, but what's the issue then? :laughing:

himdel:
The file is physically present on the appliance, so this has to be in the apache config.

himdel:
=> In `/etc/httpd/conf.d/manageiq-redirects-ui`:

```
RewriteRule ^/ui/service(?!/(assets|images|img|styles|js|fonts|vendor|gettext)) /ui/service/index.html [L]
```

Which means anything not in those directories (such as console/) will get redirected to index.html, which then decides to redirect to 404.

PR #1566, opened 2019-07-25
Service detail - don't show custom button for lists

himdel:
When adding a custom button (in ops ui, Automate > customization > buttons, under Services in the tree),
you can select a "Display for" value: single, list, both.

Single means the button should only be displayed in service detail screens,
list means the button should only be displayed in service list screens,
both means both :).

Service UI ignores the distinction until now, displaying all the buttons in service detail.

This partly merges the almost identical (except for html and css) components custom-button and custom-button-menu,
and adds an optional display-for attribute, listing the values of display_for to match against.

Thus, `<custom-button display-for="['both', 'single']"` will only show buttons for individual services,
while `<custom-button display-for="['both', 'list']"` will only show buttons for service lists.

And using it to only show the both/single buttons for service detail screens.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1650066


himdel:
Before:
before

After:
after

client/app/services/custom-button/custom-button.component.js
```diff
@@ -1,24 +1,51 @@
...
-    vmId: '<'
+    vmId: '<',
+    displayFor: '<?',
   },
   controller: CustomButtonController,
   controllerAs: 'vm',
-  templateUrl
 }
 
+export const CustomButtonComponent = Object.assign({}, base, {
```
himdel: This should really be just `= {...base, templateUrl: templateInline}` but alas the current build system does not support `...` yet.

himdel:
ugh, the tests are failing on missing Array#includes, adding array-includes shim

PR #1573, opened 2019-08-20
Node 12 compatibility

himdel:
node-sass < 4.12 fails with node 12, updating :).

(This is the SUI equivalent of https://github.com/ManageIQ/manageiq-ui-classic/pull/6008, to fix https://github.com/ManageIQ/manageiq-ui-classic/issues/5589)


himdel:
cc @martinpovolny

PR #1579, opened 2019-08-21
Make strings in notification drawer translatable

mzazrivec:
Changes in this PR:
* make notification sub-heading translatable
* make notification event types translatable
* fix `sprintf()` and `gettext` invocation
* add translatable `no-notifications-text`

https://bugzilla.redhat.com/show_bug.cgi?id=1655464
https://bugzilla.redhat.com/show_bug.cgi?id=1655536


himdel:
Before:

before

After:

after

PR #886, opened 2017-08-25
PT#149703867 standardjs/eslint swap

AllenBW:
https://www.pivotaltracker.com/story/show/149703867

What just happened? https://standardjs.com/rules.html

## A ss to demonstrate error reporting is as normal, browser and server console still show offending code
screen shot 2017-08-25 at 11 44 07 am

## A ss showing we still build and run correctly (notice that url), ALSO super important 😆 
screen shot 2017-08-25 at 12 06 51 pm

PR #1591, opened 2019-09-10
Remove standard js, use eslint

himdel:
This undoes parts of https://github.com/ManageIQ/manageiq-ui-service/pull/886, removing the standard js linter and replacing it with eslint, same as ui-classic.

The new eslint config is:

* globals and ignore from standard js config
* eslint defaults,
* eslint-config-angular,
* disable a bunch of angular rules which were failing - all of these are essentially "use angular version of whatever you're using" and make little sense in the context of a react future
* disable `no-unused-vars` - we should have this enabled and set to ignore underscore-prefixed names, but currently there are 71 places with this issue

That brings us to a minimal default config which currently passes validation.
Enabling all the same rules ui-classic has would mean mass-converting everything again, little point to that IMO.

But we can now add/remove rules on a case by case basis :).
(And trailing commas or semicolons are OK by default now.)

---

The first commit also drops the remains of typescript support, as the typescript version is ancient and we have no .ts or .tsx files.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+/reports/
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,53 @@
+module.exports = {
+  env: {
+    amd: true,
+    browser: true,
+    es6: true,
+    jasmine: true,
+    jquery: true,
+    node: true,
+    protractor: true,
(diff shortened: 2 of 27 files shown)


himdel:
Cc @mzazrivec

PR #914, opened 2017-09-06
Created functionality to automate Zanata upload

chalettu:
@miq-bot add_label housekeeping
@miq-bot add_label enhancement
@miq-bot add_label fine/no
@miq-bot add_label internationalization

PR #1593, opened 2019-09-17
Fix extraction of "Order # 608000000000012 and all of its service requests will be permanently removed!"

himdel:
When a string starts with leading space, the extractor trims the space, meaning the text will never get translated.

Making sure the whole phrase ends up in the catalog :).

Cc @mzazrivec 
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1752469

### DIFF
--- a/client/app/orders/process-order-modal/process-order-modal.html
+++ b/client/app/orders/process-order-modal/process-order-modal.html
@@ -6,7 +6,7 @@
 </div>
 <div class="modal-body">
   <i class="pf pficon-warning-triangle-o"></i>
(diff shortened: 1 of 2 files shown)


himdel:
`ag '" [^"]*"\s*\|\s*translate'` finds this, and nothing else. So, fingers crossed, hoping this is the only problematic case.

himdel:
Fails on:

```
$ node language/validate.js
Errors exist in language file
[ '{{ vm.order.name }}' ]
```

comes from https://github.com/ManageIQ/manageiq-ui-service/pull/914.

Removing, leaving only code to validate that HTML is not present in the strings.
Strings with angular templates are completely valid.
(Second commit)

PR #1595, opened 2019-10-10
DialogUser - use DialogData.outputConversion before submit and refresh

himdel:
Depends on https://github.com/ManageIQ/ui-components/pull/422

Right now, when submitting a service dialog,
we simply take the dialogData object, convert to JSON and send to the API.

To allow for output conversions, adding a DialogData.outputConversion method,
which should return the dialogData in a suitable format.

(Right now, all this really means is that it converts Dates to strings before JSON stringification.)

And the same needs to happen in refreshField for dynamic fields.

https://bugzilla.redhat.com/show_bug.cgi?id=1744413

@miq-bot add_label bug, ivanchuk/yes

PR #1606, opened 2019-11-04
[IVANCHUK] DialogUser - use DialogData.outputConversion before submit and refresh

himdel:
Ivanchuk version of https://github.com/ManageIQ/manageiq-ui-service/pull/1595
(`yarn.lock` conflicts, and `client/app.js` was missing the `array-includes` shim (added in master by https://github.com/ManageIQ/manageiq-ui-service/pull/1566))

---

~~Depends on ManageIQ/ui-components#422~~

Right now, when submitting a service dialog,
we simply take the dialogData object, convert to JSON and send to the API.

To allow for output conversions, adding a DialogData.outputConversion method,
which should return the dialogData in a suitable format.

(Right now, all this really means is that it converts Dates to strings before JSON stringification.)

And the same needs to happen in refreshField for dynamic fields.

https://bugzilla.redhat.com/show_bug.cgi?id=1744413

@miq-bot add_label bug


himdel:
Tests should go green once https://github.com/ManageIQ/manageiq-ui-service/pull/1607 is merged in ivanchuk.

himdel:
Rebased on top of #1607 :)

PR #1576, opened 2019-08-20
React support via react2angular

himdel:
This updates babel enough to support using react components in SUI.

(Also moving all the non-runtime dependencies to devDependencies for consistency.)

Docs: https://github.com/coatue-oss/react2angular

But the short form is, if you have a react component with PropTypes set,

```
   angularModule.component('fooBar', react2angular(FooBar))
```

makes it possible to use `<foo-bar num="5" str="'abc'" expr="vm.services.count"></foo-bar>` from angular components.

Cc @martinpovolny , @mzazrivec

PR #1609, opened 2019-11-14
[HAMMER] DialogUser - use DialogData.outputConversion before submit and refresh

himdel:
Hammer version of https://github.com/ManageIQ/manageiq-ui-service/pull/1595
With most of https://github.com/ManageIQ/manageiq-ui-service/pull/1607 (https://github.com/ManageIQ/manageiq-ui-service/pull/1591)
And array-includes shim (originally added in https://github.com/ManageIQ/manageiq-ui-service/pull/1566)
And package.json cleanup originally added in https://github.com/ManageIQ/manageiq-ui-service/pull/1576

---

~~Depends on ManageIQ/ui-components#422 (via https://github.com/ManageIQ/manageiq-ui-service/pull/1608, merged)~~

Right now, when submitting a service dialog,
we simply take the dialogData object, convert to JSON and send to the API.

To allow for output conversions, adding a DialogData.outputConversion method,
which should return the dialogData in a suitable format.

(Right now, all this really means is that it converts Dates to strings before JSON stringification.)

And the same needs to happen in refreshField for dynamic fields.

https://bugzilla.redhat.com/show_bug.cgi?id=1768457

@miq-bot add_label bug


simaishi:
@himdel 5.10.z build is now failing: `/bin/sh: webpack: command not found`

I see `webpack` moved from `dependencies` to `devDependencies` in this PR. The hammer/5.10.z build runs with `--production`, so I guess webpack is no longer being fetched.

So the question is... should webpack stay in `dependencies` as well (which is how ivanchuk branch is), or dp we need to switch to running without `--production`?

himdel:
Aaah, sorry about this, you should switch to running without `--production`.

(The difference is that https://github.com/ManageIQ/manageiq-appliance-build/pull/342 removed `--production` from ivanchuk+ but never made it to hammer)

simaishi:
@himdel thanks, I'll backport the appliance-build PR.

PR #1617, opened 2020-01-22
AuthenticationApi - remove Promise wrapping

himdel:
by wrapping $http promises in Promise, we're no longer running the code within the angular digest cycle,
so any login error will fail to display a flash message until the user causes a digest cycle by clicking the submit button again, or changing the inputs.

Undoing that wrap :).

Caused by https://github.com/ManageIQ/manageiq-ui-service/pull/1142
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1793532

### DIFF
--- a/client/app/core/authentication-api.factory.js
+++ b/client/app/core/authentication-api.factory.js
@@ -3,37 +3,35 @@ const TextEncoderLite = require('text-encoder-lite').TextEncoderLite
 
 // utf8-capable window.btoa
 function base64encode (str, encoding = 'utf-8') {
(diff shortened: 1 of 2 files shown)


himdel:
Testing:

go to login screen,
use invalid credentials,
fail to log in

before: no flash message until you click or change something
after: error flashes right away (+- server roundtrip)

himdel:
(updated to prefer `return $q.reject(error)` over `throw error`, consistency with https://github.com/ManageIQ/manageiq-ui-service/pull/1618)

PR #1618, opened 2020-01-22
Unify promise handling, fix non-angular promises ignoring digest cycle

himdel:
Follow-up to https://github.com/ManageIQ/manageiq-ui-service/pull/1617

`Promise` is a browser-implemented promise implementation, with no angular linkages,
`$q` is an angular Promise implementation, which guarantees the code to run within angular's digest cycle.

So, any code along the lines of `return new Promise((resolve, reject) => { ..... resolve(something) ... reject(something) }` does not really work with angular, unless the body never touches any angular models and unless we always wrap the return in a `$q.when`.

The angular equivalent would be `return $q((resolve, reject) => {...}`, but most of the time, we don't need another promise layer because we're already dealing with promises returned by other code. (Any promise returned by `$http` (and thus our `CollectionsApi`) is already a `$q` promise.)

Also, we had a lot of constructs along the lines of..

```
return new Promise((resolve, reject) => {
  otherPromise.then((data) => resolve(data);
})
```

which is equivalent to `return otherPromise`, except that the first throws away any errors or exceptions. Which is seldom really what we want.

=> So, removing all but one `new Promise` (the remaining one being webmks consoles, which run outside angular).

---

Two more catches:

* order matters, `vanillaPromise.then(() => qPromise)` will still return a vanilla promise, while `qPromise.then(() => vanillaPromise)` returns a `$q` promise
* `$q` promises never change state outside angular's digest scope, which means they don't really work great with tests

That means that using `Promise.resolve` and `Promise.reject` is usually fine, except when really creating a new promise, outside any promise context.

But it also means using `$q.reject` and `$q.resolve` can be problematic for tests, which don't normally trigger the digest cycle.

But, not wrapping allows us to choose which kind of promises to use in specs by mocking the original source of the promise (mocking http.get to return Promise.reject means no digest is needed, as long as there's no wrapping in other promise kinds anywhere on the way).

Therefore settling on `Promise.resolve` and `Promise.reject` for returning resolves/rejects wherever possible.


himdel:
@miq-bot remove_label wip

Issue #1527, opened 2019-02-28
Direct link to service redirects to login page even if user authorized


bekasovoleg:
Description of problem:
Direct link to service redirects to login page even if user is authorized

How reproducible:

Steps to Reproduce:
1. Login to UI Service
2. Open new tab and go to ui/serivce/services/:id. We get login page instead of specified service page.

Actual results:
We get login page instead of specified service page.

Expected results:
Page with specified service (ui/serivce/services/:id).

Additional info:

bekasovoleg:
@chalettu looks like problem in code below
https://github.com/ManageIQ/manageiq-ui-service/blob/master/client/app/core/authorization.config.js#L66
`console.log(RBAC.all())` returns empty object

bekasovoleg:
@AllenBW, take a look pls

himdel:
Definitely a bug, if you can open a PR, I'll merge :)

himdel:
When logged in as admin with valid credentials...

going to / .. should load the dashboard (at the same url), but goes to /login instead
going to /services/80 .. should load the url, redirects to /login

(opening the url in the same tab seems to work, just not different tabs, probably caused by the use of sessionStorage as opposed to localStorage)

Looking into it.

himdel:
Originally implemented in https://github.com/ManageIQ/manageiq/pull/5904 (although back then, sessionStorage was shared between tabs)

bekasovoleg:
@himdel good job, thanks a lot!

PR #1623, opened 2020-01-29
Session - use localStorage instead of sessionStorage

himdel:
sessionStorage is not shared between tabs (unless one gets opened from the other),
and disappears when closing the tab.

Thus, opening a link in a new tab just doesn't work and redirects to `/login`.
Changing to use localStorage instead - the same change was needed in ui-classic: https://github.com/ManageIQ/manageiq-ui-classic/pull/3065

The one place that's still using sessionStorage after this is the ApplianceInfo service, for caching. That looks more like ephemeral cache which doesn't have to be shared.

Fixes https://github.com/ManageIQ/manageiq-ui-service/issues/1527

### DIFF
--- a/client/app/core/authorization.config.js
+++ b/client/app/core/authorization.config.js
@@ -44,13 +44,15 @@ export function authConfig ($httpProvider) {
 }
 
 /** @ngInject */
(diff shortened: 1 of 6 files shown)


himdel:
@miq-bot remove_label wip

It works, opening /services/80 in a new tab no longer redirects to /login.
Opening / goes to Dashboard.

PR #1624, opened 2020-01-29
login - add spinner to Log In button

### LINKED ISSUE #1620: No visual indication after clicking Login
On login screen,
after filling out username and clicking login,
nothing seems to happen until the server responds.

Which is a problem beacuse the server doesn't always respond that fast, leading the user to believe nothing happened.

We should add a spinner, same as ui-classic.

himdel:
Adds a spinner to the Log In button...

before (and when idle):
login

after, after pressing the button:
20200130183354 mp4

(minus the gif artifacts)

Fixes #1620

~~@epwinchell Does this make sense? I'm.. not so sure about the CSS I'm adding here.., it works, but that's the best thing I can say about it :)~~
(Without *some* css fix, the spinner just blows the button up to a huge size because `.spinner-sm` 18px x 18px becomes ~25px when rotating.)
Thanks @epwinchell for much better CSS :).

### DIFF
--- a/client/app/states/login/_login.sass
+++ b/client/app/states/login/_login.sass
@@ -32,13 +32,7 @@
     max-width: 200px
 
   .spinner-wrap
-    bottom: -5px
-    display: inline-block
-    height: 14px
-    position: relative
-    width: 21px
-
-    &>.spinner
-      left: 3px
-      position: absolute
-      top: -5px
+    height: 16px
+    position: absolute
+    right: 40px
+    z-index: 1000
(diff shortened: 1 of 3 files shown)


epwinchell:
@miq-bot add_reviewer @epwinchell

epwinchell:
@himdel That's a unique solution... :) How about hooking it to the container div rather than the button?

himdel:
We don't have a full screen spinner in SUI, so I was hoping to use the patternfly spinner instead of copying the ui-classic solution.

But if you prefer that, I'll just add `spin.js` and use it as in ui-classic :)

(But IMO the solution is not *so* unique, putting the spinner in a button we're waiting for is defintely something I've seen before :).)

himdel:
@miq-bot remove_label wip

PR #1631, opened 2020-03-17
[JANSA] Update source catalog for jansa

mzazrivec:
This includes current (alpha) translations for Jansa branch.


himdel:
```
Language #setLocale should allow a locale to be set FAILED
	AssertionError: expected 'es' to equal 'es-ES'
	    at Context.<anonymous> (client/app/core/language.service.spec.js:96:25)
```

is that intentional?

(The directory is still es-ES too)

mzazrivec:
@himdel fixed

PR #1636, opened 2020-04-06
build:git-hash: prefer reading from BUILD if available

### LINKED ISSUE #1634: About screen "SUI Version" is empty in container build
In container image build, source is downloaded as a tarball ahead of time and there is no .git directory. This causes "SUI Version" to be empty in About screen.

Is it possible to look for a file existence and use value from there if exists (which is how OpsUI works), rather than always calling `git rev-parse`?

himdel:
`yarn run build` runs `build:git-hash` which runs `config/githash.js`, which creates `client/version/version.json` with the current git sha, by running git rev-parse.

For container builds, we don't have git, so the commit is empty.

This updates the task to try reading a `BUILD` file in SUI root, and using its contents as the sha.
(If a `BUILD` file is not present, it falls back to `git rev-parse`, and empty commit if that fails.)

Fixes #1634, cc @simaishi :)

### DIFF
--- a/config/githash.js
+++ b/config/githash.js
@@ -1,11 +1,15 @@
-const shell = require('shelljs')
-const fs = require('fs')
-const path = require('path')
-const filePath = path.join(__dirname, '../client/version/version.json')
-const gitCommit = shell.exec('git rev-parse HEAD', {silent: true}).stdout
+const { exec } = require('shelljs');
+const { existsSync, readFileSync, writeFileSync } = require('fs');
+const { join } = require('path');
 
-const gitFileContents = `{"gitCommit": "${gitCommit.replace("\n", '')}"}` // eslint-disable-line 
-fs.writeFile(filePath, gitFileContents, (err) => {
-  if (err) throw err
-  console.log(`Successfully wrote Git hash - ${gitCommit}`)
-})
+const input = join(__dirname, '../BUILD');
+const output = join(__dirname, '../client/version/version.json');
+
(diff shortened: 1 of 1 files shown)


config/githash.js
```diff
@@ -1,11 +1,15 @@
...
+const { exec } = require('shelljs');
+const { existsSync, readFileSync, writeFileSync } = require('fs');
+const { join } = require('path');
+
+const input = join(__dirname, '../BUILD');
+const output = join(__dirname, '../client/version/version.json');
+
+let gitCommit = existsSync(input)
+  ? readFileSync(input, { encoding: 'utf-8' })
+  : exec('git rev-parse HEAD', { silent: true }).stdout;
```
h-kataria, line 10: should this be `shell.exec`

himdel: Just exec, it's imported as `const { exec } = require('shelljs');`

(eslint was complaining that `const exec = require('shelljs').exec` should prefer destructuring)

simaishi:
Jansa backport details:

```
$ git log -1
commit 36b8b0338fd5e8845ad76f311410b5f50ddf7f44
Author: Harpreet Kataria <hkataria@redhat.com>
Date:   Sun Apr 19 11:01:37 2020 -0400

    Merge pull request #1636 from himdel/BUILDsha

    build:git-hash: prefer reading from BUILD if available
    (cherry picked from commit 0cee78c9886a8b6a56dd09177a21dc771623ad1b)
```

PR #1637, opened 2020-04-15
_login: change #brand img height from default to what ui-classic uses

himdel:
Change the login size brand image from height 18 to 38.

ui-classic and ui-service are using the same brand.svg image,
but ui-service defaults to the patternfly 18px height,
while ui-classic overrides the height to 38px.

Updating to do the same.

Before:

before

After:

after

ui-classic:

miq

Cc @epwinchell @john-dupuy

### DIFF
--- a/client/app/states/login/_login.sass
+++ b/client/app/states/login/_login.sass
@@ -7,8 +7,10 @@
   margin-bottom: -38px
   margin-top: -60px
 
-  #brand img
-    height: 38px !important
+  // override .login-pf #brand img to match ui-classic brand.scss
+  #brand // sass-lint:disable-line no-ids
+    img
+      height: 38px !important // sass-lint:disable-line no-important
 
   .container
     padding-top: 0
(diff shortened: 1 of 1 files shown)


himdel:
Without this...

`manageiq-ui-classic/vendor/assets/stylesheets/brand.scss`

```
.login-pf #brand img {
  height: 38px !important;
}
```

`manageiq-ui-service/node_modules/patternfly/dist/sass/patternfly/_login.scss`

```
.login-pf #brand img {
   height: 18px
}
```

john-dupuy:
@himdel can we also override the branding above the left nav for the SSUI vs UI in this PR?

himdel:
@john-dupuy nothing to fix there...

It's the same image,
again, ui-service is using the patternfly default (20px this time) (`.navbar-pf-vertical .navbar-brand .navbar-brand-icon` in `_navbar_vertical.scss`)
while ui-classic is setting this in `brand.scss` - but to 20px as well.

So, looks like both have the same size there.
(If not, changing the size there may or may not involve having to rewrite the angular-patterfly navbar.)

PR #1647, opened 2020-05-26
Update ui-components to 1.4.1

mzazrivec:
ManageIQ/ui-components#439
ManageIQ/ui-components#440


himdel:
`yarn add yarn add`? :)

We probably shouldn't depend on `yarn` or `add` in `package.json` :)

mzazrivec:
@himdel Fixed.

PR #1693, opened 2021-01-06
My Orders list: fix TypeError

himdel:
Pagination check all support from https://github.com/ManageIQ/manageiq-ui-service/pull/1533 fails when not passing a `checkAll` object to pagination from the start.

    angular.js:15635 TypeError: [ManageIQ] Cannot read property 'checked' of undefined
        at initCheckAll (app-6c33a85c672b91a2d834.js:327476)
        at ComponentController.vm.$onChanges (app-6c33a85c672b91a2d834.js:327459)

Make sure we default to false in such case.

### DIFF
--- a/client/app/shared/pagination/pagination.component.js
+++ b/client/app/shared/pagination/pagination.component.js
@@ -60,7 +60,7 @@ function ComponentController () {
   }
 
   function initCheckAll () {
-    vm.checkAllState = vm.checkAll.checked || false
+    vm.checkAllState = vm.checkAll && vm.checkAll.checked || false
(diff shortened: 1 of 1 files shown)


mzazrivec (approved):

mikebutak:
hello @himdel.  thanks for working on this!  I changed that line of code at `/opt/manageiq/manageiq-ui-service/client/app/shared/pagination/`, restarted the evmserverd, and yet the error still occurs.  Do you have any suggestions?  

Also, I wonder if anyone uses the self-service UI since it has this bug that makes it unusable?

PR #1702, opened 2021-03-18
[LASKER] Translations provided on 2021-03-12

chessbyte:


chessbyte:
Closing since the net result of these translations is not a single file was updated.

PR #1711, opened 2022-01-07
update node and other versions

kavyanekkalapu:
For https://github.com/ManageIQ/manageiq-ui-classic/issues/7930

Issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/7930

PR #1773, opened 2022-02-22
Downgrade node version to 14

kavyanekkalapu:
Follow up of https://github.com/ManageIQ/manageiq-ui-service/pull/1711
Issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/7930
Downgrading node version to 14.

in 16, Node-sass 4.14 ( which is coming from angular patternfly) is giving an error.

Node-sass 4.X supports only 14. 16 needs node-sass 6.X. 

@miq-bot assign @Fryguy 
@miq-bot add_reviewer @Fryguy 
@miq-bot add-label technical_debt


Fryguy:
Ugh this is going to be a problem for build then.  I'm not sure we can do node 16 at all across the board if this is the case.  cc @bdunne @chessbyte

Fryguy:
I'm surprised that if angular pattern ply is a problem then why didn't we hit this with UI classic?

kavyanekkalapu:
> I'm surprised that if angular pattern ply is a problem then why didn't we hit this with UI classic?

Yes, i will further investigate once this pr is merged as we need this for release.

Fryguy:
If we have to keep 2 different node versions, we may have to do something at build time where we install both node 16 and node 14 and use nvm, then switch before we go to compile each one.  I wonder if we can read the node version from the package.json, and do it somewhat dynamically.

kavyanekkalapu:
> If we have to keep 2 different node versions, we may have to do something at build time where we install both node 16 and node 14 and use nvm, then switch before we go to compile each one. I wonder if we can read the node version from the package.json, and do it somewhat dynamically.

@Fryguy Do we have to do this in this repo? Can you share more details like where to add code?

Fryguy:
https://github.com/ManageIQ/manageiq-rpm_build/blob/master/lib/manageiq/rpm_build/generate_core.rb#L53 is where we build the service UI , but node is installed in the build container itself here https://github.com/ManageIQ/manageiq-rpm_build/blob/master/Dockerfile#L18.  The classic UI is compiled here https://github.com/ManageIQ/manageiq-rpm_build/blob/master/lib/manageiq/rpm_build/generate_core.rb#L41.

Fryguy:
@kavyanekkalapu I'm starting to lean towards maybe downgrading temporarily to node 14 across the board, then do 16 as a follow up.  At least then we get off of 12, which is really why we are upgrading.  That may also be a better candidate for backport should we choose to do that.  WDYT? cc @chessbyte .

chessbyte:
>  I'm starting to lean towards maybe downgrading temporarily to node 14 across the board, then do 16 as a follow up.

Smaller steps lead to continuous improvement!

kavyanekkalapu:
@chessbyte Ok i will try to raise prs for node 14. i will build those on currently merged node 16 ones.

Fryguy:
Merging on red...There will be a separate PR for the existing issue which has to do with the novnc version.

PR #1774, opened 2022-02-25
Cron for GitHub Actions

Fryguy:
Cron for GitHub Actions


chessbyte:
@Fryguy this one is failing, but seems unrelated

PR #1787, opened 2022-04-20
Remove services header svg

kavyanekkalapu:
**Bug:**If we don't have svg, we are seeing broken svg.
**After** If we don't have svg, it will display alt text as span.

@miq-bot assign @Fryguy 
@miq-bot add-label bug
@miq-bot add_reviewer @Fryguy

### DIFF
--- a/client/app/layouts/application.html
+++ b/client/app/layouts/application.html
@@ -1,5 +1,5 @@
 <div class="layout-application" ng-controller="NavigationController as vm">
-  <pf-vertical-navigation items="vm.items" brand-src="{{vm.applianceInfo.brand}}" brand-alt="{{ vm.applianceInfo.nameFull }}"
+  <pf-vertical-navigation items="vm.items" brand-alt="{{ vm.applianceInfo.nameFull }}"
        update-active-items-on-click="true"
        pinnable-menus="true" item-click-callback="vm.handleItemClick" show-badges="true">
     <pf-toast-notification-list notifications="vm.toastNotifications" show-close="true"


Fryguy:
If we do have the svg, though, will this no longer show it?

kbrock:
so I just heard:

change the code to include the value `if brand.presence?`

kavyanekkalapu:
@Fryguy @kbrock Yes we no longer show it, right now it is displaying alt text as header. Do you want me to change it like svg present?

kbrock:
@Fryguy it seems like @kavyanekkalapu is never seeing a brand image so she dropped it.

Can customers still upload a brand image?
I thought this feature was dropped quite a while ago.
We used to have database syncing or something for this a while back. But it does not seem pod friendly and I thought it was dropped

kbrock (approved):
I am pretty sure we do not support customer specific images any more.
And that we do not sync brand images across images via postgres.

But I am not the one who removed either of these features, so I defer.

Fryguy:
Technically I think we can still upload customer logos, though admittedly they probably don't work in podified.  However, this is not a customer logo, this is the ManageIQ logo (or the replacement during productization).  Correct me if I'm wrong though.

kavyanekkalapu:
@Fryguy @kbrock Opensource version doesn't have header at all.

Fryguy:
@kavyanekkalapu Are you still looking into this one?

kavyanekkalapu:
@Fryguy if we have to have conditions, we have to have some backend code to check if svg is present..

jeffibm (approved):

GilbertCherrie:
@Fryguy I think we can close this issue now since it was fixed with the new svg

GilbertCherrie:
@Fryguy I think we can close this issue now since it was fixed with the new svg

PR #1834, opened 2023-03-02
Force upgrading node-sass to ^8.0.0

MelsHyrule:
There are many package dependencies that our current version of `node-sass` brings in that are security risk (ex tar-2.2.2, trim-newlines-1.0.0, scss-tokenizer-0.2.3). So rather than upgrade each of these this PR aims to fix these issues at the source. 

```
➜  manageiq-ui-service git:(master) ✗ yarn why node-sass
└─ angular-patternfly@npm:5.0.3
   └─ node-sass@npm:4.14.1 (via npm:^4.11.0)
```
Note: `node-sass` is a package brought in by `angular-patternfly`, but because we cannot upgrade `angular-patternfly` we're instead upgrading `node-sass`. 

For this PR to be merged we need to be on node v16 see [here](https://www.npmjs.com/package/node-sass?activeTab=readme). The node upgrade is being done [here](https://github.com/ManageIQ/manageiq-ui-service/pull/1833).

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,6 +140,7 @@
   "packageManager": "yarn@3.0.2",
   "resolutions": {
     "nth-check": "^2.1.1",
+    "node-sass": "^8.0.0",
     "moment": "~2.29.4",
     "moment-timezone": "^0.5.41"
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1481,6 +1481,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@tootallnate/once@npm:1":
+  version: 1.1.2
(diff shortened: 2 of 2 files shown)


Fryguy:
@MelsHyrule Does it make sense to pin to v6 for now? or is that still a problematic version?

MelsHyrule:
@Fryguy Pinning to v6 is possible, it will just not fix as many issues as it could have. Updated the PR now to match the node upgrade.

Fryguy:
Merging this on red as it contributes to fixing the Node 18 failures (see #1844)

PR #1910, opened 2024-06-26
Fix menu item order

GilbertCherrie:
Fixed the side nav menu item order.

Before:
Screenshot 2024-06-26 at 11 45 42 AM

After:
Screenshot 2024-06-26 at 12 35 49 PM


Fryguy:
I think My Services should come before My Orders.  In a real database, orders would be completed, and users wouldn't care about them, per se, whereas they do care about their running services.

GilbertCherrie:
> I think My Services should come before My Orders. In a real database, orders would be completed, and users wouldn't care about them, per se, whereas they do care about their running services.

I fixed the order and repushed.

PR #1925, opened 2024-09-16
Pin path-to-regexp

GilbertCherrie:
Update sinon to ~19.0.2 and pinned path-to-regexp to 0.1.10 to ensure we are using either the latest version of path-to-regexp or a version with patched vulnerabilities.

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "sass-loader": "^11.1.1",
     "save-remote-file-webpack-plugin": "^1.1.0",
     "shelljs": "~0.8.4",
-    "sinon": "~9.2.1",
+    "sinon": "~19.0.2",
     "sinon-chai": "~3.5.0",
     "style-loader": "~2.0.0",
     "url-loader": "~4.1.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -144,6 +144,8 @@
     "moment-timezone": "^0.5.41",
     "moment": "~2.29.4",
     "node-sass": "^8.0.0",
-    "nth-check": "^2.1.1"
+    "nth-check": "^2.1.1",
+    "express/path-to-regexp": "0.1.10",
+    "sinon": "~19.0.2"
```
Fryguy, line 149: Why do we need this resolution for sinon if we've updated the dependency?

package.json
```diff
@@ -144,6 +144,8 @@
     "moment-timezone": "^0.5.41",
     "moment": "~2.29.4",
     "node-sass": "^8.0.0",
-    "nth-check": "^2.1.1"
+    "nth-check": "^2.1.1",
+    "express/path-to-regexp": "0.1.10",
```
Fryguy, line 148: Can you add this in alphabetically?

GilbertCherrie: Should be fixed now

package.json
```diff
@@ -141,9 +141,11 @@
   "packageManager": "yarn@4.3.1",
   "resolutions": {
     "d3-color": "~3.1.0",
+    "express/path-to-regexp": "0.1.10",
```
Fryguy, line 144: Can we add a `~` here?  I don't want to get stuck on this specific version if there are further fixes.

Fryguy:
I think you accidentaly removed the yarn.lock changes from this PR

GilbertCherrie:
> I think you accidentaly removed the yarn.lock changes from this PR

Yeah, it should be fixed now

Fryguy (approved):

PR #1939, opened 2024-11-21
Update jquery to 3.7.1

GilbertCherrie:
Update jquery to ~3.7.1 using resolutions. Need to use resolutions since both angular-patternfly and patternfly are already on the latest versions and no longer receiving updates. Both these packages are pulling in jquery 3.4.1 so need to use resolutions to update jquery to ~3.7.1.


Fryguy:
Looking at the yarn.lock, I don't think you need to pin the resolution at all...just rebuild the jquery results in the lockfile and you're good.

Fryguy:
Oh yeah `~` is only the patch version. Ok, then yeah, you can't do it other than resolutions

PR #1944, opened 2025-01-20
Update bootstrap

GilbertCherrie:
Update Bootstrap to 5. Also added popper.js since it is a dependency of bootstrap 5.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1328,6 +1328,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@popperjs/core@npm:~2.11.8":
+  version: 2.11.8
+  resolution: "@popperjs/core@npm:2.11.8"
+  checksum: 10/ddd16090cde777aaf102940f05d0274602079a95ad9805bd20bc55dcc7c3a2ba1b99dd5c73e5cc2753c3d31250ca52a67d58059459d7d27debb983a9f552936c
+  languageName: node
+  linkType: hard
+
 "@sindresorhus/is@npm:^0.7.0":
   version: 0.7.0
   resolution: "@sindresorhus/is@npm:0.7.0"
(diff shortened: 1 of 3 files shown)


GilbertCherrie:
@Fryguy I fixed the conflicts here if you can review again when you time

Fryguy:
Why is popper added directly. That is, won't it just come in naturally as a dependency?

GilbertCherrie:
> Why is popper added directly. That is, won't it just come in naturally as a dependency?

When I try to run npm start without the popper package in the package.json it throws this error:
Screenshot 2025-03-31 at 11 39 34 AM

Fryguy:
ok, so from what I understand - the better way to declare missing peer dependencies is to use [packageExtensions](https://yarnpkg.com/configuration/yarnrc#packageExtensions).  This allows you to essentially force bootstrap to pull in the dependency (and we don't have to declare it as "our" dependency.

So in the .yarnrc.yml you would add something like:

```yml
packageExtensions:
  "bootstrap@":
    dependencies:
      "popper.js": "~1.16.1"
  ```

**Comment by GilbertCherrie** [comment]
@Fryguy I fixed the dependency issue if you can review this again when you get a chance.

**Comment by GilbertCherrie** [comment]
@Fryguy I fixed the dependency issue if you can review this again when you get a chance.

**Review by elsamaryv — APPROVED** [review]

**Comment by Fryguy** [comment]
@GilbertCherrie This went into conflict

File: yarn.lock [hunk 1]
```diff
@@ -3260,19 +3258,12 @@ __metadata:
...
-  checksum: 10/c46285f9d46f6d7a716fa2d294a68b83d8da5f394b2df80e96e0015055e78899312186ac998474568949545e8edcea5c396d500fb9de71c2c81ad5bac1fb56b6
-  languageName: node
-  linkType: hard
-
-"bootstrap@npm:>=2.3.2, bootstrap@npm:>=3":
-  version: 5.3.6
-  resolution: "bootstrap@npm:5.3.6"
+"bootstrap@npm:~5.0.0":
+  version: 5.0.2
+  resolution: "bootstrap@npm:5.0.2"
```
Fryguy, line 3263: This ended up in a downgrade? Can we go to at least 5.3.6? (and by extension popperjs/core ^2.11.8, which also is showing a downgrade?)

elsamaryv (approved):

PR #2007, opened 2025-08-26
[Morphy] Update node-sass

GilbertCherrie:
Updated node-sass to ~5.0.0 to remove semver version ~5.3.0.


Fryguy:
I should have asked this before merging, but did you try building with the same node version for morphy? node-sass has a well-defined lookup table of version compatibility: https://www.npmjs.com/package/node-sass#node-version-support-policy

PR #2054, opened 2026-02-24
Update minimatch

GilbertCherrie:
Replaces #2053

Fixes: #2052

Update minimatch to latest version


Fryguy:
oh sweet they updated all the major versions...that's great

PR #2078, opened 2026-04-24
Ignore sass warnings from dependencies

Fryguy:
This drops all of the output from the webpack build, which is not actionable. Specifically, we can't upgrade patternfly, so this removes all of the warnings related to patternfly.

@asirvadAbrahamVarghese Please review. Similar to https://github.com/ManageIQ/manageiq-ui-classic/pull/10009.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "postcss-loader": "~4.0.4",
     "raw-loader": "~4.0.2",
     "resolve-url-loader": "~5.0.0",
-    "sass": "~1.34.1",
+    "sass": "~1.35.1",
     "sass-lint": "~1.13.1",
     "sass-loader": "^11.1.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -116,7 +116,7 @@
     "postcss-loader": "~4.0.4",
     "raw-loader": "~4.0.2",
     "resolve-url-loader": "~5.0.0",
-    "sass": "~1.34.1",
+    "sass": "~1.35.1",
```
Fryguy, line 119: 1.35.1 is the version that introduced the `quietDeps` option. Note that I intentionally was conservative in the upgrade here, but I think we should be able to get to sass `^1.99.0`, perhaps in a separate PR.

asirvadAbrahamVarghese (approved):
LGTM, anything else can be handled as a follow-up

PR #2080, opened 2026-04-24
Drop line numbers from locale template

Fryguy:
This avoids having to update the locale for little changes that don't actually change the locale strings. We have done this in other parts of the application as well. Note that I would prefer to drop the file names as well, but that's not an option in angular-gettext-tools.

Since the locale is updated on every `yarn run test`, this also avoids unnecessary changes to the locale file that create a dirty git repo on test.

@asirvadAbrahamVarghese Please review.
cc @jrafanie 

Similar to https://github.com/ManageIQ/ui-components/pull/584

### DIFF
--- a/client/gettext/po/manageiq-ui-service.pot
+++ b/client/gettext/po/manageiq-ui-service.pot
@@ -4,1653 +4,1592 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Project-Id-Version: \n"
 
-#: ./client/app/components/dashboard/dashboard.html:110
+#: ./client/app/components/dashboard/dashboard.html
 msgid "$0"
 msgstr ""
 
-#: ./client/app/services/resource-details/resource-details.component.js:211
+#: ./client/app/services/resource-details/resource-details.component.js
 msgid "%"
 msgstr ""
 
-#: ./client/app/components/notifications/notification-body.html:22
-#: ./client/app/components/notifications/notification-body.html:47
+#: ./client/app/components/notifications/notification-body.html
 msgid "% Complete"
 msgstr ""
 
-#: ./client/app/core/event-notifications.service.js:296
+#: ./client/app/core/event-notifications.service.js
 msgid "%d new notifications"
 msgstr ""
 
-#: ./client/app/core/navigation/navigation-controller.js:176
+#: ./client/app/core/navigation/navigation-controller.js
 msgid "%d unread notifications"
 msgstr ""
 
-#: ./client/app/shared/action-button-group/action-button-group.component.js:32
+#: ./client/app/shared/action-button-group/action-button-group.component.js
 msgid "%s"
 msgstr ""
 
-#: ./client/app/services/edit-service-modal/edit-service-modal.component.js:32
+#: ./client/app/services/edit-service-modal/edit-service-modal.component.js
 msgid "%s was edited."
 msgstr ""
 
-#: ./client/app/services/poweroperations.service.js:179
+#: ./client/app/services/poweroperations.service.js
 msgid "%s was retired."
 msgstr ""
 
-#: ./client/app/services/poweroperations.service.js:170
+#: ./client/app/services/poweroperations.service.js
 msgid "%s was started."
 msgstr ""
 
-#: ./client/app/services/poweroperations.service.js:173
+#: ./client/app/services/poweroperations.service.js
 msgid "%s was stopped."
 msgstr ""
 
-#: ./client/app/services/poweroperations.service.js:176
+#: ./client/app/services/poweroperations.service.js
 msgid "%s was suspended."
 msgstr ""
 
-#: ./client/app/layouts/application.html:87
+#: ./client/app/layouts/application.html
 msgid "(Current Group)"
 msgstr ""
 
-#: ./client/app/services/resource-details/resource-details.component.js:260
+#: ./client/app/services/resource-details/resource-details.component.js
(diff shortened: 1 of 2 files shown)


asirvadAbrahamVarghese (approved):
LGTM, @jrafanie do you want to take a look?

jrafanie:
@Fryguy @asirvadAbrahamVarghese this isn't in the list for the locale rake task and it's not a Vmdb::Plugin (I think), so I don't think it's being used... do we need to add it?  https://github.com/ManageIQ/manageiq/blob/master/lib/tasks/locale.rake#L192-L195

Fryguy:
I think we should include it yes

Fryguy:
Actually better question is how it's used. If we include it in the main one and it's not used here that's pointless. If it's separate then we might need a separate entry in the globalization tool

jrafanie:
> Actually better question is how it's used. If we include it in the main one and it's not used here that's pointless. If it's separate then we might need a separate entry in the globalization tool

I honestly have no idea if and how ui-components or react-ui-components pull in the translations... 

This was added in https://github.com/ManageIQ/manageiq/commit/9ab5a2ad5fed1abdb16cddf1fe32d318f2d77e8b, which I can't find the related PR.  I assume the ui-service or ui-classic pulls in the ui-components react-ui-components but I don't know for sure if translations for strings in ui-components/react-ui-components would show up in ui-service or ui-classic.

@asirvadAbrahamVarghese are you able to figure out how it works?  I know [this is one of the main entrypoints](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/javascript/oldjs/i18n.js) where ui-classic loads the UI json files, which would be based on the pot files we get translated

jrafanie:
> @asirvadAbrahamVarghese are you able to figure out how it works? I know [this is one of the main entrypoints](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/javascript/oldjs/i18n.js) where ui-classic loads the UI json files, which would be based on the pot files we get translated

I did find some gettext setup in the ui service client but not sure if we correctly merge the message catalogs from ui-components, ui-service, etc. into the main one, that we'd also load merged one in ui-service. 

```
client/app.js:require('angular-gettext')
client/app/core/core.module.js:import { gettextInit } from './gettext.config.js'
client/app/core/core.module.js:    'gettext',
client/app/core/core.module.js:  .run(gettextInit)
client/app/core/gettext.config.js:import languageFile from '../../gettext/json/manageiq-ui-service.json'
client/app/core/gettext.config.js:export function gettextInit ($window, gettextCatalog, gettext) {
client/app/core/gettext.config.js:  gettextCatalog.debug = false
client/app/core/gettext.config.js:  gettextCatalog.loadAndSet = function (lang) {
client/app/core/gettext.config.js:      gettextCatalog.setCurrentLanguage(lang)
client/app/core/gettext.config.js:        gettextCatalog.setStrings(lang, languageFile[lang])
client/app/core/gettext.config.js:  $window.N_ = gettext
client/app/core/gettext.config.js:  $window.__ = gettextCatalog.getString.bind(gettextCatalog)
client/app/core/language.service.js:import languageFile from '../../gettext/json/available_languages.json'
client/app/core/language.service.js:export function LanguageFactory ($http, $q, $log, $localStorage, Session, $window, gettextCatalog, lodash) {
client/app/core/language.service.js:    gettextCatalog.loadAndSet(code)
```

Fryguy:
I believe `import languageFile from '../../gettext/json/manageiq-ui-service.json'` is the main i18n entry point for the service UI. If we're not updating that, then I'd consider it missed translations.

If we merge both catalogs together, then this could just be the same json file as in ui-classic, but that frankly feels inefficient and we probably should have 2 separate catalogs one for each UI.  Additionally, that would mean we'd also need ui-components in both catalogs (i.e. imported twice).

Fryguy:
Note this PR kind of has nothing to do with that, as I'm just trying to get rid of the line numbers to avoid unneeded churn. (I do understand that we need to solve that though)

jrafanie:
> Note this PR kind of has nothing to do with that, as I'm just trying to get rid of the line numbers to avoid unneeded churn. (I do understand that we need to solve that though)

yeah, I can merge but then we'll not respond to this PR and followup... 🤣 

> I believe `import languageFile from '../../gettext/json/manageiq-ui-service.json'` is the main i18n entry point for the service UI. If we're not updating that, then I'd consider it missed translations.
> 
> If we merge both catalogs together, then this could just be the same json file as in ui-classic, but that frankly feels inefficient and we probably should have 2 separate catalogs one for each UI. Additionally, that would mean we'd also need ui-components in both catalogs (i.e. imported twice).

I couldn't find discussion of the merging of catalogs but clearly that was the intent re: 

> This was added in [ManageIQ/manageiq@9ab5a2a](https://github.com/ManageIQ/manageiq/commit/9ab5a2ad5fed1abdb16cddf1fe32d318f2d77e8b),

I don't know if the frontend code, in service ui or other non-ui classic repos, is reading in the merged catalog.  It should either have it's own json/pot and be similar to 'bring your own translations' such as for power or be merged with core and read the merged pot/json.

Fryguy:
> yeah, I can merge but then we'll not respond to this PR and followup... 🤣

I'll open a separate issue.

jrafanie (approved):
Merging... there will be a followup issue to track how these translations are being used.

jrafanie:
> I'll open a separate issue.

ok, merged, link here when you create it. Thanks!

PR #2099, opened 2026-07-07
Update webpack packages

GilbertCherrie:
This PR updates webpack-dev-server and all other webpack related package versions.


client/app/core/event-notifications.service.js
```diff
@@ -238,6 +238,18 @@ export function EventNotificationsFactory ($log, $timeout, lodash, CollectionsAp
   function actionCableInit () {
     if (ApplianceInfo.get().asyncNotify) {
       const cable = ActionCable.createConsumer('/ws/notifications')
+
+      // actioncable 5.2.x does not guard against messages where `identifier`
```
Fryguy: Is this something we can push upstream?

Fryguy:
Can we also bump actioncable to 5.2.8-1 ? Noticed that's what's available, but we have 5.2.8.  Not sure what's the difference between 5.2.8 and 5.2.8-1

client/app/core/event-notifications.service.js
GilbertCherrie: These changes were removed as they are no longer needed

package.json
```diff
@@ -125,10 +124,10 @@
     "sinon-chai": "~3.5.0",
     "style-loader": "~2.0.0",
     "url-loader": "~4.1.1",
-    "webpack": "~5.94.0",
-    "webpack-cli": "~4.10.0",
-    "webpack-dev-server": "~3.11.3",
-    "webpack-stats-plugin": "~1.0.2"
+    "webpack": "~5.108.4",
```
Fryguy: Can we start using `^` instead here and elsewhere? This make upgrades continuous, and makes it simpler to gradually upgrade without having to do big jumps.

GilbertCherrie: yea I can change that

asirvadAbrahamVarghese:
@GilbertCherrie 
Looks like the prod build(`yarn build`) is failing because `HtmlWebpackPlugin` from both the dev and prod configs are trying to generate the same index.html file:
image

For prod builds, could we remove the plugin from the dev config and then let the prod config override it? (also, I think we need to keep `WebMKS`)
Could you try adding this to the top of the [webpack-prod config](https://github.com/ManageIQ/manageiq-ui-service/blob/master/config/webpack.prod.js#L15)?
```
config.plugins = config.plugins.filter(plugin => {  
  if (plugin instanceof HtmlWebpackPlugin) {
    return plugin.userOptions?.filename === 'console/webmks.html';
  }
  return true;
});
```

Not entirely sure why this worked with v4. My guess is that it was more lenient, and the last plugin instance silently overwrote the previous one

GilbertCherrie:
@asirvadAbrahamVarghese I fixed the build issue. @Fryguy I had to add back the actioncable fixes as it was breaking

client/app/core/event-notifications.service.js
Fryguy: Do we know what causes the undefined identifier?  Feels like a backend bug.

Asking because this feels like a Band-Aid around some other deeper underlying issue

GilbertCherrie: @Fryguy I updated the actioncable library to match ui-classic and the core repo. This also required some changes to make the notifications work properly but we now no longer need this Band-Aid

asirvadAbrahamVarghese (approved):
Everything looks good to me, including action cable. @Fryguy, can you please take a look and merge?
Image

Not sure if we want to include replacing `actioncable` with `@rails/actioncable` in this PR, or handle it separately

GilbertCherrie:
@asirvadAbrahamVarghese actioncable is needed for the newer version of webpack or we get bugs. We can handle this in a new PR but I feel like it should be ok since I pushed it as its own commit

GilbertCherrie:
@asirvadAbrahamVarghese can you review this again please

asirvadAbrahamVarghese:
@GilbertCherrie Everything seems to be working fine for me, but I didn't quite understand what problem this is solving - [Fix notification bug - 06bfa94](https://github.com/ManageIQ/manageiq-ui-service/pull/2099/changes/06bfa94b195e003dfe6bf327c2bd72632dd8e08e). I'm not sure what the expected behavior should be, so I'd like @Fryguy to take a look as well.

Anyway, from my testing, ui-service consumes the WebSocket message correctly, and the toast message is displayed. Also, could you please rebase this?

GilbertCherrie:
@asirvadAbrahamVarghese the notifications weren't being shown until the user would reload the page. We need the notifications to show up when they happen, which the last commits I made fix

asirvadAbrahamVarghese:
>@asirvadAbrahamVarghese the notifications weren't being shown until the user would reload the page. We need the notifications to show up when they happen, which the last commits I made fix

I'm a bit confused here. Even without this fix, I saw the toast being displayed without reloading the page when ordering a service catalog:
image

Anyway, 🤖 suggested this is a more safer approach, so I'm happy to merge it

asirvadAbrahamVarghese (approved):
LGTM, thanks! Anything else can be a follow-up

PR #2100, opened 2026-07-07
Update eslint to v9

GilbertCherrie:
Update all ESLint packages to the latest supported versions. This PR updates ESLint to version 9, which is the latest version that is supported because of 2 blockers:

@babel/eslint-parser 7.x only supports ESLint ≤9 — upgrading to 8.x requires @babel/core v8, which would break babel-loader (the webpack build)

eslint-plugin-angular@5 and eslint-plugin-import@2 don't declare ESLint 10 support yet


.eslintrc.js
asirvadAbrahamVarghese, line 1: Do we still need both `.eslintrc.js` & `eslint.config.js`? From what I see, `eslint.config.js` is the recommended format going forward. Could we remove `.eslintrc.js`, or is there a reason we need to keep both?

GilbertCherrie: Yea it is not used, I deleted the .eslintrc.js

client/app/states/services/details/details.state.spec.js
```diff
@@ -1,5 +1,5 @@
 /* global $state, $componentController */
-/* eslint-disable no-unused-expressions */
+
```
asirvadAbrahamVarghese, line 2: Just noticed that many files picked up unwanted whitespace like this where the eslint-disable line was removed

asirvadAbrahamVarghese: Or we can ignore them for now since our lint rules aren't strict enough to catch them anyway

asirvadAbrahamVarghese (approved):
Looks fine, thanks

Fryguy:
Backported to `tal` in commit fcac467ca4f5158e34c8c08c16f4e0bcd829a146.

```text
commit fcac467ca4f5158e34c8c08c16f4e0bcd829a146
Author: Asirvad Abraham <Asirvad.Abraham.Varghese@ibm.com>
Date:   Thu Jul 16 10:33:26 2026 +0530

    Merge pull request #2100 from GilbertCherrie/update-eslint
    
    Update eslint to v9
    
    (cherry picked from commit c970bdd862fcfc86be25f6b593a5cc8039afd903)
```

PR #2101, opened 2026-07-08
[WIP] Replace decompress with @xhmikosr/decompress

Fryguy:
Because of https://github.com/advisories/GHSA-mp2f-45pm-3cg9

@asirvadAbrahamVarghese Please review.

WIP because I'm not sure this works - The one place it's used, the packages doesn't seem to be working as intended.

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,6 +143,7 @@
     "bootstrap-select": "~1.13.6",
     "cheerio": "1.0.0-rc.12",
     "d3-color": "~3.1.0",
+    "decompress": "npm:@xhmikosr/decompress@^11.1.3",
     "jquery": "~3.7.1",
     "minimist": "~1.2.8",
     "moment-timezone": "^0.5.41",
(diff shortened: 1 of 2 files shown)


asirvadAbrahamVarghese:
@Fryguy Can we wait on this one for now?
I see this is a transitive dependency of `save-remote-file-webpack-plugin`. Its no longer maintained (the last release was over 5y ago), and it still doesn't have a Webpack 5 version, it only has the original Webpack 4 implementation (although it doesn't seem to break with Webpack 5). I am trying to replace this with a custom Webpack plugin instead

>The one place it's used, the packages doesn't seem to be working as intended

Did you run into a specific issue? From what I can tell, we want `vnc_lite.html` to be downloaded into the `vendor/noVNC/` output path, right?

asirvadAbrahamVarghese:
@Fryguy Could you try this and see if it works: #2109

Fryguy:
🎉 I like that approach so much better - I was looking for a different "download a file" plugin, but writing our own seems better for this case anyway as it's much more explicit.

PR #2103, opened 2026-07-08
Fix sporadic tagging spec failures by using $q instead of native Promises

Fryguy:
The tagging component spec stubbed CollectionsApi with native Promise.resolve(), then used setTimeout(..., 500) to wait for those promises to settle before asserting. Native Promises are microtasks not controlled by Angular's digest cycle, so on a loaded CI machine the 500ms window was occasionally insufficient, causing intermittent test failures.

Replace Promise.resolve() with $q.resolve() so promises are tied to Angular's digest cycle, and replace each setTimeout + done callback with a synchronous $rootScope.$apply() call to flush them deterministically.

@asirvadAbrahamVarghese Please review. I did this with AI, so I'm not sure 100% sure this fixes it, however, this is the only test I can see that uses setTimeout, so it seems plausible. It's impossible to tell what the failing test actually is in CI.

### DIFF
--- a/client/app/shared/tagging/tagging.component.spec.js
+++ b/client/app/shared/tagging/tagging.component.spec.js
@@ -1,10 +1,9 @@
 /* global inject, $httpBackend, readJSON, CollectionsApi */
-/* eslint-disable angular/timeout-service */
 
 describe('Component: taggingWidget', () => {
   beforeEach(() => {
     module('app.components')
-    bard.inject('$httpBackend', 'CollectionsApi')
+    bard.inject('$httpBackend', '$rootScope', '$q', 'CollectionsApi')
   })
 
   describe('with $compile', () => {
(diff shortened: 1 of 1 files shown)


asirvadAbrahamVarghese:
@Fryguy I guess these intermittent failures can't be reproduced locally. If you have any CI run links where it occurred, could you please add them here?

asirvadAbrahamVarghese:
The changes look good to me. Removing the 500ms `setTimeout` delays cut the execution time by over 99% 🎉
**Before:** `Chrome Headless 150.0.0.0 (Mac OS 10.15.7): Executed 6 of 6 SUCCESS (2.655 secs / 2.526 secs)`
**After:** `Chrome Headless 150.0.0.0 (Mac OS 10.15.7): Executed 6 of 6 SUCCESS (0.07 secs / 0.002 secs)`

Can you please rebase this?

asirvadAbrahamVarghese:
@Fryguy could you please rebase this one

Fryguy:
@asirvadAbrahamVarghese Rebased.

asirvadAbrahamVarghese (approved):
Looks great

PR #2109, opened 2026-07-10
Replace save-remote-file-webpack-plugin

asirvadAbrahamVarghese:
Alternative to #2101 
PR to replace [save-remote-file-webpack-plugin](https://www.npmjs.com/package/save-remote-file-webpack-plugin)

### DIFF
--- a/config/download-novnc-plugin.js
+++ b/config/download-novnc-plugin.js
@@ -1,0 +1,74 @@
+const fs = require('fs');
+const https = require('https');
+const path = require('path');
+
+class DownloadNoVNCPlugin {
+  constructor(options = {}) {
+    const packageJson = require('../package.json');
+    const novncVersion = packageJson.dependencies['@novnc/novnc'].replace(/[~^]/, '');
+    
+    this.targetFile = 'vnc_lite.html';
+    this.url = options.url || `https://raw.githubusercontent.com/novnc/noVNC/v${novncVersion}/${this.targetFile}`;
+    this.outputPath = options.outputPath || `vendor/noVNC/${this.targetFile}`;
+    this.downloaded = false;
+  }
+
+  apply(compiler) {
+    // Use 'afterEmit' hook to download AFTER webpack has written all files
+    // This ensures CleanWebpackPlugin doesn't delete our file
+    compiler.hooks.afterEmit.tapAsync('DownloadNoVNCPlugin', (compilation, callback) => {
+      // Only download once per build (prevents duplicate downloads if hook fires multiple times during HMR)
+      if (this.downloaded) {
+        console.log(`\n✓ ${this.targetFile} already downloaded in this build session`);
+        callback();
+        return;
+      }
+
+      const fullOutputPath = path.join(compiler.options.output.path, this.outputPath);
+      
+      // Skip download if file already exists (e.g., from previous build or when switching between yarn build/start)
+      if (fs.existsSync(fullOutputPath)) {
+        console.log(`\n✓ ${this.targetFile} already exists, skipping download`);
+        this.downloaded = true;
+        callback();
+        return;
+      }
+      
+      console.log(`\n⬇ Downloading ${this.targetFile} from:`, this.url);
+      
+      // Create directory if it doesn't exist
+      fs.mkdirSync(path.dirname(fullOutputPath), { recursive: true });
+      console.log('Creating write stream to:', fullOutputPath);
+      const file = fs.createWriteStream(fullOutputPath);
+      
+      https.get(this.url, (response) => {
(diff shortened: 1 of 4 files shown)


config/download-novnc-plugin.js
```diff
@@ -0,0 +1,74 @@
...
+      
+      // Skip download if file already exists (e.g., from previous build or when switching between yarn build/start)
+      if (fs.existsSync(fullOutputPath)) {
+        console.log(`\n✓ ${this.targetFile} already exists, skipping download`);
+        this.downloaded = true;
+        callback();
+        return;
+      }
+      
+      console.log(`\n⬇ Downloading ${this.targetFile} from:`, this.url);
```
asirvadAbrahamVarghese, line 37: On every prod build `fs.existsSync(fullOutputPath)` will be `false` and the file will be downloaded again because [CleanWebpackPlugin clears the entire output directory](https://github.com/ManageIQ/manageiq-ui-service/blob/master/config/webpack.prod.js#L20) before the build:

Image


The benefit is when prod build is run first and then the dev server is started. Since the file is already present, it doesn't need to be downloaded again:
Image

config/download-novnc-plugin.js
```diff
@@ -0,0 +1,74 @@
...
+  apply(compiler) {
+    // Use 'afterEmit' hook to download AFTER webpack has written all files
+    // This ensures CleanWebpackPlugin doesn't delete our file
+    compiler.hooks.afterEmit.tapAsync('DownloadNoVNCPlugin', (compilation, callback) => {
+      // Only download once per build (prevents duplicate downloads if hook fires multiple times during HMR)
+      if (this.downloaded) {
+        console.log(`\n✓ ${this.targetFile} already downloaded in this build session`);
+        callback();
+        return;
+      }
```
asirvadAbrahamVarghese, line 25: We should also skip the download during HMR. When the dev server recompiles after a code change, the file is already present, so there's no need to download it again:

Image

config/download-novnc-plugin.js
```diff
@@ -0,0 +1,74 @@
...
+      
+      // Create directory if it doesn't exist
+      fs.mkdirSync(path.dirname(fullOutputPath), { recursive: true });
+      console.log('Creating write stream to:', fullOutputPath);
+      const file = fs.createWriteStream(fullOutputPath);
+      
+      https.get(this.url, (response) => {
+        if (response.statusCode !== 200) {
+          const error = new Error(`Failed to download: HTTP ${response.statusCode}`);
+          console.error('✗', error.message);
```
asirvadAbrahamVarghese, line 47: And if the download fails, the error is thrown:
Image

Fryguy (approved):

config/download-novnc-plugin.js
```diff
@@ -0,0 +1,74 @@
+const fs = require('fs');
+const https = require('https');
+const path = require('path');
+
+class DownloadNoVNCPlugin {
+  constructor(options = {}) {
+    const packageJson = require('../package.json');
+    const novncVersion = packageJson.dependencies['@novnc/novnc'].replace(/[~^]/, '');
```
Fryguy, line 8: @asirvadAbrahamVarghese The only downside to this is since we use `~`, if they ever release, say, a 1.2.1, we'd pull 1.2.0. Ideally we'd want to pull the value from the lockfile instead which is more accurate. That being say, novnc is up to 1.7.0 and it's very unlikely a 1.2.1 will ever be released so it's fine for now, but if we update to 1.7+ we might run into issues.

asirvadAbrahamVarghese:
>The only downside to this is since we use ~, if they ever release, say, a 1.2.1, we'd pull 1.2.0. Ideally we'd want to pull the value from the lockfile instead which is more accurate.

Yeah, that could definitely be an issue, good catch.
I'm thinking of just reading the version directly from `node_modules`, something like:
```
require('@novnc/novnc/package.json')?.version;
```
The other approaches I can think of are using `yarn why`, but that requires spawning a child process, or extracting it from `yarn.lock`, which would mean using a parser because the lockfile isn't in JSON format

Fryguy:
Backported to `tal` in commit 10d79b949a1db2e031618e8296e6bf0fb7045d5e.

```text
commit 10d79b949a1db2e031618e8296e6bf0fb7045d5e
Author: Jason Frey <fryguy9@gmail.com>
Date:   Fri Jul 10 12:24:43 2026 -0400

    Merge pull request #2109 from asirvadAbrahamVarghese/replace-save-remote-file-webpack-plugin
    
    Replace save-remote-file-webpack-plugin
    
    (cherry picked from commit 8f401705157fb6719637fc37b01ee441345d10a3)
```

PR #2111, opened 2026-07-13
Get exact novnc version from node_modules

asirvadAbrahamVarghese:
Follow up to https://github.com/ManageIQ/manageiq-ui-service/pull/2109

PR to update the version lookup as mentioned in [this comment](https://github.com/ManageIQ/manageiq-ui-service/pull/2109#discussion_r3560468840)
>The only downside to this is since we use ~, if they ever release, say, a 1.2.1, we'd pull 1.2.0. Ideally we'd want to pull the value from the lockfile instead which is more accurate. That being say, novnc is up to 1.7.0 and it's very unlikely a 1.2.1 will ever be released so it's fine for now, but if we update to 1.7+ we might run into issues.

### DIFF
--- a/config/download-novnc-plugin.js
+++ b/config/download-novnc-plugin.js
@@ -4,8 +4,7 @@ const path = require('path');
 
 class DownloadNoVNCPlugin {
   constructor(options = {}) {
-    const packageJson = require('../package.json');
-    const novncVersion = packageJson.dependencies['@novnc/novnc'].replace(/[~^]/, '');
+    const novncVersion = require('@novnc/novnc/package.json')?.version;
     
     this.targetFile = 'vnc_lite.html';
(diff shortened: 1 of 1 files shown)


Fryguy (approved):

Fryguy:
Backported to `tal` in commit 37053ef10e302a6cbb8f1eb349411745f795aeff.

```text
commit 37053ef10e302a6cbb8f1eb349411745f795aeff
Author: Jason Frey <fryguy9@gmail.com>
Date:   Fri Jul 31 10:15:56 2026 -0400

    Merge pull request #2111 from asirvadAbrahamVarghese/enhance-novnc-download-plugin
    
    Get exact novnc version from node_modules
    
    (cherry picked from commit a459032b83bcdb866c93f31cfa98a952c9a8f35b)
```

PR #2166, opened 2026-08-27
Update babel v8

GilbertCherrie:
Replaces https://github.com/ManageIQ/manageiq-ui-service/pull/2160 as this PR causes yarn test to fail.

Update babel monorepo to v8.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -35,7 +35,6 @@ module.exports = function (config) {
     ],
 
     plugins: [
-      'karma-babel-preprocessor',
       'karma-chai',
       'karma-chai-sinon',
       'karma-chrome-launcher',
@@ -47,13 +46,6 @@ module.exports = function (config) {
 
     proxies: {
       '/images/': '/base/client/assets/images/',
-    },
-
-    // preprocess matching files before serving them to the browser
-    // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
-    preprocessors: {
-      './tests/**/*.js': ['babel'],
-      './client/app/**/*.spec.js': ['babel'],
     },
 
     // test results reporter to use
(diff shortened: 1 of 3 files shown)


karma.conf.js
```diff
@@ -52,10 +83,12 @@ module.exports = function (config) {
     // preprocess matching files before serving them to the browser
     // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
     preprocessors: {
-      './tests/**/*.js': ['babel'],
-      './client/app/**/*.spec.js': ['babel'],
+      './tests/**/*.js': ['webpack'],
+      './client/app/**/*.spec.js': ['webpack'],
```
asirvadAbrahamVarghese: Do we actually need this preprocessing stuff to transform test files? As of now I don’t see tests using any syntax or features that would require preprocessing

karma.conf.js
```diff
@@ -52,10 +83,12 @@ module.exports = function (config) {
...
     },
 
+    webpack: webpackConfig,
```
asirvadAbrahamVarghese: If we are removing those preprocessors, that means we no longer need this webpack config that they use

karma.conf.js
```diff
@@ -35,7 +66,7 @@ module.exports = function (config) {
     ],
 
     plugins: [
-      'karma-babel-preprocessor',
+      'karma-webpack',
```
asirvadAbrahamVarghese: This would also mean we no longer need to register the webpack preprocessor with karma

GilbertCherrie:
@asirvadAbrahamVarghese I fixed the pr if you can check

asirvadAbrahamVarghese (approved):
Looks fine, thanks!

PR #2186, opened 2026-09-16
Update chai to latest version

GilbertCherrie:
Replaces https://github.com/ManageIQ/manageiq-ui-service/pull/2162

This PR updates chai from v4 to v6 and removes karma-chai and karma-chai-sinon as they are not able to used with this version of chai. Also, updates sinon-chai


tests/helpers/chai-bundle-entry.js
```diff
@@ -0,0 +1,10 @@
+import * as chai from 'chai'
+import sinonChai from 'sinon-chai'
+
+chai.use(sinonChai)
+
+// Expose as individual globals that tests and karma-sinon expect
+window.chai   = chai
+window.expect = chai.expect
+window.assert = chai.assert
+window.should = chai.should()
```
asirvadAbrahamVarghese: Looks like we don't use `assert` or `should` anywhere; we only use `expect`. If they're there for potential future use, we can keep them. I also don't think we need `window.chai`, since we'd probably want to add any new ones as individual globals, like the others here

asirvadAbrahamVarghese:
Looks fine to me, @GilbertCherrie could you please rebase this one?

GilbertCherrie:
@asirvadAbrahamVarghese rebased and ready

asirvadAbrahamVarghese (approved):

PR #2190, opened 2026-09-16
Update sass loader

GilbertCherrie:
Replaces: https://github.com/ManageIQ/manageiq-ui-service/pull/2188

This PR updates sass-loader to the latest version


client/assets/sass/styles.sass
```diff
@@ -1,11 +1,11 @@
 // Global constants
 $img-base-path: '/' !default
 
-@import 'patternfly'
+@import './_patternfly'
 
 // Include font-fabulous
 $ff-font-path: '~@manageiq/font-fabulous/assets/fonts/font-fabulous'
-@import 'font-fabulous'
+@import '_font-fabulous'
```
asirvadAbrahamVarghese: Do we need these `_` prefixes? I believe they should be resolved naturally by Dart Sass or sass-loader level

client/assets/sass/_patternfly.sass
```diff
@@ -6,4 +6,4 @@ $fa-font-path: '~font-awesome/fonts'
 $img-path: '~patternfly/dist/img/'
 $font-path: '~patternfly/dist/fonts/'
 
-@import '~patternfly/dist/sass/patternfly'
+@import '~patternfly/dist/sass/_patternfly.scss'
```
asirvadAbrahamVarghese: Same thought as [above](https://github.com/ManageIQ/manageiq-ui-service/pull/2190/changes/5c4a69f5bd359e52e39de07550bc3a2685f62493#r4035319004), but this won't be resolved by Dart Sass because of the `~`, so it should be handled by sass-loader

config/webpack.dev.js
```diff
@@ -218,10 +218,15 @@ module.exports = {
   ],
 
   resolve: {
-    extensions: ['.js'],
+    modules: [
+      path.resolve(root, 'assets/sass'),
+      'node_modules',
+    ],
+    extensions: ['.js', '.sass', '.scss', '.css'],
```
asirvadAbrahamVarghese: I think this is dead configuration, since 
- sass-loader's resolver should already handle the `.scss`, `.sass`, and `.css` extensions, so we probably don't need to specify them here
- Also, `resolve.modules` should be handled through `sassOptions.loadPaths`, so we likely don't need `resolve.modules` here either

asirvadAbrahamVarghese (commented):
Looks good overall, just a few minor thoughts on the changes

GilbertCherrie:
@asirvadAbrahamVarghese I have made the fixes if you can review again now

asirvadAbrahamVarghese (approved):

